PR #1, opened 2018-10-30
Add constant stories and update colors

snags88:
Colors:
screen shot 2018-10-29 at 9 01 50 pm

Not pretty, but a way to dynamically list all constant import names and key value pairs.
screen shot 2018-10-29 at 9 11 56 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.11.1",
     "husky": "^1.1.2",
+    "isobject": "^3.0.1",
     "jest": "^23.6.0",
     "lint-staged": "^7.3.0",
     "lodash.round": "^4.0.4",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,6 @@
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.11.1",
     "husky": "^1.1.2",
-    "isobject": "^3.0.1",
     "jest": "^23.6.0",
     "lint-staged": "^7.3.0",
     "lodash.round": "^4.0.4",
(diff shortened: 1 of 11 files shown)


stories/constants/Color.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'react-emotion';
+
+const ColorContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+
```
smmullen90: [nit] can we get rid of these extra lines?

stories/constants/index.js
```diff
@@ -0,0 +1,63 @@
...
+
+const stories = storiesOf('CONSTANTS', module);
+
+stories.add(
+  'available constants',
+  withDocs(ConstantsReadme, () => (
+    <div css="text-align: left;">
+      {
+        Object.keys(CONSTANTS).map(category => {
+          if (['COLORS'].includes(category)) {
```
smmullen90: not really necessary for this since there is only one, but you could pull out in separate const so its easy to add to. 

```
const CONSTANT_BLACKLIST = ['COLORS'];
```
could also use a `filter` which might make it more readable, this way is more optimized tho

package.json
```diff
@@ -70,6 +70,7 @@
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.11.1",
     "husky": "^1.1.2",
+    "isobject": "^3.0.1",
```
smmullen90: any reason you chose this package specifically? can we not use `typeof`?

snags88: It's a 1 liner package that i can just move into a util. But the package also checks that it's not an array

snags88:
@smmullen90 updated with changes

PR #2, opened 2018-10-30
rename some Typography components

snags88:
Chatted with @benkolde regarding the naming and making it consistent.

Before, we were importing the typography components at the top level:
```jsx
import { ErrorText } from 'shared-components\typography';

<ErrorText>Error Message!</ErrorText>
```

However, now it's namespaced so we have more context AND we won't run into name collision issues
```jsx
import { Typography } from 'radiance-ui';

<Typography.Error>Error Message!</Typography.Error>
```

Also took care to not make any breaking changes until a major release version. Aka:
- Removed old usage from documentation but kept it in the code
- Added note to remove for next major

### DIFF
--- a/src/shared-components/typography/index.js
+++ b/src/shared-components/typography/index.js
@@ -89,24 +89,30 @@ export const style =  {
   link: linkStyle,
 };
 
-const ButtonText = styled.span(buttonStyle);
+const Button = styled.span(buttonStyle);
 const Caption = styled.p(captionStyle);
 const Display = styled.h1(displayStyle);
-const ErrorText = styled.p(errorStyle);
+const ErrorComponent = styled.p(errorStyle);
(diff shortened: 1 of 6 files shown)


src/shared-components/typography/index.js
```diff
@@ -89,24 +89,30 @@ export const style =  {
...
   Heading,
   Label,
-  LinkTag,
-  SuccessText,
+  Link,
+  Success,
   Title,
+
+  // Legacy names. Will be removed in v2 (next major)
+  LinkTag: Link,
```
smmullen90: wonder if you can add in a proxy for deprecation warnings here

snags88: @smmullen90 updated in 98c2825

PR #4, opened 2018-10-31
Configure enzyme and add Icon component

snags88:
This PR adds the `Icon` component, tests, and story.

I also needed to configure enzyme for jest so there's some config files in the diff.

Story:
screen shot 2018-10-31 at 2 42 37 pm


snags88:
[Convert Icon component](https://trello.com/c/4vOyK3uM/177-convert-icon-component)

src/shared-components/icon/index.js
```diff
@@ -0,0 +1,66 @@
...
+  // eslint-disable-next-line no-undef
+  const icon = window.__SVG_INLINE__
+    ? `#${iconName}`
+    : `${iconFilePath}#${iconName}`;
+  return (
+    <svg
+      width={width}
+      height={height}
+      fill={fill}
+      data-dusk-id={duskId}
```
smmullen90: dont think we should have this in component library.

src/shared-components/icon/index.js
```diff
@@ -0,0 +1,66 @@
...
+      css={`
+        display: ${inline ? 'inline-block' : 'block'};
+        transform: rotate(${rotate}deg);
+        color: ${fill};
+        transition: color 350ms;
```
smmullen90: can we take this out of the component library?

src/shared-components/icon/index.js
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <svg
+      width={width}
+      height={height}
+      fill={fill}
+      data-dusk-id={duskId}
+      css={`
+        display: ${inline ? 'inline-block' : 'block'};
+        transform: rotate(${rotate}deg);
+        color: ${fill};
```
smmullen90: fill doesnt really mean the same thing as color, i think we should update one of the two

src/shared-components/icon/index.js
gwlortscher: I believe we can just add this during usage of the component and don't need to pass it in— `<Icon data-dusk-id="something" />`

snags88: yup, I can definitely remove it. Was doing a lot of brainless copy pasta 🍝

src/shared-components/icon/index.js
snags88: so it does in certain cases. There was a movement once (started by me of course 😅) to set the fill on the definition of svgs to `currentColor` so that svgs can basically inherit the color set on the parent container rather than using `fill` css property which would only work on SVGs.

Looking at it now, we probably just want granular control over the fill by being more declarative on the icon set i guess? I'm just trying to think of times where the text + icon inside of a container would need to change color at the same time and having `currentColor` would be easier... Actually that might have been why we did it this way to begin with. May have been something on the signup flow where we have icons as radio buttons. It would also explain the transition on the color attribute.

src/shared-components/icon/index.js
snags88: are you talking about the color transition? or this entire CSS block?

docs/icon.md
```diff
@@ -0,0 +1,28 @@
+# Icon
+### Usage
+
+```jsx
+import { Icon } from 'radiance-ui';
+
+<Icon width={14} height={14} iconName="camera" />
```
benkolde: Icons _should_ all be 16x16 by default. I don't know what this looks like everywhere else. 

Might be worth a look but I am considering adding a smaller size in the future.

snags88:
@smmullen90 responded to your comments. Lmk what you think

src/shared-components/icon/index.js
smmullen90: color transition, but maybe we actually do need it? ask ben, up to you

src/shared-components/icon/index.js
smmullen90: yea i like using currentColor. i just mean the exposed API should map to what is happening, so the prop should  probably be named `color` rather than fill

snags88: For the sake of timeliness of converting things, I'd rather not change the interface and continue using `fill`. Also, `fill` isn't a stretch since all of our icons are SVG icons

PR #5, opened 2018-11-03
Add Alert component + story + tests

snags88:
This PR adds the Alert component, story, and tests.

The Alert component was updated with the following:
- No longer requires `alert` object as a prop. Instead the props required are not nested and passed in at the top level (i.e. `alert.alertText` is now just `text`)
- `duration` prop used to be in units of milliseconds, but is now in seconds (could go either way on this)
- The component is now self-cleaning. Basically before, you would need something external to remove the alert component from the dom after it was dismissed. However, now it holds an `exited` state and will render null so that cleans itself up after exiting.

Other additions in this PR:
- `react-test-renderer` & `jest-emotion` was installed to do snapshot testing 
- Alert story was added

screen shot 2018-11-02 at 9 58 46 pm
screen shot 2018-11-02 at 9 58 52 pm


snags88:
[Convert Alert component](https://trello.com/c/SikR9Umd/205-convert-alert-component)

src/shared-components/alert/style.js
```diff
@@ -0,0 +1,118 @@
...
+  from { opacity: 0; transform: translate3d(24px, 0, 0); }
+  to { opacity: 1; transform: translate3d(0, 0, 0); }
+`;
+
+const fadeInMobile = keyframes`
+  from { opacity: 0; transform: translate3d(0, -24px, 0);  }
+  to { opacity: 1; transform: translate3d(0, 0px, 0); }
+`;
+
+const successAlertStyles = css`
```
smmullen90: nit, pref if you pull out `baseAlertStyles` (`border-width` & `border-style`)

src/shared-components/alert/style.js
```diff
@@ -0,0 +1,118 @@
...
+
+export const AlertContainer = styled.div`
+  align-items: flex-start;
+  animation: ${fadeInMobile} ${ANIMATION.defaultTiming} 1;
+  cursor: pointer;
+  display: flex;
+  padding: ${SPACING.xsmall} 0px;
+  font-size: ${TYPOGRAPHY_CONSTANTS.fontSize.caption};
+  justify-content: space-around;
+  opacity: ${props => (props.exiting ? '0' : '1')};
```
smmullen90: werent you talking about how we need to get away from interpolated functions like this (and the one later)

snags88: i believe it's only on `css` functions. When it's `styled.div` it's ok to do because it is an actual React component

smmullen90: 👍

snags88:
@smmullen90 addressed in 59e5850

PR #13, opened 2018-11-11
[Icons] Configure new inline Icons and convert some icons

snags88:
This PR lays down a few things:
- Adds configuration for our new Icon system to use individual svg files and use [SVGR](https://github.com/smooth-code/svgr) to convert the svg files to react components at build time (svgr plugin for Rollup for distribution files && webpack plugin for storybook). The added configurations include:
  - A svg to jsx transform template
  - `create-index` package to add index files to both `src/icons/` and `src/svgs` 
- Fixed the `Alert` component to use the new svg icons
- Also fix the icon storybook and documentation


src/utils/icons/index.js
```diff
@@ -0,0 +1,29 @@
...
+  rotate: 0,
+  width: 16,
+  height: 16,
+};
+
+export const iconStyles = props => css`
+  display: ${props.inline ? "inline-block" : "block"};
+  transform: rotate(${props.rotate}deg);
+  color: ${props.fill};
+  transition: color 350ms;
```
mattam: Do we perhaps want the transition to be something with a defaultProp which we could change too?

snags88: Currently, there hasn't been a need to transition anything outside of the color. If we need to transition more properties, we can update the code at the time. The preference is to keep things simple until we need it, especially with how much more work is left for this project

snags88:
[Convert Icon component](https://trello.com/c/4vOyK3uM/177-convert-icon-component)

PR #27, opened 2018-11-30
[Chip] Add `<Chip />` component, test, and story

snags88:
This adds the Chip component to Radiance

[Storybook](https://curology-radiance-pr-27.herokuapp.com/?selectedKind=Tag&selectedStory=Usage&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs)


src/shared-components/tag/style.js
```diff
@@ -0,0 +1,74 @@
+import styled, { css } from 'react-emotion';
+
+import { style as TYPOGRAPHY_STYLE } from '../typography';
+import { COLORS, SPACING } from '../../constants';
+
+const closedStyle = css`
+  background-color: ${COLORS.statusGreyBackground};
+  border-color: ${COLORS.statusGreyBorder};
+  span {
```
smmullen90: [note] i **think** if you rearrange, you can actually interpolate `TagText` here to make this more specific

snags88: Yup that makes sense. Will update 👍

snags88:
@smmullen90 I can chat with ben regarding the naming!

src/shared-components/tag/style.js
benkolde: @snags88 @smmullen90 I believe we have `badge` already. 

It's this: 
image

It is similar though, wonder if we could have it as two separate types of `badges`. 

I'll throw it up in the slack channel for people to discuss!

snags88:
Consensus was to go with `chip`

snags88:
[Convert Chip component](https://trello.com/c/sJKwmiji/241-convert-chip-component)

PR #28, opened 2018-12-01
Add in the container components

smmullen90:
Resolves: https://github.com/PocketDerm/radiance-ui/issues/10

Please feel free to merge this while I am away if it looks good!


snags88:
@smmullen90: @benkolde and I talked about this offline and we want to explore doing something like:

```jsx
<Container>
  <Container.Image src="something" />

  <Container.Section>
    Some stuff
  </Container.Section>

  <Container.Divider />

  <Container.Section>
    Some more stuff
  </Container.Section>
</Container>
```

Essentially we're seeing a lot of the same patterns of things inside of the containers with dividers and sections so we want to expose these building blocks from `Container` and let people compose what they want. If the design doesn't match up with the standard section, divider, or image, then they can create their own emotion component.

- `Container.Image` would just make sure that the item is 100% width
- `Container.Section` would only have opinions on basic padding. Easily overridden by `css` prop.
- `Container.Divider` provides us with the same "partial" divider


Examples of repeated patterns:
screen shot 2018-12-04 at 2 05 08 pm

screen shot 2018-12-04 at 2 05 16 pm

screen shot 2018-12-04 at 2 12 16 pm

benkolde:
@snags88 @smmullen90 LGTM

Curious on how much we can customize this to change layouts when these containers are on desktop as well

snags88:
@benkolde can you be a little bit more specific about the changes? Currently i'm imagining the only changes would be like padding on `Divider`/`Section` and leave it pretty unopinionated. Within the `Container` you'll be able to put other components in it that reacts to screen sizes

mattam:
I like the idea of this generic container that covers a majority of our use cases and also the standardization of the internal components so that we don't have to keep reinventing the wheel :D 

Regarding the desktop thing, the advantage of this standardized component is that we can automatically have it adjust to desktop layout in a standardized way.  So all the containers on the page could have more consistent adjustments to the layout.

snags88:
@mattam yeah the only part about the standard desktop layout is that we don't really have a standard layout except for maybe padding. We don't have a consistent way of things flexing from column to row or stuff like that so i'm more in favor of leaving it for the particular UI's implementation to decide that. It also leaves room for our designers to play with various styles

src/shared-components/container/index.js
```diff
@@ -0,0 +1,26 @@
+import {
+  Container,
+  ClickableContainer,
+  MessageContainer,
+  Section,
+  Divider,
+  Image,
+} from './style';
+
+Container.Section = Section;
```
smmullen90: no reason to have this as 3 separate things. clickable vs message should now be passed as a prop to outer container  i think

snags88:
[Create Container component in Radiance](https://trello.com/c/CAN9utUD/32-create-container-component-in-radiance)

src/shared-components/container/index.js
snags88: fixed in 6e41aff

snags88:
@smmullen90 @mattam @benkolde updated the containers with what was discussed in this PR. You can check out the usage on [storybook](https://curology-radiance-pr-28.herokuapp.com/?selectedKind=Containers&selectedStory=Container&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs)

smmullen90:
@snags88 i cant approve my own PR, but your updates look good. One thing that might be interesting as a follow up is determining whether or not we want to add some logic for the desktop versions wrt image placement.

pre-consultation-module

snags88:
@smmullen90 yup, we had a discussion around that and where we ended up was to leave the containers super flexible for now and then build these common components we're seeing as reusable components with in our app. If the reusable components make it into the design system, then we can promote it from a reusable component in PocketDerm repo to Radiance-UI

PR #32, opened 2018-12-17
added accordion to radiance

ZeMunchkin:
Trello: https://trello.com/c/S0ijYQ8O/261-move-accordion-to-radiance

Moving accordion to radiance from shared-components

### DIFF
--- a/docs/accordion.md
+++ b/docs/accordion.md
@@ -1,0 +1,93 @@
+# Accordion
+### Usage
+
+```jsx
+import { Accordion } from 'radiance-ui';
+
+<Accordion
+  title={<div>Title</div>}
+  isOpen={boolean}
+  onClick={() => {}}
+>
+  <div>Expandable portion as children</div>
+</Accordion>
+
+```
+
+<!-- STORY -->
+
+### Proptypes
+| prop     | propType           | required | default | description                                                                                                                  |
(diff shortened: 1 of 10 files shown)


stories/accordion/index.js
```diff
@@ -0,0 +1,49 @@
...
+          <Accordion.Container>
+            <Accordion
+              title={
+                <Accordion.Content>
+                  <Accordion.Truncate>
+                    {text('Title text', 'Title')}
+                  </Accordion.Truncate>
+                </Accordion.Content>
+              }
+              onClick={() => {}}
```
snags88: we can use an "action" to log the click similar to this: https://github.com/PocketDerm/radiance-ui/blob/644500fc6ec02eb3d41af1168a4bafcd1185096f/stories/radioButton/index.js#L5

docs/accordion.md
```diff
@@ -0,0 +1,93 @@
+# Accordion
+### Usage
+
+```jsx
+import { Accordion } from 'radiance-ui';
+
+<Accordion
```
snags88: for these example codesnippets, I've been writing the almost the exact same code as what gets rendered in the story below it with some modifications (i.e. don't need show the exact function definition of `onClick` etc.

You can check out the example on the existing stories

ZeMunchkin:
@snags88 Updated!

PR #39, opened 2019-01-24
added immersive and regular modals

ZeMunchkin:
Trello: https://trello.com/c/5arVBzoV

- Added Modal and ImmersiveModal (and necessary utils) to radiance library 
- Added MenuIcon to icon library

### DIFF
--- a/docs/immersiveModal.md
+++ b/docs/immersiveModal.md
@@ -1,0 +1,89 @@
+# ImmersiveModal
+### Usage
+
+```jsx
+import React from 'react';
+
+import { ImmersiveModal, Button } from 'radiance-ui';
+
+const HeaderImage = () => (
+  <div>
+    This is a placeholder for an optional header image.
+  </div>
+);
+
+class DefaultImmersiveModal extends React.Component {
+  state = {
+    defaultIsOpen: false,
+    headerIsOpen: false,
+  };
+
+  onOpenDefaultModal = () => this.setState({ defaultIsOpen: true });
+
+  onOpenHeaderModal = () => this.setState({ headerIsOpen: true });
+
+  onClose = () => this.setState({
+    defaultIsOpen: false,
+    headerIsOpen: false,
+  });
+
+  render() {
+    const { defaultIsOpen, headerIsOpen } = this.state;
+
+    return (
+      <div>
+        <Button onClick={this.onOpenDefaultModal}>Open ImmersiveModal</Button>
+        {defaultIsOpen && (
+          <ImmersiveModal onClose={this.onClose}>
+            <ImmersiveModal.Title>This is styled with ImmersiveModalTitle</ImmersiveModal.Title>
+            <ImmersiveModal.Body>This is styled with ImmersiveModalBody.</ImmersiveModal.Body>
+            <ImmersiveModal.Footer>
+              This is styled with ImmersiveModalFooter. It gives us a padding to separate
+              from the body.
+              <Button.Container>
+                <Button onClick={this.onClose}>Close ImmersiveModal</Button>
+              </Button.Container>
+            </ImmersiveModal.Footer>
+          </ImmersiveModal>
+        )}
+
+        <Button onClick={this.onOpenHeaderModal}>Open ImmersiveModal with Header</Button>
+        {headerIsOpen && (
+          <ImmersiveModal
+            onClose={this.onClose}
+            header={<HeaderImage />}
+          >
+            <ImmersiveModal.Title>This is styled with ImmersiveModalTitle</ImmersiveModal.Title>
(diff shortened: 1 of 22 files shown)


docs/immersiveModal.md
```diff
@@ -0,0 +1,89 @@
...
+
+  render() {
+    const { defaultIsOpen, headerIsOpen } = this.state;
+
+    return (
+      <div>
+        <Button onClick={this.onOpenDefaultModal}>Open ImmersiveModal</Button>
+        {defaultIsOpen && (
+          <ImmersiveModal onClose={this.onClose}>
+            <ImmersiveModal.Title>This is styled with ImmersiveModalTitle</ImmersiveModal.Title>
```
snags88: probably want to update the text to say "This is styled with `ImmersiveModal.Title`". Similar comment for the rest of the file

docs/immersiveModal.md
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+```
+
+<!-- STORY -->
+
+### Proptypes
+| prop     | propType           | required | default | description                                                                                                                  |
+|----------|--------------------|----------|---------|------------------------------------------------------------------------------------------------------------------------------|
+| onClose | func | no | () => {} | function is that executed when either the close icon is clicked or any part of the page outside the modal is clicked |
```
snags88: function that is*

docs/modal.md
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+```
+
+<!-- STORY -->
+
+### Proptypes
+| prop     | propType           | required | default | description                                                                                                                  |
+|----------|--------------------|----------|---------|------------------------------------------------------------------------------------------------------------------------------|
+| onClose | func | no | () => {} | Function is that executed when the close icon is clicked |
```
snags88: function that is*

src/constants/keycodes/index.js
```diff
@@ -0,0 +1,12 @@
+import throwOnUndefinedProperty from '../../utils/throwOnUndefinedProperty';
+
+const keycodes = throwOnUndefinedProperty({
+  backspace: 8,
+  enter: 13,
+  escape: 27,
+  space: 32,
+  h: 104, // used for hugpacks
```
snags88: We can strip out the ones we don't need. Probably only need `escape`

src/shared-components/immersiveModal/index.js
```diff
@@ -0,0 +1,82 @@
...
+  static Body = Body;
+
+  static Footer = Footer;
+
+  constructor(props) {
+    super(props);
+
+    this.htmlNode = document.querySelector('html');
+    this.domNode =
+      document.querySelector('#reactPortalSection') || document.body;
```
snags88: just curious what happens if query selector is null. It seems from the code that it's not a hard fast requirement to have a node for the react portal

src/shared-components/immersiveModal/index.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  componentDidMount() {
+    this.htmlNode.classList.add('no-scroll');
```
snags88: just wanted to check if `no-scroll` class was part of the inject global styles

src/shared-components/modal/index.js
```diff
@@ -0,0 +1,126 @@
...
+  static Body = Body;
+
+  static Footer = Footer;
+
+  state = {
+    isVisible: false,
+    parentNode: document.querySelector('#reactPortalSection') || document.body,
+  };
+
+  componentDidMount() {
```
snags88: can we add these listeners to the immersive modal?

src/shared-components/modal/index.js
```diff
@@ -0,0 +1,126 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ReactModal from 'react-modal';
```
snags88: as a follow up (not high priority), i'd like for us to refactor this modal to not use ReactModal. I don't think we need it and we're doing the immersive modal without it.

src/shared-components/immersiveModal/index.js
ZeMunchkin: Confirmed -- just checked in radiance, gatsby, and main app and it's the same in all three.

src/shared-components/modal/index.js
ZeMunchkin: Created an issue for follow up: https://github.com/PocketDerm/radiance-ui/issues/40#issue-403271409

ZeMunchkin:
@snags88 Updated!

src/shared-components/immersiveModal/index.js
snags88: 🙌

PR #41, opened 2019-01-25
Emotion 10

ZeMunchkin:
Trello: https://trello.com/c/KuLojsv2

Updating Radiance to use Emotion 10. This will be a major version bump to V2 when published, as it is a breaking change. 

Also created a `Radiance_V1` branch for updates to V1, to be maintained until our app is also updated to Emotion 10. 

Docs: https://github.com/emotion-js/emotion/blob/master/docs/migrating-to-emotion-10.md

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,13 +1,17 @@
 {
   "extends": "curology",
   "overrides": [
-     {
-       "files": ["test.js"],
-       "rules": {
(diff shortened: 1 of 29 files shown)


ZeMunchkin:
@snags88 
Updated with tests fixed

snags88:
👍

PR #47, opened 2019-01-30
export legacy colors

ZeMunchkin:
Added legacy colors to the exported object

### DIFF
--- a/src/constants/colors/index.js
+++ b/src/constants/colors/index.js
@@ -135,6 +135,7 @@ export const guideColors = {
 export default Object.assign(
   {},
   brandColors,
+  legacyColors,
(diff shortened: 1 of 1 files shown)


ZeMunchkin:
I think we're going to find them scattered throughout the static pages. For example, the legacy orange color is the signup button color on the landing page / footers. I currently need that color for the nav bar. 

screen shot 2019-01-30 at 3 43 35 pm

Could check with @mvandy if she wants to use a different color for this? It's in a few places on the landing page.

ZeMunchkin:
I can also just put these in the gatsby constants for now and revist when we move towards a redesign.

PR #48, opened 2019-02-06
[Icons] Add clipboard

Yihwan:
screen shot 2019-02-06 at 10 58 09 am


src/svgs/icons/clipboard-icon.svg
```diff
@@ -0,0 +1,8 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
snags88: no need to specify width and height on all the tags. The view box will take care of it and it'll allow us to scale the SVG when we use it

Yihwan:
Investigating why icon seems to be smaller and lighter than others: 

screen shot 2019-02-06 at 11 01 15 am

src/svgs/icons/clipboard-icon.svg
Yihwan: seems to work now: 

screen shot 2019-02-06 at 11 10 20 am

removed mask and g tags and other misc seemingly unnecessary info.

ZeMunchkin:
Just going to add that if you need to use this on the main app and it hasn't been fully converted to emotion 10 yet, you'll also need to merge this into the `Radiance_V1` branch and publish both

PR #49, opened 2019-02-06
Creating Banner that looks like a full width Alert

mattam:
Currently the Alert component has a fixed width of 327px, but we want to allow it to expand out to the container.  (This is for the shipmentV2 changes, specifically for the delinquency PR: https://github.com/PocketDerm/PocketDerm/pull/4123 ) 

This PR adds in a fullWidth bool on the Alert box so that we can allow it to expand to the full width of the container.

screen shot 2019-02-06 at 1 13 52 pm


docs/banner.md
```diff
@@ -1,55 +1,63 @@
...
-  <Alert
+  <Banner
     text="Oops! We couldn't update your information!"
     type="danger"
     duration="sticky"
     onExit={() => {}}
   />
-</Alert.Container>
+</Banner.Container>
+  <Banner
```
snags88: looks weird to have this out on its own. Wouldn't we want to use this banner inside of the container?

mattam:
@snags88 : ok, updated the <Banner /> component as discussed

screen shot 2019-02-07 at 4 44 00 pm

docs/banner.md
```diff
@@ -0,0 +1,25 @@
...
+  onClick={() => { console.log('banner clicked')}}
+/>
+```
+
+<!-- STORY -->
+
+### Proptypes
+| prop        | propType           | required | default | description                                                                                                                  |
+|-------------|--------------------|----------|---------|------------------------------------------------------------------------------------------------------------------------------|
+| content     | string             | yes      | -       | banner content to be displayed                                                                                               |
```
snags88: We need to update propType to `node`

src/shared-components/banner/index.js
```diff
@@ -0,0 +1,50 @@
...
+} from './style';
+
+const bannerIconMapping = {
+  success: CheckmarkIcon,
+  danger: ErrorIcon,
+  info: InfoIcon,
+};
+
+class Banner extends React.Component {
+  static Container = ({ children }) => <BannersContainer>{children}</BannersContainer>;
```
snags88: we don't need to do this. we can just do:
```js
static Container = BannersContainer;
```

src/shared-components/banner/index.js
```diff
@@ -0,0 +1,50 @@
...
+
+  static propTypes = {
+    content: PropTypes.node.isRequired,
+    type: PropTypes.oneOf(['success', 'danger', 'info']).isRequired,
+    onClick: PropTypes.func.isRequired,
```
snags88: why is `onClick` required?

src/shared-components/banner/style.js
```diff
@@ -0,0 +1,92 @@
...
+
+import {
+  COLORS,
+  MEDIA_QUERIES,
+  BOX_SHADOWS,
+  SPACING,
+  TYPOGRAPHY_CONSTANTS,
+} from '../../constants';
+
+const successBannerStyles = css`
```
snags88: i think this is from the alerts styles. Can we refactor this a bit to be more DRY since we can reuse this "success" theme with the background color, font color, etc.

src/shared-components/banner/style.js
```diff
@@ -0,0 +1,92 @@
...
+  border-color: ${COLORS.statusGreyBorder};
+  color: ${COLORS.statusGrey};
+  fill: ${COLORS.statusGrey};
+`;
+
+export const BannerContainer = styled.div`
+  align-items: flex-start;
+  border-width: 1px;
+  border-style: solid;
+  box-shadow: ${BOX_SHADOWS.message};
```
snags88: does the banner need box shadow? just curious since it's not the exact same as alerts in that it floats above things. Maybe they do?

src/shared-components/banner/style.js
```diff
@@ -0,0 +1,92 @@
...
+  })}
+
+  ${MEDIA_QUERIES.lgUp} {
+    font-size: ${TYPOGRAPHY_CONSTANTS.fontSize.caption};
+    margin-bottom: ${SPACING.small};
+    padding: ${SPACING.xsmall} 0px;
+  }
+`;
+
+export const BannersContainer = styled.div`
```
snags88: Not entirely sold that we need this. Banners currently are being used in the flow of the natural dom, whereas the alert needs this container.

mattam:
The onClick doesn't need to be required, but we do want to use it for the Delinquent credit card case where we want the user to have the option of clicking on the banner to go to the payment page to update their credit card info.  Removing the `.isRequired` requirement.

snags88:
Sounds good regarding the `onClick`

src/shared-components/banner/style.js
mattam: Thinking of doing a separate PR to DRY out the color styles between the <Alert />, <Banner /> and <Chip /> since they are using the same groupings of background, border and text colors.

mattam: made new Pivotal task to track this DRY change after the emotion stuff goes in so that there's less churn around the `<Banner /> <Alert /> <Chip />` stuff.

src/shared-components/banner/style.js
mattam: The design shows it as having shadows 
screen shot 2019-02-14 at 2 49 17 pm

snags88:
@mattam what's the status on this?

mattam:
Using this change is queued up with this Pivotal task: https://www.pivotaltracker.com/story/show/163775188
Will be looping back to this soon

PR #51, opened 2019-02-06
[Emotion v10] Finish upgrading all things to emotion 10

snags88:
This PR adds the following to fully convert to emotion 10
- Use `babel-preset-css-prop`
- Remove `babel-plugin-emotion`
- Update all `css` prop to not use string and align with new usage from emotion 10.
  - This includes docs, stories, and src files.
  - Also includes the svg -> react component template that SVGR uses.
- Update jest configuration to use new `jest-emotion` config in serializer.
- Update babel configurations to stop using emotion plugin and use the emotion css prop preset.
- Update rollup config globals.


src/shared-components/alert/index.js
```diff
@@ -82,7 +83,7 @@ class Alert extends React.Component {
       >
         <AlertContentContainer>
           <Icon
-            css={[alertIconStyles]}
+            css={css`${alertIconStyles};`}
```
ZeMunchkin: Can you clarify something? `alertIconStyles` is created in style.js using `css`. Is there a difference between creating it with `css` vs wrapping in `css` here?

src/shared-components/selectorButton/index.js
```diff
@@ -45,9 +46,9 @@ const SelectorButton = ({
   >
     <SelectorContainer>
       <CheckmarkIcon
-        css={`
-          color: ${checked ? COLORS.white : 'transparent'};
+        css={css`
           ${selectorIcon};
+          color: ${checked ? COLORS.white : 'transparent'};
```
ZeMunchkin: 🙌 Good refactor on ordering

src/shared-components/alert/index.js
snags88: oh nope, there's not. Both ways here are valid. https://emotion.sh/docs/composition

I just did a global search on `css=` and updated all of it to use the template literal.

ZeMunchkin: 👍

PR #53, opened 2019-02-12
added loadingSpinner to library

ZeMunchkin:
In this PR:
- Adds LoadingSpinner component and story

### DIFF
--- a/src/shared-components/loadingSpinner/index.js
+++ b/src/shared-components/loadingSpinner/index.js
@@ -1,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { COLORS } from '../../constants';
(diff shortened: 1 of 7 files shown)


src/shared-components/loadingSpinner/index.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { COLORS } from '../../constants';
+import { LoadingContainer, LoadingOverlay, LoadingDot } from './style';
```
snags88: probably can drop the `Loading` prefix on the styles

ZeMunchkin:
@snags88 Updated to remove prefix. I didn't see the radiance trello board to make the card -- Is it private?

PR #56, opened 2019-02-16
Added spaceship glyph.

jcmustin:
Added a spaceship glyph for use in the two-week checkin flow.

screen shot 2019-03-06 at 9 57 40 am


jcmustin:
> did you run this through kraken.io?

I have now. Thanks for the heads up!

PR #57, opened 2019-02-16
Course glyph

jcmustin:
Added a course glyph for use in the two-week checkin flow.

screen shot 2019-03-06 at 9 56 53 am


smmullen90:
For these it might be useful to add an image of what the icon looks like

PR #59, opened 2019-02-19
VerticalDivider component

jcmustin:
Added VerticalDivider component to Container.
screen shot 2019-02-19 at 3 21 39 pm

### DIFF
--- a/src/shared-components/container/index.js
+++ b/src/shared-components/container/index.js
@@ -4,11 +4,12 @@ import {
   Container,
   Section,
   Divider,
-  Image,
+  Image, VerticalDivider,
 } from './style';
 
 Container.Section = Section;
 Container.Divider = Divider;
+Container.VerticalDivider = VerticalDivider;
 Container.Image = Image;
 
 Container.propTypes = {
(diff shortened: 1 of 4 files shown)


src/shared-components/container/index.js
```diff
@@ -4,11 +4,12 @@ import {
   Container,
   Section,
   Divider,
-  Image,
+  Image, VerticalDivider,
```
snags88: Need to move this to a new line. Weird that eslint didn't catch it.

jcmustin:
> Could you update the storybook example so that we can see what a container looks like with the divider?

@snags88 Oh, gosh, of course, sorry! I took screenshots for just that purpose and then just forgot to attach them. I've added the output of the snapshotted component, which includes a vertical divider at the bottom.

src/shared-components/container/index.js
jcmustin: That is bizarre; I feel like I did put it on its own line originally, but I must be mistaken. Fixed in fc3cce7f0a2862631ffa28c2fff08b64155cbb5e.

snags88:
@jcmustin Oh I meant in the actual code base under the `stories` directory so that we can see what the vertical divider looks like when we go to http://radiance-ui.curology.com

jcmustin:
Closing this, as changes were made to the design that removed need for this component.

PR #62, opened 2019-03-04
Update changelog v2.0.5

Yihwan:


Yihwan:
For some reason the icons are looking too small when used; should I make any changes to the dimensions in radiance? 

screen shot 2019-03-04 at 9 36 45 am

Yihwan:
also running into a strange problem where glyphs on storybook and prod don't match. maybe it's not an issue with radiance, but the implementation in the component? 

screen shot 2019-03-04 at 9 45 21 am
screen shot 2019-03-04 at 9 45 25 am

snags88:
Probably don't need to do a PR for changelog and version bump in the future.

Also side note, you can test all these things out without having to publish the Radiance-ui version to npm. You use `yarn link` functionality https://yarnpkg.com/lang/en/docs/cli/link/

PR #66, opened 2019-03-07
Adjust immersive modal to take width param

mattam:
For the Change Plan modal in the Cancellation flow, the width has been set to 776px to match the new grid:

screen shot 2019-03-07 at 11 28 11 am

But the current immersiveModal's width is set to `${BREAKPOINTS.sm}px` which is `576px`

This PR adjusts the ImmersiveModal to allow for a `width` parameter of `small` (default) or `medium` to be passed in, with the `medium` option setting the modal to be `776px` wide.

### DIFF
--- a/src/shared-components/immersiveModal/style.js
+++ b/src/shared-components/immersiveModal/style.js
@@ -25,13 +25,14 @@ export const ModalContainer = styled.div`
   background-color: ${COLORS.white};
   height: 100vh;
   margin: 0 auto;
-  max-width: ${({width}) => (width === 'medium' ? `${BREAKPOINTS.md}px` : `${BREAKPOINTS.sm}px`)}
+  max-width: ${BREAKPOINTS.sm}px;
   overflow-y: auto;
   position: relative;
   width: 100%;
 
-  ${MEDIA_QUERIES.smUp} {
+  ${MEDIA_QUERIES.mdUp} {
     height: auto;
+    max-width: ${({maxWidth}) => (maxWidth === 'medium' ? `${BREAKPOINTS.md}px` : `${BREAKPOINTS.sm}px`)};
     margin-bottom: ${SPACING.base};
     margin-top: ${SPACING.base};
   }
--- a/src/shared-components/immersiveModal/style.js
+++ b/src/shared-components/immersiveModal/style.js
@@ -25,7 +25,7 @@ export const ModalContainer = styled.div`
   background-color: ${COLORS.white};
   height: 100vh;
   margin: 0 auto;
-  max-width: ${BREAKPOINTS.sm}px;
+  max-width: ${({width}) => (width === 'medium' ? `${BREAKPOINTS.md}px` : `${BREAKPOINTS.sm}px`)}
   overflow-y: auto;
   position: relative;
   width: 100%;
--- a/src/shared-components/immersiveModal/style.js
+++ b/src/shared-components/immersiveModal/style.js
@@ -32,7 +32,7 @@ export const ModalContainer = styled.div`
 
   ${MEDIA_QUERIES.mdUp} {
     height: auto;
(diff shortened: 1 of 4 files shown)


src/shared-components/immersiveModal/style.js
```diff
@@ -30,8 +30,9 @@ export const ModalContainer = styled.div`
   position: relative;
   width: 100%;
 
-  ${MEDIA_QUERIES.smUp} {
+  ${MEDIA_QUERIES.mdUp} {
     height: auto;
+    max-width: ${({maxWidth}) => (maxWidth === 'medium' ? `776px` : `${BREAKPOINTS.sm}px`)};
```
ZeMunchkin: `BREAKPOINTS.md` is 768px. Can we use that so we're consistent with our design system?

mattam: I did initially look at using `BREAKPOINTS.md` since it (`768`) was the closest to the value of `776`, but are the current breakpoints based off the old grid?  If that's the case, then they are currently not consistent with our newest design system.  Or are the breakpoints more for use in media queries, and not necessarily intended to be used for fitting our components onto the grid?

If it's intended for grid usage, I think it's possible we might need a followup PR to adjust the Breakpoint values to fit in with the new grid that the designers are using.  If it's not intended for grid usage, we might need a new constants file  so that we can snap our components to grid values.

But for the meanwhile, I'm thinking of setting this to the hardcoded `776px` value until we have a chance to update all the breakpoints/grids across our whole app.  (which has a chance to break a lot of things -____-)

ZeMunchkin: Hmmm. I know the breakpoints are definitely used in media queries but I don't know if they always cross to container size as well. Maybe @benkolde can weigh in? If it's a larger design system issue, then I agree it makes sense to go with hardcoding for the time being.

benkolde: @ZeMunchkin @mattam Does using the old measurement break anything with the design? The reason I say that is by limiting the hardcoding values (and risking forgetting them) we can change it all easier later

mattam: @benkolde : The `BREAKPOINTS.md` value does kinda work for the modal width (have to adjust some of the internal padding to make things fit), but 3 points:
1)  We might be overloading the usage of the Breakpoints constants for something it's not designed for (causes confusion)
2)  The `BREAKPOINTS.md` value isn't actually Grid accurate, and using it actually means that the UI won't be lined up with the grid.
3)  We could setup a task/issue to track how we want to deal with fitting container widths to the grid to ensure that we don't forget to change this.

Some background information, as far as I can tell in our UI, we don't use the Breakpoints constants for anything other than media queries except in 2 places in the radiance-UI (for the accordian and the immersiveModal component), possibly that shouldnt' be

benkolde: @mattam Okay, then let's just go with the hard coded width since that's where we're going

mattam: Did a brief discussion with the Eng team about this at Demos & Discussion: https://www.notion.so/curology/Breakpoints-Spacing-for-Grid-layout-9cd5239a83644142b3c30dba927db996

Some things from the discussion:
1)  Question around if we should revisit the Breakpoint values (which were probably based off device screen widths, so that we know when to show certain layouts) in light of the new Grid layout that we have (do we need the breakpoints adjusted so that they encompass certain grid widths better?)
2)  We do want to eventually have some way of accessing specific grid friendly/specific values, not necessarily right now, but soon as we're getting more designs aligned on the Grid.  Having some constants there would be nice.
3)  Meanwhile on the Eng side, we're currently hard coding in these large widths (Anything larger than the `Spacing` constant `xxlarge: 64px`)

So to move forwards, I would prefer to keep use the hard coded `776px` value here (since it matches the mocked design and it matches the Grid), and would like for the Radiance team to start looking at some standardized grid-aligned container widths which we could gradually roll out for the various places we set our component widths.

PR #70, opened 2019-03-12
Added optional icon to checkbox and radio button

jcmustin:
Added the option to include arbitrary icons within checkboxes and radio buttons. This shows up in dthe redesigned checkin and resubscription flows. Updated to now support multiple sizes, per master design doc from figma. Default size is small. Icons are only visible at the large size.

rdosmll
rdolrg
chksmll
chcklrg


jcmustin:
> Added the option to include arbitrary icons within checkboxes and radio buttons. This shows up in the redesigned checkin and resubscription flows. Icons have a default width and height of 16px, but these can be set manually using the `iconWidth` and `iconHeight` properties.
> 
> image
> image

Just realized the output for the added SelectorButton snapshot test is empty. (I previously tried to implement this with `renderer.create`, which would fix this issue, but I ran into issues.) I have a few things to take care of on the immediate horizon, but will fix this soon.

src/shared-components/selectorButton/index.js
```diff
@@ -21,12 +21,18 @@ const propTypes = {
...
-  onClick: () => {},
+  onClick: () => {
+  },
   type: 'primary',
   selector: 'radio',
+  iconWidth: 16,
+  iconHeight: 16,
 };
 
 const SelectorButton = ({
```
youngmanalive: Probably not a big deal but I usually try to avoid passing in capitalized props --> `Icon={...}`. 
You could reassign it in the function and skip the ternary as well for readability. Just a thought. Looks good otherwise!

```jsx
const SelectorButton = ({
  checked,
  onClick,
  children,
  type,
  selector,
  icon,
  iconWidth,
  iconHeight,
  ...rest
}) => {
  let Icon = CheckMarkIcon;
  let width = 11;
  let height = 11;

  if (icon) {
    Icon = icon;
    width = iconWidth;
    height = iconHeight;
  }

  return (
    <OuterContainer
      onClick={onClick}
      onKeyPress={onClick}
      tabIndex="0"
      selector={selector}
      {...rest}
    >
      <SelectorContainer>
        <Icon 
          css={css`
            ${selectorIcon};
            color: ${checked ? COLORS.white : 'transparent'};
          `}
          width={width}
          height={height}
        />
        <Selector type={type} checked={checked} selector={selector} />
      </SelectorContainer>

      {children && (
        <TextContainer>{children}</TextContainer>
      )}
    </OuterContainer>
  );
};
```

zbruno:
@jcmustin We have something very similar to this already in the codebase called an option button. I do see that there are slight differences between the styles, but before you merge I would like to sync with the designers about this!  Thanks

youngmanalive:
@zbruno Yeah its getting confusing. `OptionButton` is in the component-library and you just specify 'checkbox' or 'radio' as a prop. 

Radiance has `Checkbox` and `RadioButton` which both call `SelectorButton`  >>> [Issue](https://github.com/PocketDerm/radiance-ui/issues/68)

jcmustin:
> @jcmustin We have something very similar to this already in the codebase called an option button. I do see that there are slight differences between the styles, but before you merge I would like to sync with the designers about this! Thanks

@zbruno thanks for the heads up! I actually specifically added this because of mockups from design which use icons in conjunction with radiance components instead of OptionButtons. Will check in with designers one last time before finalizing though 👍

src/shared-components/selectorButton/index.js
jcmustin: @youngmanalive Same actually re: non-capitalized props; I had a recent memory of seeing a capitalized prop which made me change my mind on that, but I was actually mistaken, it was just assigned to a capitalized variable within the render method. Fixed in 23f15cd999e65e8464948b9f0efb7381f031c8e8.

jcmustin: (Ternary also obviated in that same commit)

youngmanalive: 👍

jcmustin:
> @jcmustin We have something very similar to this already in the codebase called an option button. I do see that there are slight differences between the styles, but before you merge I would like to sync with the designers about this! Thanks

@zbruno Just following up, Ben gave a vote of approval for adding icon support to radiance buttons over updating the styling of OptionButtons.

PR #77, opened 2019-03-29
New Tooltip Component

daigof:
- Complete tooltip with docs and demo with knobs
- added eslint rule to prevent error in tests files
- added prettierrc rules to standard format on editor

### DIFF
--- a/src/shared-components/tooltip/index.js
+++ b/src/shared-components/tooltip/index.js
@@ -1,0 +1,122 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import OffClickWrapper from '../offClickWrapper';
+import { MainContainer, Trigger, TooltipBox } from './style';
+
+const propTypes = {
+  alignRightPercent: PropTypes.number,
+  alignTopPercent: PropTypes.number,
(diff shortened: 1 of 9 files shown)


benkolde:
Looks good! 👍

src/shared-components/tooltip/index.js
```diff
@@ -0,0 +1,122 @@
...
+  onMouseLeave = () => {
+    this.setState({ hovered: false });
+  };
+
+  onClick = () => {
+    const { clicked } = this.state;
+
+    if (clicked) {
+      // if clicked is true, we're about to update to false so remove class
+      document.querySelector('body').classList.remove('cursor-pointer');
```
snags88: Since `cursor-pointer` css class isn't a thing, you'll want to inject it like we do here https://github.com/PocketDerm/PocketDerm/blob/28ad86f826ce599fd9171332c5c768218c274547/resources/assets/jsx/dashboard/components/shipmentRescheduler/index.js#L182-L188

daigof: right,done. Thanks for catching this up, I understand now, tested it works.

PR #79, opened 2019-04-02
New Carousel Component

daigof:
### Differences with Curology component:
- Moved the CAROUSEL_CONTAINER_SIZES  out of the Carousel component and into Radiance Constants.
- Refactor styles, cleanup code.
- Removed Mobx dependency.
- Improved Docs, added all properties to example with knobs.


src/shared-components/carousel/arrow/index.js
```diff
@@ -0,0 +1,48 @@
...
+    prev,
+    next,
+    disabled,
+    ...rest
+  } = props;
+
+  return (
+    <ArrowContainer
+      {...rest}
+      style={{ ...style }}
```
snags88: were we using this `style` prop before? We typically use the `css` prop for emotion and never the `style` prop.

src/shared-components/carousel/arrow/index.js
```diff
@@ -0,0 +1,48 @@
...
+      onClick={() => {
```
snags88: We should probably just turn this into a class component and define an `onClick` handler for the class that is equal to this function definition so that we can avoid defining an anonymous function here.

src/shared-components/carousel/arrow/style.js
```diff
@@ -0,0 +1,27 @@
+import styled from '@emotion/styled';
+
+import { SPACING, COLORS, ANIMATION } from 'src/constants';
+
+const ArrowContainer = styled.div`
+  cursor: pointer;
+  position: absolute;
+  top: 50%;
+  z-index: 2000;
```
snags88: could we try to use one of the z-index constants here?

yarn.lock
```diff
@@ -6735,6 +6747,11 @@ mkdirp@^0.5.0, mkdirp@^0.5.1, mkdirp@~0.5.0, mkdirp@~0.5.1:
   dependencies:
     minimist "0.0.8"
 
+mobx@^5.9.4:
```
snags88: are we using mobx still?

src/shared-components/carousel/arrow/index.js
daigof: I dont think we even need that, I just removed the prop and still working fine

src/shared-components/carousel/arrow/index.js
daigof: wow, it turns out we neednt any of those props (only onClick should be passed as part of specs) I refactored and cleanup this component. I also fixed a bug where if you click the arrow space once it disappear (opacity:0) it throws an error now I handle it with display:none

src/shared-components/carousel/arrow/style.js
daigof: you know what I actually checked the zScale constant before and I thought z-index:20 may not be sufficient ?? the other next value is modal:9999 which is a no go.
Should we create a new value or leave at 20?

yarn.lock
daigof: no, those are leftover from a previous commit with mobx

docs/carousel.md
```diff
@@ -0,0 +1,35 @@
+# Carousel
+### Usage
+
+```jsx
+import { Carousel } from 'radiance-ui';
+import { CAROUSEL_CONTAINER_SIZES } from 'radiance-ui/lib/constants';
+
+<div style={{maxWidth: CAROUSEL_CONTAINER_SIZES.threeCards}}>
```
youngmanalive: `css` prop here.

youngmanalive:
I'm wondering if the `CAROUSEL_CONTAINER_SIZES` should be built into the component with the width passed in as prop.

e.g. `<Carousel width='oneCard' .... />` . 

If a container is always required, it should be baked into the Carousel. Otherwise devs need to wrap it every time and import the constants. @snags88 what do you think?

src/shared-components/carousel/arrow/style.js
snags88: I guess we can create a new constant for now so that we at least have it in 1 place. Then if we ever straighten these z-index values out, it'll be easy for us to set it up

snags88:
@youngmanalive that's not a bad idea. It seems like we only have 3 variations so it could also be unrelated to the width. We could just pass in "cardCount" or something like that and we should be able to figure out the width from the constants from there

src/shared-components/carousel/arrow/style.js
daigof: done

daigof:
- Refactored to include the Width variation with the same numCards property.
- Fix tests
- Improved docs

src/shared-components/carousel/style.js
```diff
@@ -156,7 +156,32 @@ const reactSlickStyles = css`
   }
 `;
 
-export const CarouselContainer = styled.div`
+const parseOuteStyle = numCardsVisible => {
```
snags88: *parseOuterStyle

PR #82, opened 2019-04-04
New Form components

daigof:
- I included all the changes requested in https://trello.com/c/WNGdAWOk/73-convert-form-component
- I tried to leave the main structure/naming as they were so there is less impact for the Curology site refactor
- However, these component may still need to be refactored, I need your input on the current implementation. Just be aware the refactor for Curology may get more complex


docs/forms/bulkErrors.md
```diff
@@ -0,0 +1,21 @@
+# BulkErrors Component
+### Usage
+
+```jsx
+import { BulkErrors, baseInputStyles } from 'radiance-ui';
+    
+const errors = {
+  required: 'This field is required',
+  maxLenght: 'Must be at least 8 characters',
```
snags88: *maxLength

docs/forms/bulkErrors.md
```diff
@@ -0,0 +1,21 @@
...
+};
+
+<input type="text" css={baseInputStyles} />
```
snags88: Probably want to include `React.Fragment` around the two react elements

docs/forms/bulkErrors.md
```diff
@@ -0,0 +1,21 @@
...
+<BulkErrors errors={errors} />
+```
+
+<!-- STORY -->
+
+### Proptypes
+| prop                | propType   | required | default      | description                                                                                                                  
+|---------------------|------------|----------|--------------|---------------|
+| errors              | object     | yes      | {}           | errors object |
```
snags88: What shape should the error object be in? Any key name => error message?

docs/forms/formSection.md
```diff
@@ -0,0 +1,78 @@
+# FormSection Component
+### Usage
+
+```jsx
+import { FormSection, inputStyles, baseInputStyles } from 'radiance-ui';
```
snags88: is this how we were implementing it in PocketDerm? It feels like we should just create a component for `<Input />` and use that as a way to wrap styling for the input field.

src/shared-components/forms/index.js
```diff
@@ -0,0 +1,58 @@
...
+import Typography from '../typography';
+import HelperText, { ErrorItem, HintItem } from './helperText';
+import { FormContainer, InputContainer } from './style';
+import ErrorIcon from '../../svgs/icons/error-icon.svg';
+
+const FormSection = ({
+  children,
+  displayValidation,
+  errorMessage,
+  isFocused,
```
snags88: i wonder if there's a way to get this value without having the user pass it in from outside of the component

src/shared-components/forms/index.js
```diff
@@ -0,0 +1,58 @@
...
+  displayValidation: PropTypes.bool,
+  errorMessage: PropTypes.string,
+  isValid: PropTypes.bool,
+  label: PropTypes.string,
+  labelFor: PropTypes.string,
+  isFocused: PropTypes.bool,
+  hint: PropTypes.string,
+};
+
+export { baseInputStyles, inputStyles, Textarea } from './style';
```
snags88: we should use dot notation to export stuff out.

I imagine we can do something where we export a `Form` object that has `<Form.Section />`, `<Form.Textarea />`, `<Form.Input />`

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,99 @@
+import styled from '@emotion/styled';
```
snags88: there are places in this file that we could use constants like spacing and transition timing

daigof:
@snags88 I am not happy with the current implementation either, I think I spent more time trying to migrate to Radiance without causing impact in PocketDerm than if I had to rewrite the Form components from the scratch.

Would you agree if I try to refactor and make it clean and elegant these Form components in Radiance according to Figma designs and only then after approved we can think on refactor in PocketDerm?

@youngmanalive fyi

snags88:
@daigof so your plan is to get these Form components into Radiance, convert PocketDerm so we get that part over with. And then follow up with a refactor of these Form components?

daigof:
@snags88 no, my plan is to get it right into radiance first. Publish a polished version. only then look at how we incorporate into Pocketderm

snags88:
ah gotcha. I actually agree with that approach!

youngmanalive:
Definitely think this is a good opportunity to get a clean version into radiance and then implement into pocketderm later. @snags88 - I was thinking same thing on the dot notation like Form.Textarea, etc. @daigof nice work so far and solid game plan!

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  ${({ shouldRenderError }) => shouldRenderError && errorIconStyles};
+`;
+
+export const baseInputStyles = css`
+  appearance: none;
+  border: 1px solid ${COLORS.border};
+  border-radius: 0;
+  color: ${COLORS.primary};
```
benkolde: body text should always be `purple85`

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,99 @@
...
+
+  &::placeholder {
+    color: ${COLORS.primaryTint3};
+  }
+`;
+
+export const inputStyles = (showError, focusStyle) => {
+  const borderColor = showError ? COLORS.statusRed : false;
+  const focusBorderColor =
+    focusStyle === 'focusSecondary' ? COLORS.secondary : COLORS.primary;
```
benkolde: focus style should be `borderColor: COLORS.primary` and should include a box shadow under the `focusSecondary` constant

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const Textarea = styled.textarea`
+  ${TYPOGRAPHY_STYLE.body};
+  appearance: none;
+  background: ${COLORS.white};
+  border: 1px solid ${COLORS.border};
+  border-radius: 0;
+  box-shadow: 0 1px 0 0 ${COLORS.tertiary};
+  color: ${COLORS.primary};
```
benkolde: `purple85`

src/shared-components/forms/style.js
daigof: Ok. Note that I refactored this Form component to a more clean simpler to use version

src/shared-components/forms/style.js
daigof: @benkolde can you review the focus styles. I added the box-shadow:focusSecondary but it looks very bright noticeable (not subtle?)

daigof: should be here https://curology-radiance-pr-82.herokuapp.com

daigof:
@snags88 Pending, your thoughts?:
- Incorporate BulkError into Field directly so instead of errorMessage we just have a single error object: if you need 1 error then 1 obj with 1 field, if you need more, more field in error obj, etc. (remove errorMessage prop)
- Update react-transition-group to latest version: HelperText is using v1 which is not compatible with latest (v2) need to rework it a little, also I am not fond on injecting GlobalStyles, will try to get rid of it

src/shared-components/forms/style.js
daigof: I put primaryTint1 which in turn will be purple85 once colors branch is merged

benkolde:
@daigof 

It looks like theres an extra border or something being added to the bottom of this text area

image

benkolde:
Other than that, this looks good. So feel free to merge when that's fixed!

daigof:
@snags88 this is ready for review again.
- removed Bulk errors, included in the Field component as errors obj. 
- migrate react-transition lib to latest version
- update docs and examples

benkolde:
@daigof can we change the color of the disabled label to `purple30`? just to further reinforce that it's disabled.

src/shared-components/forms/index.js
```diff
@@ -0,0 +1,95 @@
...
+          <ErrorIcon />
+
+          {React.cloneElement(inputChild, {
+            disabled,
+          })}
+
+          <TransitionGroup component={HelperList}>
+            {!!hintMessage && (
+              <HelperTransition key="hint-message">
+                <HintItem>{hintMessage}</HintItem>
```
youngmanalive: Does `HintItem` need to be wrapped in `HelperTransition` here? Emotion is giving it height and opacity on focus but doesn't seem like it will be entering/exiting like the errors will.

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,124 @@
...
+
+    &:active,
+    &:focus {
+      border-color: ${COLORS.statusRed};
+      box-shadow: none;
+    }
+  }
+`;
+
+export const InputContainer = styled.div`
```
youngmanalive: Should `position: relative` be in here whether the svg is displayed or not? Might be nice to have a transition on the svg as well to match the transitioning errors and border.

src/shared-components/forms/index.js
```diff
@@ -0,0 +1,95 @@
...
+          <Label htmlFor={htmlFor} disabled={disabled}>
+            {label}
+          </Label>
+        )}
+
+        <InputContainer showErrors={showErrors}>
+          <ErrorIcon />
+
+          {React.cloneElement(inputChild, {
+            disabled,
```
youngmanalive: Only concern here is if devs pass in more than one child. Not likely, but if they do it might break without them knowing why. If `cloneElement` throws an error stating you can only use one child, then I guess not a big deal.

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,124 @@
+import styled from '@emotion/styled';
+import { css } from '@emotion/core';
+
+import { style as TYPOGRAPHY_STYLE } from '../typography';
+import { COLORS, BOX_SHADOWS, SPACING, ANIMATION } from '../../constants';
+
+export const HelperList = styled.ul`
+  list-style-type: none;
+  margin: 8px 0;
```
snags88: 8px should be one of our spacing constants

src/shared-components/forms/style.js
```diff
@@ -0,0 +1,124 @@
...
+const errorStyles = css`
+  position: relative;
+
+  svg {
+    display: block;
+    width: 16px;
+    height: 16px;
+    position: absolute;
+    margin-top: 20px;
+    right: 16px;
```
snags88: 16px should be a spacing constant

stories/forms/index.js
```diff
@@ -0,0 +1,71 @@
...
+const MainContainer = styled.div`
+  text-align: left;
+`;
+
+const FormContainer = styled.div`
+  margin: 1rem 0 2rem 0;
+  width: 350px;
+`;
+
+const stories = storiesOf('Forms', module);
```
snags88: update to `Field` instead of `Forms`

src/shared-components/forms/index.js
daigof: yes it does enter/leave on focus and the animation does trigger. this is another <li> item like the errors

src/shared-components/forms/style.js
daigof: that's actually a good idea let me see if I can put the animation on `fill`

src/shared-components/forms/index.js
daigof: I actually researched the PropTypes docs for this, instaead of use `node` like everywhere else I am using `element` which force you to put only 1 child form the docs
https://reactjs.org/docs/typechecking-with-proptypes.html

```
Requiring Single Child
With PropTypes.element you can specify that only a single child can be passed to a component as children.
```

src/shared-components/forms/style.js
daigof: done

src/shared-components/forms/style.js
daigof: fixed

src/shared-components/forms/style.js
daigof: - I understand that SPACING constant are used for padding/margin/gaps. 
- Dimension values may either use other constant (ICON_SIZE maybe?) or left untouched?
- there are special cases where we need to pixel perfect accomodate the icon (like top:20px above)

Having said that I did these changes:
- changed margin-top to top:20px (this is special case pixel perfect)
- width and height left untouched with 16px
- changed right:${SPACING.small}

stories/forms/index.js
daigof: done

src/shared-components/forms/index.js
youngmanalive: 👍

src/shared-components/forms/style.js
youngmanalive: 👍

src/shared-components/forms/index.js
youngmanalive: 👍

src/shared-components/forms/style.js
snags88: yup that sounds like exactly what I had in mind. Thanks!

PR #84, opened 2019-04-05
Colors adjustments

daigof:
- task: https://trello.com/c/mSqJr1Ch/89-update-colors
- I tested the deprecation warning and works, but we may need to review the message: `[Deprecation Warning]: purple60 will be deprecated in v2. Use purple70 instead`
is it correct?

### DIFF
--- a/src/constants/colors/index.js
+++ b/src/constants/colors/index.js
@@ -12,7 +12,9 @@ export const brandColors = {
   // Primary 1
   purple100: '#332e54',
   purple85: '#524D6E',
+  purple80: '#5C5876',
   purple70: DEFAULT,
+  purple60: '#858298',
   purple30: '#c3c0cd',
   purple15: DEFAULT_BORDER,
   purple10: DEFAULT_BACKGROUND,
@@ -58,8 +60,8 @@ export const legacyColors = {
   orange: '#fc7b65',
   teal: '#d5e9e5',
   yellowLight: '#f4f2b0',
-  purple80: '#5C5876',
-  purple60: '#858298',
+  purple80: brandColors.purple80,
+  purple60: brandColors.purple60,
 };
 
 export const colorAliases = {
--- a/src/constants/colors/index.js
+++ b/src/constants/colors/index.js
@@ -2,6 +2,7 @@
  * Note: Each config is exported separately to be rendered in storybook:
  * https://github.com/PocketDerm/PocketDerm/blob/master/resources/assets/component-library/stories/colors/index.js#L5
  */
+import { withDeprecationWarning } from '../../utils';
 
 const DEFAULT = '#706D87';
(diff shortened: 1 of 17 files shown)


src/constants/colors/index.js
```diff
@@ -132,10 +151,16 @@ export const guideColors = {
...
-);
+const colorsCompilation = {
+  ...brandColors,
+  ...colorAliases,
+  ...postcardColors,
+  ...guideColors,
+};
+
+const deprecatedProperties = {
+  purple80: 'purple80 will be deprecated in v2. Use purple85 instead',
```
snags88: It should probably say `COLORS.purple80 will be deprecated in v3. Use COLORS.purple85 instead.`

src/constants/colors/index.js
```diff
@@ -2,15 +2,22 @@
...
+
+const DEFAULT = '#706D87';
+const DEFAULT_BACKGROUND = '#ededf0';
+const DEFAULT_BORDER = '#e1e0e6';
 
 export const brandColors = {
   // Primary 1
   purple100: '#332e54',
+  purple85: '#524D6E',
   purple80: '#5C5876',
```
snags88: you probably want to move this and purple 60 directly into the `legacyColors` object rather than referencing this `brandColors.purple80` from the legacy colors.

src/constants/colors/index.js
```diff
@@ -132,10 +151,16 @@ export const guideColors = {
...
 
-export default Object.assign(
-  {},
-  brandColors,
-  colorAliases,
-  postcardColors,
-  guideColors
-);
+const colorsCompilation = {
+  ...brandColors,
```
snags88: you can add `...legacyColors` for now so that we can still have purple80/60 exposed as our color constants

src/constants/colors/index.js
daigof: I understand deprecation as discouragement to use, So we are effectively deprecating now in v2 and remove in v3. or we will deprecate in v3 and remove in v4?

src/constants/colors/index.js
daigof: I did that but later realized legacyColor was not exported.

src/constants/colors/index.js
daigof: ok fixed

src/constants/colors/index.js
snags88: we can deprecate now (v2) and remove in v3

daigof:
> Everything looks good besides one thing, and I may just be missing it but it looks like the typography constants aren't updating. Specifically the error and success text. But I can't tell if body and caption are updated either.

Good catch, error color was `${COLORS.accent}`  I fixed it, success and the other have the right color as far I can tell form the code

PR #85, opened 2019-04-09
Tabs Component initial

daigof:
- Similar to Form component this is the initial version compatible with PocketDerm. However:
- I dont like how it handle the Tabs as object {id,title}, worse it couples the initialTab to Id=1 forcing us to provide at least one Tab object with Id=1
- I propose drop the tab objects and just work with arrays of string titles and the active Id would be the index now. As usual this would require more refactor in PocketDerm as well.

### DIFF
--- a/docs/tabs.md
+++ b/docs/tabs.md
@@ -4,54 +4,23 @@
 ```jsx
 import { Tabs } from 'radiance-ui';
 
-const Content = styled.div`
-  ${TYPOGRAPHY_STYLE.display};
-  display: flex;
-  justify-content: center;
-  align-items: center;
-
-  background-color: ${props => props.bgColor};
-  color: ${COLORS.primary};
-  height: calc(100vh / 2);
-`;
-
-class TabsExample extends React.Component {
-  state = {
-    activeTabId: 1,
-  };
-
-  tabItems = [
-    { id: 1, text: 'Tab 1', bgColor: COLORS.postcardGreen },
-    { id: 2, text: 'Tab 2', bgColor: COLORS.postcardPeach },
-    { id: 3, text: 'Tab 3', bgColor: COLORS.postcardYellow },
+const TabsExample = () => {
+  const tabItems = [
+    { id: 1, text: 'Tab 1' },
+    { id: 2, text: 'Tab 2' },
+    { id: 3, text: 'Tab 3' },
   ];
 
-  onClickTab = tab => {
-    this.setState({ activeTabId: tab.id });
+  const onClickTab = tab => {
+    console.log(tab);
   };
 
-  renderContent = () => {
-    const { activeTabId } = this.state;
-    const tab = this.tabItems.find(tabItem => tabItem.id === activeTabId);
-
-    return <Content bgColor={tab.bgColor}>{tab.text}</Content>;
-  };
-
-  render() {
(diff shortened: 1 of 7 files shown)


benkolde:
@daigof 

You can totally refuse this but on mobile, tabs become left aligned with the first string having a margin of `24px` from the left side of the window. It would be nice to bake this into the component if possible.

daigof:
> You can totally refuse this but on mobile, tabs become left aligned with the first string having a margin of `24px` from the left side of the window. It would be nice to bake this into the component if possible.

@benkolde Yeah absolutely, I refactor all the styles a little bit. Note that the Tab Item already had a padding left/right: 24px so I didnt add the extra margin-left,  means that the Tab first letter will be exactly 24px to the left of the Tabs component container

docs/tabs.md
```diff
@@ -0,0 +1,71 @@
...
+    return <Content bgColor={tab.bgColor}>{tab.text}</Content>;
+  };
+
+  render() {
+    const { activeTabId } = this.state;
+
+    return (
+      <section>
+        <Tabs
+          activeTabId={activeTabId}
```
snags88: Should this be `initialActiveTabId`?

docs/tabs.md
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  tabItems = [
+    { id: 1, text: 'Tab 1', bgColor: COLORS.postcardGreen },
+    { id: 2, text: 'Tab 2', bgColor: COLORS.postcardPeach },
+    { id: 3, text: 'Tab 3', bgColor: COLORS.postcardYellow },
+  ];
+
+  onClickTab = tab => {
+    this.setState({ activeTabId: tab.id });
```
snags88: it looks like from the implementation, the `<Tabs />` component handles the state for which tab is currently active. Having this in the example is a bit misleading. Maybe we can just console.log out the `tab` variable and remove having `activeTabId` as state on this component

src/shared-components/tabs/test.js
```diff
@@ -0,0 +1,51 @@
...
+import Tabs from './index';
+
+describe('<Tabs />', () => {
+  describe('default props', () => {
+    test('renders the correct props', () => {
+      const wrapper = shallow(
+        <Tabs
+          activeTabId={1}
+          tabItems={[
+            { id: 1, text: 'Rain Drop' },
```
snags88: lol we can probably update the text to be more normal 😆

docs/tabs.md
daigof: lol yeah havent noticed that bug

docs/tabs.md
daigof: done

src/shared-components/tabs/test.js
daigof: done

daigof:
@snags88 what do you think regarding flattening the tabItems structure to be just an array of strings (titles) and handle tabId as array index. Having refactored the Tabs Example I realized the tabItems object structure may be due that the objects contains part of the content displayed in page but I still think this should be decoupled.

snags88:
@daigof i think we can keep as objects for now. I briefly looked at how we use it for pocketderm and it seems like we use the object notation to help with setting the query string.

PR #86, opened 2019-04-10
New bulk errors components

daigof:
- Added Bulk error component
- Refactor Field to use the new BulkeError

### DIFF
--- a/src/shared-components/bulkErrors/index.js
+++ b/src/shared-components/bulkErrors/index.js
@@ -3,7 +3,7 @@ import React from 'react';
 import { TransitionGroup } from 'react-transition-group';
 
 import { CenteredErrorList, ErrorList, ErrorItem } from './style';
-import HelperTransition from '../field/helperTransition';
+import HelperTransition from '../../utils/helperTransition';
 import formatError from './formatError';
 
 const BulkErrors = ({ errors, centered }) => {
--- a/src/shared-components/bulkErrors/index.js
+++ b/src/shared-components/bulkErrors/index.js
@@ -1,0 +1,37 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { TransitionGroup } from 'react-transition-group';
+
+import { CenteredErrorList, ErrorList, ErrorItem } from './style';
+import HelperTransition from '../field/helperTransition';
+import formatError from './formatError';
+
+const BulkErrors = ({ errors, centered }) => {
+  const errorKeys = Object.keys(errors);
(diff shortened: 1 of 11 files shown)


src/shared-components/bulkErrors/index.js
```diff
@@ -0,0 +1,37 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { TransitionGroup } from 'react-transition-group';
+
+import { CenteredErrorList, ErrorList, ErrorItem } from './style';
+import HelperTransition from '../field/helperTransition';
```
youngmanalive: maybe move this guy into its own directory? might share this with other components down the line. not necessary tho.

src/shared-components/field/index.js
```diff
@@ -72,20 +69,14 @@ class Field extends React.Component {
           {React.cloneElement(inputChild, {
             disabled,
           })}
-
-          <TransitionGroup component={HelperList}>
-            {!!hintMessage && (
+          {!!hintMessage && (
+            <TransitionGroup component={HintList}>
```
youngmanalive: field

looks like the hint is acting up. I feel like we're mixing transitions with emotion and transitiongroup on this. The errors are being added/removed as they are validated, but if there is a hint message it mounts on the DOM and stays there, only the styling changes on focus.

youngmanalive: https://github.com/PocketDerm/radiance-ui/blob/master/src/shared-components/field/style.js#L46
^^ I think if we added max-height property here you could strip out the TransitionGroup and HelperTransition all together for the hint item

src/shared-components/bulkErrors/index.js
daigof: it is technically shared among 2 components: Bulk and Field. it wasnt copupled in any way to Field so I moved to utils folder. good advice

src/shared-components/field/index.js
daigof: that actually troubled my a lot. Since I wanted to reuse BulkErrors inside the Field I had to create 2 separate lists. Problem is Ben wanted to not show the push down layout that happens when you add messages at the bottom so thats why I kept a min-height at all times on both lists. You re right I actually didnt need Helpertransition for the Hint so I refactored it and simplified that HintItem.

I checked the code in CUrology and Hint Message is not used at all. So I just removed the min-heihgt from the Hint. Only the error list will have the min-heihgt that prevents the first error to push down the layout which is 90% of the cases.

PR #88, opened 2019-04-12
New ProgressBar component

daigof:
- Refactored/CleanedUp version of what we found in `jsx/components` in PocketDerm

### DIFF
--- a/src/shared-components/progressBar/index.js
+++ b/src/shared-components/progressBar/index.js
@@ -1,0 +1,53 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+
+import { COLORS, PROGRESS_BAR_STATUS } from '../../constants';
+import { OuterContainer, InnerBar } from './style';
+
+const ProgressBar = ({
+  backgroundColor,
+  barColor,
+  customStyles,
+  height,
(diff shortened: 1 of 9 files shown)


src/shared-components/progressBar/index.js
```diff
@@ -0,0 +1,53 @@
...
+import PropTypes from 'prop-types';
+import { css } from '@emotion/core';
+
+import { COLORS, PROGRESS_BAR_STATUS } from '../../constants';
+import { OuterContainer, InnerBar } from './style';
+
+const ProgressBar = ({
+  backgroundColor,
+  barColor,
+  customStyles,
```
snags88: Instead of using a prop named `customStyles`, we can have people use `ProgressBar` like:

```jsx
<ProgressBar css={css`display: flex;`} />
```

and then just pass along `...rest` to the `OuterContainer` in this component definition:
```jsx
const ProgressBar = ({ backgroundColor, barColor, height, loadingTime, status, ...rest }) => (
  <OuterContainer
    status={status}
    height={height}
    backgroundColor={backgroundColor}
    {...rest}
  />
)
```

daigof: yep. done

PR #89, opened 2019-04-15
New optionButton component

daigof:
- New component based on 
- https://trello.com/c/8aMso8zr/74-convert-option-button-component
- https://trello.com/c/gbkdYruo/88-update-option-buttons-to-new-specs
- compatible with pocketderm version . but completely redesigned

### DIFF
--- a/src/shared-components/optionButton/index.js
+++ b/src/shared-components/optionButton/index.js
@@ -1,8 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import Container from '../container';
-import { FlexContainer, Text, IconWrapper, CheckmarkWrapper } from './style';
+import {
+  ClickableContainer,
+  FlexContainer,
+  Text,
+  IconWrapper,
+  CheckmarkWrapper,
+} from './style';
 import CheckmarkIcon from '../../svgs/icons/checkmark-icon.svg';
 
 const OptionButton = ({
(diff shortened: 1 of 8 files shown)


benkolde:
@daigof 

A `secondary` checkbox and radio button just means that the selected state fill color is the secondary color. Border and text color should stay at `primary` and `purple85` respectively

benkolde:
@daigof It looks like the text is dark here, please change it to `purple85` if it's not already

benkolde:
@daigof Please add the a focus state to these option buttons using `BOX_SHADOW.focusSecondary`

benkolde:
@daigof 

Just double checking that the hover state matches `clickableHover`

src/shared-components/optionButton/index.js
```diff
@@ -0,0 +1,56 @@
...
+import { FlexContainer, Text, IconWrapper, CheckmarkWrapper } from './style';
+import CheckmarkIcon from '../../svgs/icons/checkmark-icon.svg';
+
+const OptionButton = ({
+  icon,
+  text,
+  onClick,
+  optionType,
+  selected,
+  styleType,
```
snags88: we probably want to get `...rest` and pass it along to one of these components that makes sense. Maybe the `Container`. One use-case is if we want to pass along any `data-*` attribute. We use that often in our integration tests

daigof:
@benkolde ,
- for clickable hover states and padding, I just used the Container component type="clickable" I though that would cover all our standard spacing and states but since we are customizing much more I will just remove it and use my personal styled div with your requirements. 

- I'll fix the font colors. they are all primary right now.


- I noticed `focusSecondary` in the requirements, although I am not exactly sure when this would activate since this is not an input text and we already have a hover state style. I wouldn't know how to activate this state since this is a custom component with no real input tag. The only way I think a user can focus this component is if I put a hidden input type checkbox and when user navigate the fields using Tab key. or am I missing something regarding foucs? @snags88

src/shared-components/optionButton/index.js
daigof: done

daigof:
> @benkolde ,
> 
> * for clickable hover states and padding, I just used the Container component type="clickable" I though that would cover all our standard spacing and states but since we are customizing much more I will just remove it and use my personal styled div with your requirements.
> * I'll fix the font colors. they are all primary right now.
> * I noticed `focusSecondary` in the requirements, although I am not exactly sure when this would activate since this is not an input text and we already have a hover state style. I wouldn't know how to activate this state since this is a custom component with no real input tag. The only way I think a user can focus this component is if I put a hidden input type checkbox and when user navigate the fields using Tab key. or am I missing something regarding foucs? @snags88

Done. I kept the Container type="clickable" since it contained the correct styles. I replaced the inner Container.Section  for the custom padding. I still dont know what to do with focus state but the rest should be good

snags88:
@daigof yup you can do focus state on radio/checkboxes. You can add a `tabIndex` attribute to the component and you can tab on to it. Similar thing done for radio/checkbox components [here](https://github.com/PocketDerm/radiance-ui/blob/4995bde5718058c5eda391e209007f20b010aa21/src/shared-components/selectorButton/index.js).

daigof:
@benkolde  added the last Focus styles.   I had no idea non form element can be focused.

PR #92, opened 2019-04-24
new Toggle component

daigof:
- redesigned according to https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance.UI-(Master)?node-id=1%3A9

- I also added the click handler to the label


src/shared-components/toggle/style.js
```diff
@@ -0,0 +1,42 @@
...
+  margin-right: ${SPACING.xsmall};
+`;
+
+export const trackStyle = {
+  borderRadius: 100,
+  height: 24,
+  width: 40,
+};
+
+export const thumbStyle = {
```
benkolde: Can we add a border of `1px` and make it `COLORS.border` to the thumb

benkolde: Also looks like the thumb needs to be nudged over 1px when it's toggled on and toggled off.

benkolde: And make the background color of the thumb, white.

src/shared-components/toggle/style.js
```diff
@@ -0,0 +1,42 @@
...
+  margin: 0;
+  font-size: ${SPACING.small};
+  line-height: ${SPACING.base};
+  text-align: left;
+  cursor: pointer;
+  user-select: none;
+  margin-right: ${SPACING.xsmall};
+`;
+
+export const trackStyle = {
```
benkolde: Can we make the track background color `purple15`? That's on me. It wasn't included in the design mocks but I think it will be better. It also matches our disabled button styling.

daigof: done

src/shared-components/toggle/style.js
daigof: done

daigof: I moved it closer to edge 1 pixel when active, moving 1 more pixel closer to each edge started to feel odd for me, take a look this is one more pixel to each side

image

daigof: current
image

benkolde: Okay let's go with whatever is marked Current. I don't know what I was seeing

PR #94, opened 2019-04-25
Box Shadow Examples

daigof:
- It made sense to add this to Constants section like Colors examples
- I also refactored/cleaned up the Colors Story


snags88:
Could we do a similar thing as Colors where we categorize the box shadow examples by context (e.g. legacyColors, brandColors, etc.). I think for box shadows, we have few different ones like Base config, alias config etc

daigof:
@snags88 yep done

snags88:
@daigof i'm not seeing the values or the box shadow on the story now:
Screen Shot 2019-04-25 at 1 50 18 PM

daigof:
ughh sorry, fixed

PR #96, opened 2019-04-29
Added an offClickWrapper Story and docs

daigof:


stories/offClickWrapper/offClickWrapperExample.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import { OffClickWrapper, Container } from 'src/shared-components';
+
+class OffClickWrapperExample extends React.Component {
+  state = {
+    text: '   Click outside this container to call the onOffClick handler...',
```
snags88: Not that important, but was there a reason for the spaces before "Click"?

daigof: must have been a copy paste from the doc, I changed that line a few times during development. Will fix and push and merge

PR #105, opened 2019-05-13
Add Deprecation warning for SPACING

daigof:
- also fixed a map key prop warning in constants story.

### DIFF
--- a/src/utils/withDeprecationWarning/index.js
+++ b/src/utils/withDeprecationWarning/index.js
@@ -1,8 +1,15 @@
+const allowedEnvironments = ['test', 'development', 'dev'];
+const conditional1 = allowedEnvironments.includes(process.env.NODE_ENV);
+
 export default function withDeprecationWarning(obj, deprecatedProperties = {}) {
   const handler = {
     get(target, property) {
-      if (Object.keys(deprecatedProperties).includes(property)) {
-        console.warn(`[Deprecation Warning]: ${deprecatedProperties[property]}`);
+      const conditional2 = Object.keys(deprecatedProperties).includes(property);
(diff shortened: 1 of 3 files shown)


snags88:
Oh actually 1 thing. Could we look at the deprecation warning util and only display that for non-production environment? Not sure how React does it, but it's pretty nice when they only show warnings/deprecation messages in dev

daigof:
> Oh actually 1 thing. Could we look at the deprecation warning util and only display that for non-production environment? Not sure how React does it, but it's pretty nice when they only show warnings/deprecation messages in dev

Added a conditional to display only in test/development environments

src/utils/withDeprecationWarning/index.js
```diff
@@ -1,8 +1,15 @@
+const allowedEnvironments = ['test', 'development', 'dev'];
+const conditional1 = allowedEnvironments.includes(process.env.NODE_ENV);
```
snags88: could we rename this variable to something more descriptive like `shouldShowForEnvironment`

src/utils/withDeprecationWarning/index.js
```diff
@@ -1,8 +1,15 @@
...
+
 export default function withDeprecationWarning(obj, deprecatedProperties = {}) {
   const handler = {
     get(target, property) {
-      if (Object.keys(deprecatedProperties).includes(property)) {
-        console.warn(`[Deprecation Warning]: ${deprecatedProperties[property]}`);
+      const conditional2 = Object.keys(deprecatedProperties).includes(property);
```
snags88: also should update this variable name to `isDeprecatedProperty`

PR #106, opened 2019-05-20
added `content` prop type to alert component

omar-b-c:
`content` prop type is PropTypes.node so this component can accept any kind of content. This is implemented because of need to pass links inside alert component.
added custom prop check to make sure one of the props(text/content) is passed to component, this can be removed after text deprecation
added deprecationWarning for Alert component
updated alert.md with `content` prop
updated storied/alert with sample of alert component with `content` prop

Screen Shot 2019-05-20 at 8 07 21 PM

### DIFF
--- a/docs/alert.md
+++ b/docs/alert.md
@@ -3,7 +3,11 @@
 
 ```jsx
 import { Alert } from 'radiance-ui';
-const SampleContentComponent = () => <div>Click <a href='https://google.com'>here</a> to go to Google!</div>;
+const SampleContentComponent = () => (
+  <div>
+    Click <a href="https://google.com">here</a> to go to Google!
+  </div>
+);
 
 <Alert.Container>
   <Alert
@@ -31,7 +35,7 @@ const SampleContentComponent = () => <div>Click <a href='https://google.com'>her
     onExit={() => {}}
   />
   <Alert
-	content= <SampleContentComponent/>
+    content={<SampleContentComponent />}
     type="success"
     duration="sticky"
     onExit={() => {}}
--- a/docs/alert.md
+++ b/docs/alert.md
@@ -3,6 +3,7 @@
 
 ```jsx
 import { Alert } from 'radiance-ui';
+const SampleContentComponent = () => (<div>Your info is updated!  <a href='http://google.com'> Click here to see changes</a></div>);
 
 <Alert.Container>
   <Alert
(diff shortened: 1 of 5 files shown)


omar-b-c:
> Is it really required to deprecate the `text` prop? If so then we should remove all examples with text in the story and just have it in the props definitions saying it's deprecated right?

I think it is unnecessary to keep both props, basically they act same. Content prop gives us more flexibility in this case.  If we all agree, i'll change all text occurrences in stories with content prop.

docs/alert.md
```diff
@@ -3,6 +3,7 @@
 
 ```jsx
 import { Alert } from 'radiance-ui';
+const SampleContentComponent = () => (<div>Your info is updated!  <a href='http://google.com'> Click here to see changes</a></div>);
```
snags88: we could just make this into a real example like `<React.Fragment>Click <a href='https://google.com'>here</a> to go to Google!</React.Fragment>`

docs/alert.md
```diff
@@ -29,18 +30,25 @@ import { Alert } from 'radiance-ui';
     duration="sticky"
     onExit={() => {}}
   />
+  <Alert
+	content=<SampleContentComponent/>
```
snags88: spacing

docs/alert.md
```diff
@@ -29,18 +30,25 @@ import { Alert } from 'radiance-ui';
...
-| onExit   | func               | yes      | -       | callback function called on dismissal of  the alert. The function will receive all  of the component's props as the argument |
+| prop     | propType                                | required                    | default | description                                                                                                                  |
+|----------|-----------------------------------------|-----------------------------|---------|------------------------------------------------------------------------------------------------------------------------------|
+| text     | string         				         | yes if content prop is empty| -       | alert text to be displayed. Will be deprecated in next major version                                                                                                  |
+| content  | numbers, strings, elements or an array  | yes if text prop is empty   | -       | renders anything passed to component. In future changes alert text.                                                          |
```
snags88: What does "In future changes alert text." mean?

docs/alert.md
omar-b-c: just used div so i don't need to add {" "} whitespace

docs/alert.md
```diff
@@ -3,35 +3,35 @@
...
   <Alert
-    text="Oops! We couldn't update your information!"
+    content="Oops! We couldn't update your information!"
     type="danger"
     duration="sticky"
     onExit={() => {}}
   />
   <Alert
-	content=<SampleContentComponent/>
+	content= <SampleContentComponent/>
```
snags88: still getting some weird spacing here. Maybe your ESlint is not configured properly? Also, doesn't it need curly braces `content={}`

stories/alert/index.js
```diff
@@ -16,31 +16,31 @@ stories.add(
...
         <Alert
-          text="Oops! We couldn't update your information!"
+          content="Oops! We couldn't update your information!"
           type="danger"
           duration="sticky"
           onExit={() => { }}
         />
         <Alert
-          content=<SampleContentComponent />
+          content= <SampleContentComponent />
```
snags88: i think this needs curly braces

PR #110, opened 2019-06-06
added legal pages as reusable components

ZeMunchkin:
Trello: https://trello.com/c/VgBHk5v8

We use the legal pages in both Gatsby (as the static legal pages) and in the main app (as part of the legal modal component). This was causing issues where there was not a single source of truth for the content. Case in point, when adding these to radiance and comparing the dates updated on the two repos, there was an instance where the information on the main app was updated and the gatsby one was not.

From here, we can use the following components in either repo:
```
<TermsOfService />
<PrivacyPolicy />
<Disclaimer />
<TelehealthConsent />
<ParentGuardianTelehealthConsent />
<LegalContainer />
```

The last component would be used for various sweepstakes / promotions terms pages so they can easily have the same styling as our other legal pages.

### DIFF
--- a/src/shared-components/index.js
+++ b/src/shared-components/index.js
@@ -2,6 +2,7 @@ export { default as Accordion } from './accordion';
 export { default as Alert } from './alert';
 export { default as Avatar } from './avatar';
 export { default as Banner } from './banner';
+export { default as BulkErrors } from './bulkErrors';
 export {
   default as Button,
   RoundButton,
@@ -14,8 +15,8 @@ export { default as Chip } from './chip';
 export { default as Container } from './container';
 export { default as Dropdown } from './dropdown';
 export { default as Field } from './field';
-export { default as BulkErrors } from './bulkErrors';
 export { default as ImmersiveModal } from './immersiveModal';
+export * from './legalComponents';
 export { default as LoadingSpinner } from './loadingSpinner';
 export { default as Modal } from './modal';
 export { default as OffClickWrapper } from './offClickWrapper';
--- a/src/shared-components/legalComponents/disclaimer/index.js
+++ b/src/shared-components/legalComponents/disclaimer/index.js
@@ -1,0 +1,74 @@
+import React from 'react';
+
+import { LegalContainer } from '../style';
+
+const Disclaimer = () => (
+  <LegalContainer>
+    <h1>Disclaimer</h1>
+    <time dateTime="2019-01-09">Last Updated: January 9, 2019</time>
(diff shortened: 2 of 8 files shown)


snags88:
@ZeMunchkin What do you think of creating a separate repo for this and the guides config that should be shared between PocketDerm and Gatsby? It doesn't feel like this belongs in Radiance

ZeMunchkin:
@snags88 Yeah, Matt and I went back a forth a bit on this while pairing yesterday. I agree this doesn't 100% fit into radiance, and I could go either way. The reason we initially landed on putting into radiance is because both forms of content do have a reusable component aspect to them. The legal page's `LegalContainer` (which is sort of a stretch as far as arguments go), and the guide content has the `GuideCard`. `GuideCard` actually takes a couple different forms currently, so it would resemble our reusable components in radiance quite a bit more than `LegalContainer`.

But it definitely makes sense to just break out and make another repo for this. Any fun name ideas? :)

snags88:
@ZeMunchkin makes sense. My main concern, especially for the legal components, is that this is open source so anyone can make PRs against it. I think merging is locked down to Curology Engineers, but just a precaution if we ever allow others outside of the organization to do things in this repo

ZeMunchkin:
@snags88 That's an interesting point that we didn't talk about... I guess I'm thinking about how it would look as a separate repo if that repo was not open-source. A published package was a nice way of having the content available at build time and ready to use in React components. It looks like NPM has a private registry, so that might be an option?: https://www.npmjs.com/products/enterprise

The other thing to consider is that Gatsby is SSR, so we have to ping an API at build time. And If we had a separate repo for content that we were making API calls to instead, we'd need to figure out the mechanics of doing that. It will introduce graphql into our codebase: https://www.gatsbyjs.org/docs/client-data-fetching/

The content we'd get from an API would have to be in JSON format (I believe?), so we'd still need to add the reusable components (`LegalContainer` and `GuideCard`) to radiance so that we can plug in the data we get from the API into those. We'd have to think a bit about how to format the legal content in a way that we could easily have it render in the components, since there's a lot of different html elements that are nested. And there's an argument that the content, how to parse it, and the component it that renders it are very tightly coupled and should be kept together.

snags88:
Hmm interesting. It looks like we can ping the pocketderm api from gatsby so we can get the info from our servers if we need (https://www.gatsbyjs.org/docs/using-gatsby-without-graphql/)

As for the actual components, it makes sense to keep them in a centralized repository if the style is always going to be the same

PR #113, opened 2019-06-06
hotfix/2.0.41

mjaltamirano:
Fixes some security issues. The remaining issue with `marked`, which is moderate severity, is linked to upgrading from Storybook 4 to 5, which I will explore separately.*

(*I've spoken with Design and it sounds like Storybook is not the most friendly tool from them, for documentation/onboarding purposes. It looks like Storybook 5 might be better suited for their purposes.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## v2.0.41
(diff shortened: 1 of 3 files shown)


mjaltamirano:
Closing in favor of #116

PR #114, opened 2019-06-06
v2.1.0 - Add CircleCI

mjaltamirano:
Standing on the shoulders of giants.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,10 +42,9 @@ jobs:
             - <<: *prospective-merge
             - <<: *yarn-cache-restore
             - run:
-                name: Install and Build
+                name: Install
                 command: |
                     yarn install
-                    yarn build
             - <<: *yarn-cache-save
             - persist_to_workspace:
                 root: *root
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,85 @@
...
+
+        docker:
+            - image: circleci/node:10-stretch
+
+        steps:
+            - checkout
+            - <<: *prospective-merge
+            - <<: *yarn-cache-restore
+            - run:
+                name: Install and Build
```
maxcountryman: Based on the [Codeship project](https://app.codeship.com/projects/312533/tests/edit) it seems like we only need to invoke `yarn install`. I would probably change this to `Install` and remove the `yarn build` invocation.

mjaltamirano:
It looks like we're trying to use semver in this repo but all of the changelog history have been patches. A workflow like [`git-flow`](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) might be worth adopting for `radiance-ui` and `gatsby` since we bump the versions in `radiance-ui`, and `gatsby` makes use of a `staging` branch that we merge into `master`. 

In the meantime I'm releasing this as `2.1.0` since it isn't a patch but is neither a breaking change for the consumable API.

PR #117, opened 2019-06-07
added accordion prop to right align the arrow

ZeMunchkin:
In this PR:
- Added an additional prop to Accordion for when the designs need the arrow to be flush on the right side of the accordion

### DIFF
--- a/docs/accordion.md
+++ b/docs/accordion.md
@@ -39,6 +39,14 @@ import { Accordion } from 'radiance-ui';
   >
     <div>Expansion</div>
   </Accordion>
+  <Accordion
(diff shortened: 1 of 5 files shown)


ZeMunchkin:
@youngmanalive Yeah, it looks funny because in this example the title is padded. But we have a design right now for the LP that needs the right arrow flush and it looks fine there

design: https://www.figma.com/file/vJsU12XrsZFtEZvDi5n1lC/WebsiteRedesign_MV?node-id=1934%3A13206
Scroll to the FAQ section on General Mobile.

youngmanalive:
👍

PR #124, opened 2019-06-27
Update emotion dependencies

mjaltamirano:
### What & Why

1. `emotion` is the framework-agnostic version of the library. [The `Global` component in `@emotion/core` achieves the same functionality as `injectGlobal`](https://emotion.sh/docs/globals), so I've replaced that functionality and removed `emotion` as a dependency.
    - This will also fix the missing dependency issue in our main app. 
2. I've upgraded all emotion dependencies to 10.0.14. They are all updated in tandem so all emotion updates should occur simultaneously. 

### Other

We log the following warning in development (from our tests for `button` and `roundButton`:

```
      Warning: Received `false` for a non-boolean attribute `loading`.
      
      If you want to write it to the DOM, pass a string instead: loading="false" or loading={value.toString()}.
      
      If you used to conditionally omit it with loading={condition && value}, pass loading={condition ? value : undefined} instead.
          in div (created by Context.Consumer)
          in Loader (created by Context.Consumer)
          in button (created by Context.Consumer)
          in div (created by Context.Consumer)
          in RoundButton (created by WrapperComponent)
          in WrapperComponent
```

I may or may not create a follow-up PR shortly to update this, depending on how big the diff would be in our other repos.


snags88:
sweet, yeah we should deprecate `loading` and use `isLoading`. Probably need to support both for a minor update and then remove `loading` in the next major

mjaltamirano:
Currently investigating why the styles that I was seeing locally are: 1. Not on review app, but, more importantly, 2. Not appearing locally either 😅

mjaltamirano:
I removed optionConfig thinking it was an unused import which broke the changes. I decided to move the optionConfig content into `config` in bf0efee because that is how all the Storybook documentation suggests doing it. Seems simpler than having multiple files in this case since we've added so much more global configuration (since I joined! 😅)

PR #125, opened 2019-06-28
[Chore] Update button loading prop usage

mjaltamirano:
### What & Why

1. Corrects warnings logged in #124 related to passing non-boolean `loading` attribute to the DOM by renaming all usage in `<Button />` and `<RoundButton />` to `isLoading`. 
1. Adds a deprecation notice to use of `loading`

### Other
- Once this is merged in, I intend to release Radiance with a minor bump to update our emotion use. I'll then make the necessary changes in PocketDerm before then removing the deprecation warning. (Gatsby does not use this prop, though it looks like it is present in exactly one snapshot, due to it being a defaultProp).

### DIFF
--- a/src/shared-components/button/index.js
+++ b/src/shared-components/button/index.js
@@ -5,6 +5,19 @@ import { css } from '@emotion/core';
 import Loader from './shared-components/loader';
 import Container from './shared-components/container';
 import { ButtonBase, ButtonText, ButtonContents } from './style';
+import withDeprecationWarning from '../../utils/withDeprecationWarning';
+
+const deprecatedProperties = {
+  loading: "The 'loading' prop is deprecated. Use 'isLoading' instead.",
+};
+
+const isLoadingPropFunction = (props, propName, componentName) => {
+  if (props[propName] !== undefined) {
+    return new Error(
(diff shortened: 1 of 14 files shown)


stories/button/button/index.js
```diff
@@ -31,14 +31,14 @@ const ButtonStory = withDocs(ButtonReadme, () => (
       <Button buttonType="quaternary">Quaternary Button</Button>
       <Button disabled>Disabled Button</Button>
 
-      <Button loading>Primary Loading</Button>
-      <Button loading buttonType="secondary">
+      <Button isLoading>Primary Loading</Button>
+      <Button isLoading buttonType="secondary">
```
mjaltamirano: I tested that the basic `loading` functionality worked by changing these locally to be `loading` and seeing no regressions.

src/shared-components/button/index.js
```diff
@@ -39,33 +53,40 @@ class Button extends React.Component {
       children,
       buttonType,
       loading,
+      isLoading,
       icon,
       textColor,
       ...rest
     } = this.props;
 
+    const loadingVal = loading === undefined ? isLoading : loading;
```
mjaltamirano: Core of the backwards compatibility.

PR #129, opened 2019-06-28
Feature/add button full width prop

mjaltamirano:
This PR adds a `fullWidth` prop to Button components to make them expand to take the width of their container. 

There is a noticeable issue where, because the Icon does not display when the component is loading, there is an empty svg present with spatial properties that inadvertently messes up the height of the button. You can see it in the lack of perfect alignment of the full-width buttons in this screenshot:

Screen Shot 2019-06-28 at 4 31 38 PM

Setting the `display: none` on the svg fixes the alignment, but we currently rely on it for positioning the text and loading animation. Similarly, if you turn off `font-size: 0.75rem` on both side-by-side button components, they are again aligned. 

I'm not concerned with it from a cosmetic standpoint in the storybook but it is something that could possibly cause an issue in use.


src/shared-components/button/index.js
```diff
@@ -56,6 +58,7 @@ class Button extends React.Component {
       isLoading,
       icon,
       textColor,
+      fullWidth,
```
mjaltamirano: I opted for `fullWidth` over `isFullWidth` because I anticipated it would be a value we would nearly always be setting explicitly, rather than conditionally.

src/shared-components/button/style.js
```diff
@@ -118,6 +118,7 @@ export const baseButtonStyles = ({
   buttonType,
   isLoading,
```
benkolde: image

Is it possible to move the loading spinner into the same area as the text here?  It currently pushes way out to the side.

benkolde:
@mjaltamirano as far as the height thing goes, it's definitely weird. Would removing the fontSize property on the full width variant change anything?

mjaltamirano:
> @mjaltamirano as far as the height thing goes, it's definitely weird. Would removing the fontSize property on the full width variant change anything?

We went over the current Loader icon spacing fix in person, which looked good to you, but for this I actually stumbled upon what appeared to be a decent fix while troubleshooting how to get the Loader spacing improved. You can see that in the code here: https://github.com/PocketDerm/radiance-ui/pull/129/files#diff-e54f785f0d62d47619cf001bc205d9fbR171

src/shared-components/button/shared-components/loader/style.js
```diff
@@ -35,7 +35,7 @@ const ButtonLoader = styled.div`
   right: 0;
   top: 0;
   margin-top: -6px;
-  width: 38px;
+  width: ${({ isFullWidth }) => (isFullWidth ? `25%` : `38px`)};
```
mjaltamirano: 25% was fairly arbitrary but was relatively close for very wide buttons (which isn't a pattern we want to use, since we're adding this for mobile), but also close enough with narrower viewports (320-400px).

PR #130, opened 2019-07-01
Feature/update lint staged config

mjaltamirano:
This PR expands our `husky` and `lint-staged` use. 

1. It adds a pre-push `yarn run test` hook so that we can find and fix tests before they fail on CI. 
1. It applies `prettier` formatting to `.md` and `.json` files
    - This applies, most significantly, to our manual markdown prop tables. Ideally, we can start pulling them automatically from the component, but we are not doing that yet. 
1. It applies `prettier` formatting to `style/styles/test.js` files.
    - This makes an assumption that we were originally ignoring these files because we didn't want to run `eslint` on them. I still think we should format them. 
1. It applies `prettier` formatting and `eslint` checking on other source files.
    - It looks like there was an incorrect path to these files in that `resources` does not exist in this repo, so we were neither applying `prettier` nor checking for `eslint` warnings/errors before pushing. 

The intent of this (and #129) is to make greater use of the tools we're using offer. I think we should rely more on auto-formatting to ensure a consistent code style (so we don't have to comment on it in PRs, and so it is easier to read), and to prevent unnecessary CI builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn run build:svgs:index && yarn run lint-staged"
+      "pre-commit": "yarn run build:svgs:index && yarn run lint-staged",
+      "pre-push": "yarn run test"
     }
   },
   "lint-staged": {
--- a/package.json
+++ b/package.json
@@ -130,7 +130,7 @@
     "concurrent": false,
     "subTaskConcurrency": 1,
     "linters": {
-      "resources/**/!(style|styles|test).js": [
+      "!(style|styles|test).js": [
         "prettier --write --single-quote",
         "eslint --fix",
         "git add"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -122,17 +122,26 @@
...
-      "pre-commit": "yarn run build:svgs:index && yarn run lint-staged"
+      "pre-commit": "yarn run build:svgs:index && yarn run lint-staged",
+      "pre-push": "yarn run test"
     }
   },
   "lint-staged": {
     "concurrent": false,
     "subTaskConcurrency": 1,
     "linters": {
-      "resources/**/!(style|styles|test).js": [
```
snags88: lol i must have copied this from the pocketderm repo. good find

snags88:
This is great, we should automate these types of checks where we can!

PR #131, opened 2019-07-01
Feature/expand jest config

mjaltamirano:
This PR expands our `jest.config.js` file. 

1. It removes the `jest-config` import because we're only relying on it to ignore `node_modules`. It's clearer to include that explicitly. 
1. It adds coverage thresholds to our components (that currently pass) so we can maintain sufficient code coverages on the components, and enforce it. 

The intent of this (and #130) is to make greater use of the tools we're using offer. In particular the coverage thresholds will help us stay sharp when we make changes and/or add components to the library, to ensure that we're including sufficient testing for those changes.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,10 +1,14 @@
 module.exports = {
   collectCoverage: true,
-  collectCoverageFrom: ['src/shared-components/**', '/src/utils/**', '!src/**/__snapshots__/**.js.snap'],
-  snapshotSerializers: ['jest-emotion'],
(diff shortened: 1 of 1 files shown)


mjaltamirano:
> Im not very good at this config/devops things, but if it works it works.
> Im curious what happen if we dont meet the coverage threshold???

What should happen, with the `pre-push` in #130, is that the commits will not push until the threshold is met again. Here's what it looks like in the terminal:

Screen Shot 2019-07-01 at 10 53 17 AM

PR #133, opened 2019-07-10
Fix option-button bug

benkolde:
**What**

@mvandy found an issue in the consult questionnaire that caused the option buttons to squish the checkmarks at smaller screen sizes / longer text blocks. 

image

This PR adds `flex: 0 0 auto;` to the component which will allow the checkbox / radio to retain its size. 

I haven't checked if this impacts production in any way. I can't imagine it would, but just thought I'd point that out.

### DIFF
--- a/src/shared-components/optionButton/style.js
+++ b/src/shared-components/optionButton/style.js
@@ -48,6 +48,7 @@ const getBaseIconWrapperStyles = ({ selected, optionType, type }) => css`
   height: 32px;
   display: flex;
   flex-flow: row nowrap;
+  flex: 0 0 auto;
   justify-content: center;
   align-items: center;
   transition: all ${ANIMATION.defaultTiming};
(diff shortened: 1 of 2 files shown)


src/shared-components/optionButton/style.js
```diff
@@ -48,6 +48,7 @@ const getBaseIconWrapperStyles = ({ selected, optionType, type }) => css`
   height: 32px;
   display: flex;
   flex-flow: row nowrap;
+  flex: 0 0 auto;
```
mjaltamirano: It looks like the reason this is working is because it's shorthand (partially) for setting `flex-shrink` to 0. 

I think we can just switch to that--`flex-shrink: 0`--to achieve the same functionality.

Given that we want it to be a uniform 32x32 px, as long as we prevent it from shrinking due to flex, this should give us the functionality we want.

daigof: I believe we should keep it as `flex: 0 0 auto` for some reason it is always recommended to use the shorthand version property and the other values set here are the default so it is set correctly.
look here: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

mjaltamirano: My reading of that excerpt is that it's preferred to use the shorthand if you're setting the other properties, but for our purposes it seems like we only want the box not to shrink--the other properties are less important. 

(It seems an open question to me why this particular styling is achieved with so much flex, too.)

In any case I'm cool to set grow explicitly to 0 and flex-basis to auto. The former is good, the latter might have some wonky behavior, but we should be fine.

PR #137, opened 2019-07-25
Add rest params to immersive modal

daigof:
- Needed to send custom styles to immersive modal

### DIFF
--- a/src/shared-components/immersiveModal/index.js
+++ b/src/shared-components/immersiveModal/index.js
@@ -74,9 +74,11 @@ class ImmersiveModal extends React.Component {
   };
 
   render() {
-    const { children, canBeClosed, onClose, header } = this.props;
+    const {
+      children, canBeClosed, onClose, header, ...rest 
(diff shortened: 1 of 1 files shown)


mjaltamirano:
@daigof do we know in advance what these props are going to be? Or rather, what the specific use case is? Are they css props? Are they style props?

I’m not a big fan of spreading props because it can obscure what props are being passed through there is strict typing

daigof:
I understand your point I added @snags88 to chime in. 

Ok so:
- we do have currently the rest params in other components
- I believe Seiji explained to me once these were to pass `data-` attributes to perform test and also to set ids like for VWO
- I need this rest for setting the custom style e.g. `css={css` margin:10px `}`
- I was gonna to propose to set the rest param for all the components

snags88:
For component libraries, i'm not opposed to having the spread/rest params since there are those types of attributes we want to pass along to render to the dom. For most other components, I agree with you @mjaltamirano

PR #141, opened 2019-09-05
Fix select dropdown on mobile with disabled default value as selected

daigof:
- Part of https://app.asana.com/0/1130276194470636/1135497036319695/f
- In this case we are passing a default selected value that is also disabled
- In desktop works OK because is  custom  
- in mobile we have a `select` tag so when we try to select a disabled option it doesnt work and skip to the next undisabled value causing the bug
- This fix is smart enough that once you select a value, the previous default become disabled so you cannot go back and select that again

### DIFF
--- a/src/shared-components/dropdown/mobileDropdown.js
+++ b/src/shared-components/dropdown/mobileDropdown.js
@@ -13,15 +13,25 @@ const MobileDropdown = ({
       value={value || ''}
       onChange={onSelectChange}
     >
-      {options.map(option => (
-        <option
-          key={option.value}
-          value={option.value}
-          disabled={option.disabled}
-        >
-          {option.label}
-        </option>
-      ))}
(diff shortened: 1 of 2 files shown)


src/shared-components/dropdown/mobileDropdown.js
```diff
@@ -13,15 +13,25 @@ const MobileDropdown = ({
...
-        <option
-          key={option.value}
-          value={option.value}
-          disabled={option.disabled}
-        >
-          {option.label}
-        </option>
-      ))}
+      {options.map(option => {
+        let disabledValue = option.disabled;
```
mjaltamirano: Would this be clearer as `isDisabled` or `shouldDisable` (or `shouldBeDisabled`)? 

Also, should we change `option.disabled` to `!!option.disabled` to coerce undefineds to false?

PR #142, opened 2019-09-26
Update accordion to determine height after mount

youngmanalive:
This fixes a small issue where the Accordion wouldn't render open if the `isOpen` prop is initially `true`. 

We're using a `ref` to determine content height inside the accordion, but we were attempting to set the height before we had access to the node, thus it was still mounting as '0px'.

This PR waits for node availability and then updates the height in `componentDidMount`.

@heyamykate --  RE: [this PR](https://github.com/PocketDerm/PocketDerm/pull/6633) -- I think this is why the photo upload accordion originally had the `currentPerspective` set to null, then it would update on mount so the accordion would actually open. Right now, it loads as open but the content height is still set to '0px', so users need to click it twice to get it to expand.

### DIFF
--- a/src/shared-components/accordion/index.js
+++ b/src/shared-components/accordion/index.js
@@ -50,11 +50,10 @@ class Accordion extends React.Component {
 
   contentRef = React.createRef();
 
-  constructor(props) {
-    super(props);
-    this.state = {
-      contentHeight: this.getContentHeight(props.isOpen),
-    };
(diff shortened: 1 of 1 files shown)


mjaltamirano:
How did you go about testing this locally? I am still seeing the double-click issue after pulling in this source to a file and importing it in place of `import { Accordion } from 'radiance-ui';`

PR #148, opened 2019-10-30
Typescript Integration - Phase 1

daigof:
###  Scope: 
1. Write RadianceUI code in TS and be able to convert all JS code into TS with a **working build system** : Includes: Stories, Components, Utils Functions and Constants.
2. Not included: SVG icons generation which will remain being generated in JS. (we may incorporate a custom declaration file for these later)
3. Convert `utils` and `constants` completely. Have a working build for converting Stories and Components: Converted Accordion component and story as example

Extras (mostly copied from  PocketDerm repo)
-  `lint-staged` configuration so we dont need to put all TS configs flags in package.json anymore and would pick up the regular `tsconfig.json` file
- Bump eslint rules (needed because new typescript eslint incorporation)

---- 

### Background info:
There are **3 main Build Systems** that needs to play nice together and consume the same tsconfig.json:
- **Rollup**: added the typescript plugin it worked fine. Note that we in an outdated version (0.66 compared to latest 2.3.4) both main rollup and compatible plugins.
- **Babel**: There are 2 babel configs: the one in the root is for SVG generation so we don't need  TS plugin there. The one in `src` is to generate the lib folder for utils and constants, it worked fine with the babel plugin also had to add the TS extensions to the cli command.
- **Storybook**:  Stories can be written in TS now and it has its own tsconfig.json. This is because we only want one config in the root project that generates the types delcarations. if we includes stories in this config it would mess up with the folder creation inside `lib` and also we don't need declarations for stories. 

- **Conclusion**: we can run all build commands (`storybook` and `build`) and this has 0 impact in the library generation.

### Type generation (how it would work in PocketDerm or other Projects consuming Radiance)
- `tsc` command added to the build command. With the new upgraded TS version we can have `allowJs=true` and generate declaration files automatically. We run this command in the build at the end which will place all `.d.ts` inside the `lib` folder. 
- Added main library Types definition path in package.json
- For now and until all components are converted to TS it is recommended that we continue using `declare module 'radiance-ui';` 
- Until we create a custom decalration file or solution for icons we should continue using `declare module 'radiance-ui/lib/icons';` as well
- We should have full type support for constants and utils functions.


mjaltamirano:
> ### What about type generation and the promise to solve all `@ts-ignore` in PocketDerm
> * I placed a `tsc-build-types` cli command with a custom `tsconfig`. Problem is you have to have your entire project written in TS for it to generate correctly because the `declaration` flag (this generated `d.ts` declaration files) CANNOT be set with `allowJs=true`

I'm only just starting to look at this, and have a meeting very shortly, but for what it's worth, it looks like this is [slated for release in TypeScript 3.7](https://github.com/Microsoft/TypeScript/issues/7546#issuecomment-535699638) on [November 5th](https://github.com/microsoft/TypeScript/issues/33352). Did you see this in your development and play around with the release candidate? (`yarn add typescript@rc`)

daigof:
@mjaltamirano The build types script I created works,only for the constants (I excluded all other sections in the config) it just doesnt generate anything useful.  Even if we convert all files to TS we still would need to exclude some JS files related to SVG Components generation.

mjaltamirano:
> @mjaltamirano The build types script I created works, only for the constants (I excluded all other sections in the config) it just doesn't generate anything useful. Even if we convert all files to TS we still would need to exclude some JS files related to SVG Components generation.

Right, I think something I am considering proposing (after playing around with this PR and the TypeScript 3.7 release candidate for a couple hours) is:

1. Waiting for TypeScript 3.7 to be released, so we can explore generating type declarations for files currently written in JavaScript.
    - This significantly reduces the burden of needing to convert files over to .ts before being able to obtain useful types. Working on simple constants and utils is one thing, but this PR doesn't address the main components in this library. 
        - I don't think the foundational PR **needs** to do that, but it would be helpful if we could just flip a switch in order to achieve that at the same time, and we don't **urgently** need the types for `constants` and `utils`. 
    - It looks like our existing `propTypes` definitions may be sufficient for early typing, before spending time on more desired specificity. 
1. Publishing some of this TS work under an `alpha/beta` tag for testing in consuming repos.
    - We haven't, to my knowledge, made much use of this publishing functionality, but for a significant-enough step like this I think it would be prudent.
1. Based on that testing, deciding what the most appropriate next steps are. 

The reasons for this approach are based on the following:
1. A potentially very, very helpful feature being released just next week! 
    - If it can save us work, and there's no urgent need to make these changes, then we ought to wait. 
    - This PR **still** will provide great foundation, but with the knowledge we have now, it seems prudent to delay a little while longer. 
1. Insufficient resourcing for reviewing this PR to a mergeable state + releasing before then. 
    - The next two days of work are booked, and we do not know what resourcing will look like beyond that. 
1. Given the above, there is considerable risk of taking steps that we would end up undoing shortly thereafter.

This is, admittedly, some short-sightedness on my part, since I didn't realize that TypeScript 3.7 would have some functionality that would make incorporating types into this codebase much easier. Sorry about that!

I would encourage you to push the changes you've made to your local branch, and then pausing this work. I'm going to leave some comments throughout for discussion about what steps we want to take and how the end-product would look.

.gitignore
```diff
@@ -81,3 +81,6 @@ lib/
 src/icons/
 .out
 .size-snapshot.json
+
+# Typescript declaration types generated 
+.tempTypes
```
mjaltamirano: Based on your research and work, how do you think we should be generating these types? It seems like it is standard to commit the generated `.d.ts` files alongside the associated source file. Was `.tempTypes` to help you create/delete with ease?

.storybook/webpack.config.js
```diff
@@ -73,5 +73,19 @@ module.exports = webpackSettings => {
     })
   );
 
+  config.module.rules.push({
+    test: /\.(ts|tsx)$/,
+    use: [
+      {
+        loader: require.resolve('awesome-typescript-loader'),
```
mjaltamirano: When I was playing around with this PR, it seemed like it was possible to generate types just using `babel-loader` along `@babel/preset-typescript`.

I think we should try and stay within the Babel ecosystem as much as possible. Did you play around with it at all?

package.json
```diff
@@ -26,10 +26,13 @@
...
     "build:icons:index": "create-index src/icons/",
     "build:rollup": "rollup -c",
-    "build:babel": "babel src -d lib",
+    "build:babel": "babel src -d lib --extensions '.js,.ts,.tsx'",
     "build:svgs:index": "create-index src/svgs/ --recursive --update --extensions svg",
     "publish-package": "run-s build && yarn publish",
-    "heroku-postbuild": "yarn run build:icons && yarn run build:icons:index && yarn run build-storybook"
+    "heroku-postbuild": "yarn run build:icons && yarn run build:icons:index && yarn run build-storybook",
+    "tsc": "tsc",
+    "tsc-with-tsconfig": "tsc --allowJs true --baseUrl src --declaration false --emitDeclarationOnly false --esModuleInterop true --experimentalDecorators true --forceConsistentCasingInFileNames true --jsx preserve --module commonjs --moduleResolution node --noEmit true --noImplicitReturns true --noUnusedLocals true --noUnusedParameters true --skipLibCheck true --strict true --suppressImplicitAnyIndexErrors true --target es2018",
```
mjaltamirano: We are probably deprecating this in the future, in favor of just `tsc`, FYI. The consideration is about how much of the codebase is converted. This command works well for spotting issues in the specific files being committed, but it's possible that changes in one file creates issues in other files that aren't being edited, so this command cannot catch those.

mjaltamirano:
> * Turns out coding  declarations for a library is a language in itself with very confusing documentation and hard to test, I got frustrated and decided to do that in a separate PR

Does this work exist locally, or is it an idea/plan?

.storybook/webpack.config.js
mjaltamirano: Did you look into using `babel-loader` here, too? https://storybook.js.org/docs/configurations/typescript-config/#setting-up-typescript-with-babel-loader

package.json
```diff
@@ -105,8 +113,12 @@
     "rollup-plugin-commonjs": "^9.2.0",
     "rollup-plugin-node-resolve": "^3.4.0",
     "rollup-plugin-size-snapshot": "^0.7.0",
+    "rollup-plugin-typescript": "^1.0.1",
     "storybook-readme": "^5.0.5",
     "tinycolor2": "^1.4.1",
+    "ts-jest": "^24.1.0",
```
mjaltamirano: Did you run into limitations of `@babel/preset-typescript` that required `ts-jest`? The `ts-jest` repo has the following comparison doc (https://kulshekhar.github.io/ts-jest/user/babel7-or-ts), but I'm not sure it applies much to us.

package.json
```diff
@@ -105,8 +113,12 @@
...
+    "tslib": "^1.10.0",
```
mjaltamirano: It doesn't look like we're using this: https://github.com/Microsoft/tslib#usage

src/constants/boxShadows/index.ts
```diff
@@ -2,6 +2,7 @@
  * Note: Each Category is exported separately to be rendered in storybook:
  */
 
+// @ts-ignore
```
mjaltamirano: I think your branch with the unpushed changes should fix these kinds of issues.

.gitignore
daigof: this folder is the outDir for the types in the config. As I said the generator was not providing anything useful and it just cluttered all the codebase I had to manually delete all generated `d.ts` files from the `src` folder. I have looked at other project declarations files and all of them looks like they use custom declaration written by hand.

Besides the generator would put the `d.ts` files inside `src` folder and we dont want that we want that we are not publishing that

package.json
daigof: but you said the `tsc` command alone wont pick up the tsconfig.json in the hook phase? Yeah I understood this is to make the lint-staged hook work. this is literally copied this from pocketderm anyway

src/constants/boxShadows/index.ts
daigof: Oh this was temporary my focus was to make the build work. 
Pushing `utils` convertion now. Anyway I was testing just to make `1` ONE freaking constant to work `ANIMATION.defaultTiming` that didnt even use any utils. to not avail. I will put here the solution I reach but I am NOT happy with it

package.json
daigof: see below answer

package.json
daigof: Both are dependecies I was told to install in the TS setup for rollup and storybook
`ts-jest`: https://storybook.js.org/docs/configurations/typescript-config/ 
`tslib`:  https://github.com/rollup/rollup-plugin-typescript

daigof:
@mjaltamirano 
### Working declaration example
go to Curology repo and open a simple style.ts file I used `resources/assets/jsx/admin/views/phone-consultation-time-slots/components/calendarView/style.ts` . Replace `COLORS` for `ANIMATION`
remove `ts-ignore`

There are **2 goals**

- Make the import error go away (sort of easy)
- MOST important get a good declaration that makes the intellinsense pick up the `defaultTiming` property from `ANIMATION`. Step number 1 is NOT enough

Solution I found:
- create a `index.d.ts` file in `node_modules/radiance-ui`
- paste
```
declare module 'radiance-ui/lib/constants' {
  export const ANIMATION: { defaultTiming: string };
}
```

### Why I am not happy with this solution
- First you realize that the types generated by the tool are mostly useless you can copy paste some thing but you have to end up writing the yourself (at least for now)
- The idea is to `import AnymationType` from a different file, because imagine putting all constants in one file (take a look at COLORS for example). I could simply not make this work it doesn't work with `import`
- instead of import we may need to do some weird triple slash directive as I saw in some examples, couldnt make it work either (https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html)
- the documentation is obscure https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html  I may need to read it a couple of times to nail it

package.json
daigof: solved updating lint-staged config

.gitignore
daigof: not need this temp folder anymore. deleting...

.storybook/webpack.config.js
daigof: Replaced to babel-loader, removed the previous one.

.storybook/webpack.config.js
```diff
@@ -73,5 +73,18 @@ module.exports = webpackSettings => {
...
+    use: [
+      {
+        loader: require.resolve('babel-loader'),
+        options: {
+          presets: [['react-app', { flow: false, typescript: true }]],
+        },
+      },
+    ],
+  });
+  config.resolve.extensions.push('.ts', '.tsx');
```
michaeljaltamirano: Let's just add these extensions to L32 (https://github.com/PocketDerm/radiance-ui/pull/148/files#diff-6da49b24efd584154cfb4ec759745192R32) and remove this line.

babel.config.js
```diff
@@ -1,13 +1,9 @@
 module.exports = {
-  presets: [
-    '@babel/env',
-    '@babel/react',
-    '@emotion/babel-preset-css-prop',
-  ],
+  presets: ['@babel/env', '@babel/react', '@emotion/babel-preset-css-prop'],
```
michaeljaltamirano:
```suggestion
  presets: ['@babel/env', '@babel/react', '@emotion/babel-preset-css-prop', '@babel/preset-typescript'],
```

.storybook/webpack.config.js
```diff
@@ -73,5 +73,18 @@ module.exports = webpackSettings => {
     })
   );
 
+  config.module.rules.push({
+    test: /\.(ts|tsx)$/,
+    use: [
+      {
+        loader: require.resolve('babel-loader'),
+        options: {
```
michaeljaltamirano: We should be able to get this compiling without needing to add an additional babel preset. 

I added a comment that has a commit suggestion that will allow us to remove this `options` object and remove `babel-preset-react-app` from `package.json`.

package.json
```diff
@@ -103,9 +113,13 @@
     "rollup-plugin-babel": "^4.0.3",
     "rollup-plugin-commonjs": "^9.2.0",
     "rollup-plugin-node-resolve": "^3.4.0",
-    "rollup-plugin-size-snapshot": "0.7.0",
+    "rollup-plugin-size-snapshot": "^0.7.0",
+    "rollup-plugin-typescript": "^1.0.1",
     "storybook-readme": "^5.0.5",
     "tinycolor2": "^1.4.1",
+    "ts-jest": "^24.1.0",
```
michaeljaltamirano: We can remove this and just rely on `@babel/preset-typescript` (we're not configuring `jest` to use `ts-jest` in this branch).

src/utils/throwOnUndefinedProperty/test.ts
```diff
@@ -5,5 +5,6 @@ test('throws an error when trying to access undefined property', () => {
   const protectedObj = throwOnUndefinedProperty(obj);
 
   expect(protectedObj.foo).toBe('bar');
+  // @ts-ignore
```
michaeljaltamirano: Let's add to this comment explainer that TypeScript is catching the very error we're seeking to avoid throwing.

package.json
```diff
@@ -92,9 +101,10 @@
     "jest": "^24.9.0",
     "jest-emotion": "10.0.10",
     "jest-junit": "^10.0.0",
-    "lint-staged": "^7.3.0",
+    "lint-staged": "^9.4.3",
     "lodash.round": "^4.0.4",
     "prettier": "1.14.3",
+    "react-docgen-typescript-loader": "^3.3.0",
```
michaeljaltamirano: This has since been removed so let's uninstall it.

package.json
```diff
@@ -55,23 +57,30 @@
...
+    "@types/enzyme": "^3.10.3",
+    "@types/jest": "^24.0.20",
+    "@types/react-transition-group": "^4.2.3",
+    "@types/tinycolor2": "^1.4.2",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
     "babel-jest": "^24.9.0",
     "babel-loader": "^8.0.4",
     "babel-plugin-transform-svg-import-to-string": "^0.0.1",
+    "babel-preset-react-app": "^9.1.0",
```
michaeljaltamirano: Based on the previous comments, we should remove this too.

package.json
```diff
@@ -55,23 +57,30 @@
     "@babel/plugin-transform-parameters": "^7.1.0",
     "@babel/preset-env": "^7.1.0",
     "@babel/preset-react": "^7.0.0",
+    "@babel/preset-typescript": "^7.6.0",
     "@babel/register": "^7.0.0",
     "@emotion/babel-preset-css-prop": "10.0.9",
     "@storybook/addon-a11y": "^5.1.9",
     "@storybook/addon-actions": "^5.1.9",
     "@storybook/addon-centered": "^5.1.9",
+    "@storybook/addon-info": "^5.2.5",
```
michaeljaltamirano: Was this added because it was a missing `peerDependency` of something?

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "outDir": ".tempTypes",
+    "rootDirs": ["src", "stories"],
+    "skipLibCheck": true,
+    "strict": true,
+    "suppressImplicitAnyIndexErrors": true,
+    "target": "es2018"
+  },
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
```
michaeljaltamirano:
```suggestion
  "include": [
    "stories/**/*.ts",
    "stories/**/*.tsx",
    "src/**/*.ts",
    "src/**/*.tsx"
  ]
```

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+  "exclude": ["node_modules"]
```
michaeljaltamirano: I'm not sure this is doing anything. Did you see differently if you remove it?

michaeljaltamirano:
I think the two most important next steps are:

1. Convert over a couple functional directories as POC for all the various configurations.
    - We've converted utils, but most of those are `.ts` files--let's have some `.tsx` examples in this first phase, too. 
    - We'll want to make sure TS in `/src` works, as well as `/stories`, for both local development and the build step. 
        - I was working with `stories/constants` while making these comments.
1. Try and use new TypeScript features regarding emitting declaration files from js so we can just rely on `tsconfig.json` and remove `tsconfigGenerateTypes.json`.

babel.config.js
daigof: we dont need typescript here, adding it makes the build fails, the plugin is set in the `src` folder which we run the command babel cli manually for the `lib` folder that contains the constants and utils. I think the issue adding in here is that it tries to convert 2 times since rollup also has the TS plugin

.storybook/webpack.config.js
daigof: done

.storybook/webpack.config.js
daigof: removed options object

package.json
daigof: done

src/utils/throwOnUndefinedProperty/test.ts
daigof: done

package.json
daigof: done

package.json
daigof: done

package.json
daigof: done, I dont remeber adding this I vaguely remeber using withInfo but it doesn seem to exist anymore

tsconfig.json
daigof: done

tsconfig.json
daigof: removed, worked

daigof:
> I think the two most important next steps are:
> 
> 1. Convert over a couple functional directories as POC for all the various configurations.
>    
>    * We've converted utils, but most of those are `.ts` files--let's have some `.tsx` examples in this first phase, too.
>    * We'll want to make sure TS in `/src` works, as well as `/stories`, for both local development and the build step.
>      
>      * I was working with `stories/constants` while making these comments.
> 2. Try and use new TypeScript features regarding emitting declaration files from js so we can just rely on `tsconfig.json` and remove `tsconfigGenerateTypes.json`.

Solved: I ended up adding a lot of configurations just to make stories and components work, I didnt realize it needed a lot of tweaks but now it all works fine, I coverted a story and . a component for Accordion as example so we can pick up from there to fully convert all.

src/shared-components/accordion/test.tsx
```diff
@@ -7,26 +7,22 @@ import Accordion from './index';
 const testAccordion = {
   title: <div>title</div>,
   isOpen: false,
-  onClick: () => {},
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
```
michaeljaltamirano: We can remove this by defining `onClick` like so: `onClick: (): void => undefined,`

src/shared-components/accordion/index.tsx
```diff
@@ -14,9 +14,20 @@ import {
...
 
-const noop = () => {};
-
-class Accordion extends React.Component {
+type AccordionProps = {
+  children: React.ReactNode;
+  disabled?: boolean;
+  isOpen: boolean;
+  noBorder?: boolean;
+  onClick: Function;
```
michaeljaltamirano:
```suggestion
  onClick: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
```

src/shared-components/accordion/index.tsx
```diff
@@ -93,10 +104,13 @@ class Accordion extends React.Component {
     } = this.props;
 
     return (
-      <AccordionBox isOpen={isOpen} noBorder={noBorder} disabled={disabled}>
-        <TitleWrapper onClick={disabled ? noop : onClick} disabled={disabled}>
+      <AccordionBox isOpen={isOpen} noBorder={!!noBorder} disabled={!!disabled}>
+        <TitleWrapper
+          onClick={(event): void => !disabled && onClick(event)}
```
michaeljaltamirano:
```
onClick={(event): void => {
            if (!disabled) {
              onClick(event);
            }
          }}
```

If `disabled` is true then this will return false, which is incompatible with void. This was surfacing before because we were typing `onClick` as `Function`.

src/shared-components/accordion/style.ts
```diff
@@ -22,29 +22,35 @@ export const Content = styled.div`
...
   border: ${border};
 
   ${ExpansionWrapper} {
     ${isOpen && `border-top: ${border};`};
   }
 `;
 
-export const AccordionBox = styled.div`
-  ${({ noBorder, isOpen }) => !noBorder && getBorderStyle(isOpen)};
+export const AccordionBox = styled.div<{
```
michaeljaltamirano: [Nit] I think as much as possible we should try to be consistent with the values we provide in the tagged templates. I think it is undesirable to have a boolean `false` returned to what is, effectively, supposed to be css markup. I've revised this function like so:

```
export const AccordionBox = styled.div<{
  noBorder: boolean;
  isOpen: boolean;
  disabled: boolean;
}>`
  ${({ noBorder, isOpen }): SerializedStyles | string =>
    !noBorder ? getBorderStyle(isOpen) : ``};

  width: 100%;

  &:not(:last-child) {
    border-bottom: none;
  }

  ${({ disabled }): string =>
    disabled
      ? `
    opacity: 0.4;
    background-color: ${COLORS.disabled};
    border-color: ${COLORS.purple30};
  `
      : ``};
`;
```

src/utils/colors/index.ts
```diff
@@ -0,0 +1,26 @@
+import tinycolor from 'tinycolor2';
+
+function percentToInt(str: string) {
```
michaeljaltamirano:
```suggestion
function percentToInt(str: string): number {
```

src/utils/colors/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  tinycolor(color)
+    .lighten(percentToInt(modifier))
+    .toString();
+
+export const darken: ColorManipulationFunctionType = (color, modifier) =>
+  tinycolor(color)
+    .darken(percentToInt(modifier))
+    .toString();
+
+export const transparentize = (color: string, modifier: number) =>
```
michaeljaltamirano:
```suggestion
export const transparentize = (color: string, modifier: number): string =>
```

(To make consistent with typing the return values for the other helper methods.)

stories/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
```
michaeljaltamirano: I think for clarity between these two `tsconfig` files and their differences it might be clearer to add a third file, `sharedTsconfig.json`, `tsconfigShared`, or `tsconfigBase`, and pull them into both `stories/tsconfig.json` and `tsconfig` at the root. That way we can see what properties apply specifically to the root and what apply specifically to the stories. What do you think?

stories/accordion/subcomponentAccordions.tsx
```diff
@@ -1,19 +1,21 @@
 import React from 'react';
 import { css } from '@emotion/core';
-import { Accordion } from 'src/shared-components';
 
-class DefaultAccordion extends React.Component {
-  state = {
+import { Accordion } from '../../src/shared-components';
```
michaeljaltamirano: You can get these imports by adding the following to the `stories/tsconfig.json`:

```
"baseUrl": "../",
"rootDirs": ["src"],
...

File: src/types/images.d.ts [hunk 25]
```diff
@@ -0,0 +1,5 @@
+// These are to use import images without TS module not found error
```
michaeljaltamirano: It looks like this is outputting an empty  `images.d.js` file. Did you see the same?

michaeljaltamirano:
Have you tried incorporating this into the main repo? 

I get the following error when I copy-paste the emitted types (not the best process, I know) into the `node_modules` install:

```
Could not find a declaration file for module 'radiance-ui'. '/Users/[my_name]/[name_of_repo]/node_modules/radiance-ui/dist/bundle.js' implicitly has an 'any' type.
  Try `npm install @types/radiance-ui` if it exists or add a new declaration (.d.ts) file containing `declare module 'radiance-ui';`
```

Admittedly, I haven't spent too much time investigating how we bundle things, so I'm not sure what is required to get things working in the main repo. 

I think one way to approach testing this out would be to draft a release of a alpha or beta candidate:

image

stories/accordion/index.tsx
```diff
@@ -1,13 +1,16 @@
...
+// @ts-ignore TODO-TS: remove this when updating storybook version since newest includes types
 import { withKnobs, text, boolean } from '@storybook/addon-knobs';
+import { withDocs } from 'storybook-readme';
 import { action } from '@storybook/addon-actions';
 import { css } from '@emotion/core';
-import AccordionReadme from 'docs/accordion';
-import { Accordion, Typography } from 'src/shared-components';
-import { SPACER } from 'src/constants';
 
+// @ts-ignore
```
michaeljaltamirano: I don't think we should merge this in before resolving the issue. We'll want to update how we're handling resolving paths. It is probably just because it is a markdown file and we didn't add that as an extension, in addition to `.ts` and `.tsx`, somewhere.

stories/accordion/index.tsx
```diff
@@ -62,28 +65,30 @@ const accordionWithKnobs = (
   </React.Fragment>
 );
 
-storiesOf('Accordion', module)
```
michaeljaltamirano: If these changes were to try and wrangle 5.2 working, I think we should revert it to keep the diff relevant.

package.json
```diff
@@ -55,8 +57,10 @@
     "@babel/plugin-transform-parameters": "^7.1.0",
     "@babel/preset-env": "^7.1.0",
     "@babel/preset-react": "^7.0.0",
+    "@babel/preset-typescript": "^7.6.0",
     "@babel/register": "^7.0.0",
     "@emotion/babel-preset-css-prop": "10.0.9",
+    "@rollup/plugin-typescript": "^2.0.1",
```
michaeljaltamirano: This has a `peerDependency` of `"rollup": "^1.20.0",`. We're still on v0. I think we should install an older version of this (it looks like `0.8.1` is the last version before `^1`, which requires more rollup changes.)

If it is easier, it might be preferable to spend cooldown working on updating the build dependencies in here in a separate PR, if you find the older versions are blocking in some regard (they apply to much older versions of TypeScript...)

src/shared-components/accordion/test.tsx
daigof: done

src/shared-components/accordion/index.tsx
daigof: thanks I always forgot how to write these.

src/shared-components/accordion/index.tsx
daigof: done

src/shared-components/accordion/style.ts
daigof: done

src/utils/colors/index.ts
daigof: done

src/utils/colors/index.ts
daigof: done

stories/tsconfig.json
daigof: done. Yes it is much simpler now, I also removed some configs that were already in the default values, except for `jsx` for some reason it doesnt work without putting it with default

stories/accordion/subcomponentAccordions.tsx
daigof: I played a little with these values now each project `src` and `stories` have their own configs and was able to refactor all imports to start from root e.g. `src/...` or `docs/...`
As a side note I also had to add a config to Jest to indicate the root as `./` for some reason it doesnt work without it

src/types/images.d.ts
daigof: ok a lot of trial error here. I moved `types` to root so it doesnt output in `lib` had to include it in the tsconfig manually, thankfully that doesnt . make it generate an output from types folder

stories/accordion/index.tsx
daigof: solved. added as a custom module. Note: tried to add it in types but it messes up the `src` imports. Since stories project is not that important we can leave as it is now and add more types there if necessary even if we duplicate from types folder

stories/accordion/index.tsx
daigof: no, this was to make the Knobs work. But it is also bugged in `master` (also alerts components) so I gave up, maybe we can get it fixed when we upgrade the `@storybook` and tis plugins

daigof: reverted back to its initial state for now since it didnt work

package.json
daigof: done, Yeah this was when I was trying to make `shared-component` work with TS I reverted back to its initial plugin compatible with older rollup

daigof:
@michaeljaltamirano 

> Have you tried incorporating this into the main repo? Or was this testing on a new project, in local development?

Yes I tested and  it works, you have to replace `dist` `lib` `package.json` and `node_modules`. Also delete the module definition in types folder and of course the ts-ignore comments
I realised the constant `MEDIA_QUERIES` was not working due to being formed in runtime so I added a type assertion in radiance and now it at least works, too bad we cannot have autocomplete unless we hardcode  the properties?

> Beta/Alpha release

I agree we should release it with Beta tag, this `MEDIA_QUERIES` issue came out of nowhere Im afraid something else would break TS compilation as well.

.prettierignore
```diff
@@ -0,0 +1,6 @@
+src/svgs/deprecated/index.js
```
michaeljaltamirano, line 1: I'm okay with prettier changing the formatting for these files--does it interfere with `// @create-index`?

.lintstagedrc.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  '!(style|styles|test).js': filenames => [
+    `prettier --write ${filenames.join(' ')}`,
+    `eslint --fix ${filenames.join(' ')}`,
+    `git add ${filenames.join(' ')}`,
```
michaeljaltamirano, line 5: If we want to replicate `PocketDerm`'s config, we can bump `lint-staged` to ^10 and remove the `git add` steps (which are done by default in v10).

src/constants/mediaQueries/index.ts
```diff
@@ -9,4 +9,8 @@ const mediaQueries = Object.entries(BREAKPOINTS).reduce(
   {}
```
michaeljaltamirano, line 9: It looks like [Augmenting key name during mapping](https://github.com/microsoft/TypeScript/issues/12754) isn't supported by TypeScript but we can cheat by casting this {} like so:

```
const mediaQueries = Object.entries(BREAKPOINTS).reduce(
  (accumulator, [label, px]) => ({
    ...accumulator,
    [`${label}Up`]: `@media (min-width: ${px}px)`,
  }),
  {} as {
    xsUp: string;
    smUp: string;
    mdUp: string;
    lgUp: string;
    xlUp: string;
    xxlUp: string;
  },
);
```

The return value of `throwOnUndefinedProperty(mediaQueries)` will have only those specific keys as options, instead of any string key, which will allow us to catch errors at compile-time instead of relying on the `throwOnUndefinedProperty` runtime error.

tsconfigBase.json
```diff
@@ -0,0 +1,19 @@
+{
```
michaeljaltamirano, line 1: [Nit] I wonder if we should put this at `shared/tsconfig.json` so that we get the hints that come with a `tsconfig.json` file?

.prettierignore
daigof: - ok, so this is generated and used only by storybook (may need to look at possibility to use icons from libs in other PRs althought build is not required I believe for storybook).  
- So for example take the `trailingComma` format it gets auto-formatted when you do a push
- Then you are working locally in a total unrelated subject and if you run storybook it will change the files back to no commas because it is autogenerated, Then you have 5 files in GIT to add. This is infinite loop so I decided to just not autoformat these files since they are autogenerated

- Now looking at these again I think we can even gitignore them

daigof: OK, so I tried putting them in `.gitignore` but that crahsed the storybook build since it uses `create-index` in update mode and couldnt find the index files to update. We may try again removing the update flag but I just decided to leave it in prettierignore for this PR since I dont want more changes to build system here

.lintstagedrc.js
daigof: yep, thanks it is much cleaner now. Note that unlike the version used in PocketDerm I had to write the TSC command as a function, like indicated in the Docs https://www.npmjs.com/package/lint-staged

with simple string it doesnt work. Maybe it was a change for  `10.1.3` version.

Anyway tested every type of file leaving errors and spaces  and works fine

src/constants/mediaQueries/index.ts
daigof: Ok so thanks for bringing this up. I remember not being happy about not having intellisense and strong type for mediaqueries.
Due to the rare occasion that we need to add a new breakpoint I just decided to hardcode the Type so that we can have strong type and intellisense.

I left a comment  in breakpoint file that if a new one is added the mediaquery type needs to be udpated

tsconfigBase.json
daigof: yes done

daigof: Also incorporated the new flags from pocketderm repo open PR

PR #149, opened 2019-11-07
[Patch] Sequence onExit callback after setState in Alert component

mjaltamirano:
Development logs a `setState` on unmounted component warning for all `Alerts` when we transition them off the screen.

image

I looked into it and this one-line change fixes the issue. I preferred the diff to using a callback in `setState` (e.g. `this.setState({ exited: true }, () => onExit({ ...rest }));`) to try and change behavior as little as possible.


daigof:
👍

PR #162, opened 2019-12-10
[Patch] Update full button mobile width logic

michaeljaltamirano:
An earlier PR (#129) added full width button styling but did not limit it to only mobile-views. It seems a comment I left in that PR indicates that the full width buttons were only supposed to be for mobile views. 

In any case, there are currently some full width buttons in production, so we can't break that existing behavior. This PR adds a second prop, `noDesktopFullWidth`, with a default value of `false`, to maintain behavior. In places we do not want very wide buttons on non-mobile views, we can pass through a second prop. 

We can consider cleaning up all instances in a couple weeks when workload drops, but in the meantime the scope creep is too much.

### DIFF
--- a/src/shared-components/button/components/roundButton/__snapshots__/test.js.snap
+++ b/src/shared-components/button/components/roundButton/__snapshots__/test.js.snap
@@ -196,8 +196,8 @@ exports[`<RoundButton /> UI snapshots renders with props 1`] = `
   right: 0;
   top: 0;
   margin-top: -6px;
-  opacity: 0;
   width: 38px;
+  opacity: 0;
   width: 36px;
   margin: -3px -3px 0 0;
 }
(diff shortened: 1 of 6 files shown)


src/shared-components/button/components/roundButton/__snapshots__/test.js.snap
```diff
@@ -196,8 +196,8 @@ exports[`<RoundButton /> UI snapshots renders with props 1`] = `
   right: 0;
   top: 0;
   margin-top: -6px;
-  width: 38px;
   opacity: 0;
+  width: 38px;
   width: 36px;
```
michaeljaltamirano: It appears this doubly-applied width already existed, so this diff is just because I alphabetized props, presumably.

michaeljaltamirano:
Closing in favor of button container implementation.

PR #172, opened 2020-01-24
[Icon Updates] Bring over all updated Emojis from Figma

ZeMunchkin:
Asana: https://app.asana.com/0/0/1158576271189802/f

In this PR:
- Replace all current emojis with the updated figma emojis

There is one emoji group that has been renamed to match its name in Figma -- the old emoji group (`Supported` and `SupportedFilled`) will be deprecated once its usage has been updated in PocketDerm and Gatsby


ZeMunchkin:
> My only concern is that we are changing the viewport definition of some emojis, were you able to test this wont have any impact?

I saw this too and was messing around with it somewhat. I can't seem to change the viewport from what it gets exported as, since it cuts off the path of the svg. But when used, all icons get the default height width of 16px if no other height / widths are passed, so it doesn't seem to change the space of the component on the page. Was checking local storybook before / after each icon, and for the ones that have significant visual differences I'm keeping a running list and will be checking places they're used when I update the Gatsby and PocketDerm apps. Generally the Emojis did not have major changes to their visual footprint. The Glyphs and Icons have more drastic differences so I'll be checking a lot more of those.

PR #173, opened 2020-01-27
Feature/upgrade dependencies

cmgmyr:
https://app.asana.com/0/1156667858696586/1156075392616963

This PR upgrades a handful of dependencies:

- ~~Storybook (and related)~~
- Jest (and related)
- Enzyme (and related)

The current security alert for handlebars is tied to the past version of Jest. After the Jest upgrade handlebars is no longer a dependency.

PR #174, opened 2020-01-27
Feature/upgrade storybook

cmgmyr:
https://app.asana.com/0/1156667858696586/1156075392616963

Continuation of #173 

I went through the rest of the upgrade docs to double-check everything. I also found the optional change to rename the storybook files, so I did that as well. Everything appears to build, run and look good locally. LMK if I missed anything.

Note: this is just an optional upgrade since I was in there anyways. We can close this if we don't want to upgrade ATM.


package.json
```diff
@@ -57,14 +57,16 @@
...
-    "@storybook/addon-links": "^5.1.9",
-    "@storybook/addon-storysource": "^5.1.9",
-    "@storybook/addon-viewport": "^5.1.9",
-    "@storybook/react": "~5.1.9",
+    "@storybook/addon-a11y": "^5.3.9",
+    "@storybook/addon-actions": "^5.3.9",
+    "@storybook/addon-centered": "^5.3.9",
+    "@storybook/addon-knobs": "^5.3.9",
+    "@storybook/addon-links": "^5.3.9",
+    "@storybook/addon-storysource": "^5.3.9",
```
michaeljaltamirano: Should we be removing this, now?

daigof:
@cmgmyr Note that I release a beta version for 5.0 that is not working right now. I think the correct way to proceed with these releases is continue working with version 4.x.x

cmgmyr:
@daigof so you are suggesting to downgrade `@storybook/*` to 4.x?

daigof:
> @daigof so you are suggesting to downgrade `@storybook/*` to 4.x?

No, I was not talking about the code, I refer to Radiance release steps, dont go 5.x route when publishing to NPM and github. Use the 4.x.x versions

cmgmyr:
planning to tag this as `v4.2.11` with the other two dev deps from today

PR #182, opened 2020-01-29
[Icon Updates] Add / update 3/3 icons from Figma

ZeMunchkin:
Asana: https://app.asana.com/0/0/1158576271189801/f
Glyph Updates sheet: https://docs.google.com/spreadsheets/d/1FNiVhL3eV_Vc69JZycn4QAz63HOSXBb4RZAsHSc9XdY/edit?usp=sharing

In this PR:
- Updated / add third group of Icons from Figma

Note: Some of the icons have substantial design differences between the old and the new, and for those I will be checking both Gatsby and PocketDerm repos once this is published and the package is ready to be updated. There are also some icons which have been renamed, so the old names will be removed in a later PR once those repos have been updated.


stories/optionButton/optionButtonExample.js
```diff
@@ -37,15 +36,6 @@ class OptionButtonExample extends React.Component {
           type="primary"
           icon={<PlusIcon />}
         />
-        <OptionButton
```
daigof: was thi removed in other PR?

daigof: sry  I refer to the `EqualsIcon`

ZeMunchkin: Just aesthetic. The new `EqualsIcon` is now wrapped in a circle, and so it looks like this:

Screen Shot 2020-01-29 at 1 08 25 PM

It's on my radar to check how it's used in the main repos, but here it was just being used as an example, and it it was looking strange next to the plus and minus icons.

youngmanalive: yeah that looks weird.

PR #187, opened 2020-01-30
[Icon Updates] Remove masks on svgs for glyphs a-m

ZeMunchkin:
Discovered that using the masks from the exported Figma SVGs was causing some issues in some use cases with fills (where the fill would cover the whole square of the SVG instead of just the path).

This PR removes the masks on the  glyphs A-M


src/svgs/glyphs/mortor-glyph.svg
```diff
@@ -1,16 +1,4 @@
...
-    </g>
-  </mask>
-  <g mask="url(#mask0)">
-    <rect width="48" height="48" fill="currentColor"/>
-  </g>
-  <defs>
-    <clipPath id="clip0">
-      <rect x="8" y="10" width="32" height="29" fill="white"/>
-    </clipPath>
-  </defs>
```
youngmanalive: i'm sure you're taking notes on these, but we'll prob want to double check the more complicated ones like this.

ZeMunchkin: Checking as I was updating, they seem unaffected by removing this extra lines, but will definitely check them in the other repos as well.

PR #202, opened 2020-03-11
[New Components] Indicator and Callout

daigof:
- New components: Callout and Indicator
- Review link https://curology-radiance-pr-202.herokuapp.com/
- update prettier rules based on our current standards
- add `md` as auto import extension


michaeljaltamirano:
Were the dependency changes required for this PR? I don’t think we should be sneaking in major version dependency changes into PRs where that isn’t the focus.

daigof:
> Were the dependency changes required for this PR? I don’t think we should be sneaking in major version dependency changes into PRs where that isn’t the focus.

Was trying to get rid of no single export eslint rule. But just  reverted back those bumps, no really needed

daigof:
@benkolde they are at `8px` did you mean some other value?

src/shared-components/callout/style.js
```diff
@@ -0,0 +1,34 @@
...
+import { COLORS, SPACER, TYPOGRAPHY_CONSTANTS } from '../../constants';
+
+export const ParentContainer = styled.div`
+  max-width: 327px;
+`;
+
+export const CalloutContainer = styled.div`
+  background-color: ${COLORS.infoBackground};
+  padding: ${SPACER.medium};
+  border-radius: 8px;
```
daigof: here it sits at `8px`

benkolde: Oh oops, sorry. It just looked big. I'm gonna approve

ida-strom:
for the callout—the icon should always be 48x48px. would be great to add a note to that and change the demos in heroku so the icons are the same size @daigof

PR #204, opened 2020-03-13
Update Components: Button, Chip and Tooltip

daigof:
- **Buttons:** Update all buttons to have border radius: 4px.
- **Chip:** new styles, added default and secondary status. deprecated pending and closed status, but they will fallback to default styles if someone provides an incorrect status
-  **Tooltip:** update styles. Fixed arrow center positioning (now the arrow wont get in way of letters and is perfect centered). New `isSmall` prop for smaller tooltips.

### Other minor fixes 
- Callout: fixed doc to use 48x48 glyph
- Indicator: refactored to use same technique as chip using `display: inline-text` instead of `width:content-fit`
- Edit Storybook build (webpack config) that makes images broken. Compare Avatar component in PROD the images are all broken, this fixes that.

### Preview link
https://curology-radiance-pr-204.herokuapp.com/

### DIFF
--- a/src/shared-components/tooltip/__snapshots__/test.js.snap
+++ b/src/shared-components/tooltip/__snapshots__/test.js.snap
@@ -43,17 +43,16 @@ exports[`<Tooltip /> UI snapshot renders content and children 1`] = `
 }
 
 .emotion-2::after {
-  top: -10px;
-  border-left: 12px solid transparent;
-  border-right: 12px solid transparent;
-  border-bottom: 12px solid #332e54;
+  top: -8px;
+  border-left: 8px solid transparent;
+  border-right: 8px solid transparent;
+  border-bottom: 8px solid #332e54;
   left: 50%;
-  margin-left: -10px;
+  margin-left: -8px;
   width: 0;
   height: 0;
   content: '';
   position: absolute;
-  margin-top: 3px;
   z-index: 4;
 }
 
--- a/src/shared-components/tooltip/__snapshots__/test.js.snap
+++ b/src/shared-components/tooltip/__snapshots__/test.js.snap
@@ -17,6 +17,7 @@ exports[`<Tooltip /> UI snapshot renders content and children 1`] = `
 }
 
 .emotion-6 {
+  max-width: none;
   top: 120%;
   background: #332e54;
   box-shadow: 0px 8px 24px rgba(51,46,84,0.05);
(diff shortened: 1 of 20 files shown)


daigof:
> Minor thing, but for `isSmall` can we add vertical padding of `2px`
> 
> For both, can we make the trigger have `borderRadius: 1` and change the trigger width to `8px`

@benkolde  What is the trigger? did you mean the arrow? I made the arrow 8px dimension now but I dont know how border radius would affect it

benkolde:
> > Minor thing, but for `isSmall` can we add vertical padding of `2px`
> > For both, can we make the trigger have `borderRadius: 1` and change the trigger width to `8px`
> 
> @benkolde What is the trigger? did you mean the arrow? I made the arrow 8px dimension now but I dont know how border radius would affect it

I just want the arrow to look like this 

image

daigof:
@benkolde @ida-strom this is ready for re-review. One thing I noticed is that the tooltip extends to the width of the Trigger element, if wrongly set or unchecked it can go very wide. I suggest putting a `max-width: 327px` 

Unchecked can grow as:
Screen Shot 2020-03-19 at 10 38 12

With Max-width Fix:
Screen Shot 2020-03-19 at 10 37 54

benkolde:
> @benkolde @ida-strom this is ready for re-review. One thing I noticed is that the tooltip extends to the width of the Trigger element, if wrongly set or unchecked it can go very wide. I suggest putting a `max-width: 327px`
> 
> Unchecked can grow as:
> Screen Shot 2020-03-19 at 10 38 12
> 
> With Max-width Fix:
> Screen Shot 2020-03-19 at 10 37 54

Can this be a prop? @daigof I think having the option for it to expand would be great

benkolde:
Can you build a new test environment for this?

daigof:
@benkolde I added a new optional prop for restricted width (327px) with default false value so it doesn't impact existing implementations.

Deployed a review app here https://curology-radiance-pr-204.herokuapp.com/

src/shared-components/tooltip/__snapshots__/test.js.snap
```diff
@@ -16,15 +16,16 @@ exports[`<Tooltip /> UI snapshot renders content and children 1`] = `
...
+.emotion-6 {
+  max-width: none;
   top: 120%;
-  background: #ffffff;
-  border: 1px solid #ededf0;
-  box-shadow: 0 12px 20px 0 rgba(51,46,84,0.05);
-  color: #706D87;
+  background: #332e54;
+  box-shadow: 0px 8px 24px rgba(51,46,84,0.05);
+  border-radius: 0.5rem;
```
benkolde: @daigof can you make the border radius `4px` when tooltip `isSmall`? Then it lgtm!

daigof:
@benkolde  Updated the border radius of small tooltip to 4px.

https://curology-radiance-pr-204.herokuapp.com/?path=/story/tooltip--usage

PR #216, opened 2020-04-07
Allow passing an empty array of messages to VerificationMessages

matt-allan:
Currently passing a `messages` prop with an empty array to `VerificationMessages` causes a TypeError, i.e.

```jsx
 <VerificationMessages messages={ { required: [] } } />
```

This PR adds a filter so we don't attempt to render empty arrays of messages.

See https://github.com/curology/PocketDerm/pull/8609#discussion_r404881354 for more info.

### DIFF
--- a/src/shared-components/verificationMessages/test.js
+++ b/src/shared-components/verificationMessages/test.js
@@ -35,7 +35,6 @@ describe('<VerificationMessages />', () => {
         required: [],
       };
       const wrapper = shallow(<VerificationMessages messages={messages} />);
-      console.log(wrapper.debug());
       expect(wrapper.html().indexOf('li') === -1).toBe(true);
     });
   });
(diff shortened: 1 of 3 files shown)


src/shared-components/verificationMessages/test.js
```diff
@@ -0,0 +1,42 @@
...
+    });
+  });
+
+  describe('when message value is an empty array', () => {
+    it('renders nothing', () => {
+      const messages = {
+        required: [],
+      };
+      const wrapper = shallow(<VerificationMessages messages={messages} />);
+      console.log(wrapper.debug());
```
daigof, line 38:
```suggestion
```

matt-allan:
@daigof merged, if you could publish a release that would be great 😀

PR #219, opened 2020-04-14
[Buttons] Add color prop to buttons

dbushy727:
# Button Colors

TL DR; buttons can be different colors now. Just supply the `buttonColor` prop to your button and Voila!

`<Button buttonColor={COLORS.statusRed}>Click Me!</Button>`

### Check it out here --> [Review App](https://curology-radiance-pr-219.herokuapp.com/?path=/story/buttons--button)
<br>

large

<hr>

Based on the latest design for [Dupe Patients project](https://www.figma.com/file/S4CWI1jVFMxsJZWYDUaNn0/Duplicate-Patient-Support-Experience?node-id=245%3A3103) we have to render a red button. 

This PR adds the ability to supply a color prop to any of the buttons, and it will render the buttons in that color in all the various states. Please note that it is required to use one of the colors from one of the constants files. 

The story has been updated to allow a user to pick any color they like and view it. 

Certain colors are designated alternate colors for various states, while others do not. In this PR, we have linked some colors together and derive other colors where that link does not exist.

For example:
primary -> purple80 for the RoundButton background and text respectively.
statusRed -> lighter desaturated version via tinycolor
Screen Shot 2020-04-15 at 11 46 56 AM
Screen Shot 2020-04-15 at 11 47 18 AM
Screen Shot 2020-04-15 at 11 47 10 AM

Screen Shot 2020-04-15 at 11 51 06 AM

### DIFF
--- a/docs/button/button.md
+++ b/docs/button/button.md
@@ -15,6 +15,20 @@ import { CheckmarkIcon } from 'radiance-ui/lib/icons';
     <Button disabled>Disabled Button</Button>
     <Button isFullWidth>Full Width Button</Button>
 
+    <Button color="statusRed">Primary Color</Button>
+    <Button color="statusRed" buttonType="secondary">
+      Secondary Color
+    </Button>
+    <Button color="statusRed" buttonType="tertiary">
+      Tertiary Color
+    </Button>
+    <Button color="statusRed" buttonType="quaternary">
+      Quaternary Color
+    </Button>
+    <Button color="statusRed" isFullWidth>
+      Full Width Color
+    </Button>
+
     <Button isLoading>Primary Loading</Button>
     <Button isLoading buttonType="secondary">
       Secondary Loading
@@ -48,6 +62,22 @@ import { CheckmarkIcon } from 'radiance-ui/lib/icons';
       Full Width Button
     </Button>
 
+    <Button color="statusRed" icon={<CheckmarkIcon />}>
+      Primary Color
+    </Button>
+    <Button color="statusRed" buttonType="secondary" icon={<CheckmarkIcon />}>
+      Secondary Color
+    </Button>
+    <Button color="statusRed" buttonType="tertiary" icon={<CheckmarkIcon />}>
+      Tertiary Color
+    </Button>
+    <Button color="statusRed" buttonType="quaternary" icon={<CheckmarkIcon />}>
+      Quaternary Color
+    </Button>
+    <Button color="statusRed" isFullWidth icon={<CheckmarkIcon />}>
+      Full Width Color
+    </Button>
+
     <Button isLoading icon={<CheckmarkIcon />}>
       Primary Loading
     </Button>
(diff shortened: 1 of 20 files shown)


src/shared-components/button/constants.js
```diff
@@ -0,0 +1,25 @@
+import { COLORS } from '../../constants';
+
+// TODO: potentially break out these pairings to the color constants file
+export const textColorsAssociatedWithColors = {
```
dbushy727, line 4: wasnt sure if we wanted this in our actual constants/colors file since its kinda pointed to how i was using it. Maybe there's a better way to connect colors together as related.

src/shared-components/button/components/roundButton/style.js
```diff
@@ -62,6 +64,54 @@ export const roundButtonLoader = disabled => css`
...
   margin: 10px 0;
 `;
+
+/**
+ * Given a color as an argument,
+ * determine an alternate color for pairing
+ * @param  string color   the current color name of the round button (e.g purple, primary, etc.)
+ * @return string         hex string of the alternate color (e.g. #efefef)
+ */
+const determineAlternateTextColor = color => {
```
dbushy727, line 78: i didnt put this as a util as its kind of opinionated as to whats a derived alternate color. This works great for the round button color vs text color, but this may not be the formula we want across the board, so i left this as helper functions in the roundbutton styles specifically. Happy to break this out if we need.

dbushy727:
> This looks good. I would add one small thing to the docs that just says we probably want to keep usage of these buttons to be special cases only (at least for the time being)

added a comment to the docs in 14b391ed6cf3e28466ab60847646f4c5011acb9f

`We should generally try to use the default button color when possible. Only for special cases should we need to use a different button color.`

src/shared-components/button/index.js
```diff
@@ -44,6 +46,7 @@ class Button extends React.Component {
   static defaultProps = {
     disabled: false,
     buttonType: 'primary',
+    color: 'purple',
```
daigof, line 49: ok so dont hate me for this but as I pointed out above I think it would be more appropriate if we pass a COLORS constant as prop value it is safer type-wise and make more consistent with our Radiance usage (for example Icons, callout, etc we pass the hex constant value or one of COLORS constants)

docs/button/button.md
```diff
@@ -15,6 +15,20 @@ import { CheckmarkIcon } from 'radiance-ui/lib/icons';
     <Button disabled>Disabled Button</Button>
     <Button isFullWidth>Full Width Button</Button>
 
+    <Button color="statusRed">Primary Color</Button>
```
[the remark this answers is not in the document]
dbushy727: @daigof the problem with doing ${COLORS.statusRed} is that it gives the value, not the key. I need the key name like "statusRed" not "#abc123"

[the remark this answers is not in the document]
dbushy727: ok, i changed the implementation to accept the color value instead. it makes the knobs dropdown work a little weird, but otherwise everything else is fine.

I will say that it feels a bit uncommon to pass a hex argument from an enumerated list rather than key names which is typical of other ui frameworks (semantic, bootstrap, foundation, material, etc)

For the sake of this pr i will make use of the `buttonColor={COLORS.statusRed}` but would def like to talk about the pros and cons offline of this.

docs/button/linkButton.md
```diff
@@ -27,7 +25,7 @@ import { LinkButton } from 'radiance-ui';
       I am Disabled
     </LinkButton>
   </LinkButton.Container>
-</React.Fragment>
+</React.Fragment>;
```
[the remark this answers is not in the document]
dbushy727: fixed

src/shared-components/button/index.js
dbushy727: haha no dont worry, i dont hate you 😄 i get the idea of the switch over. I made the change for now but as mentioned above, would love to talk more about this for future stuff.

dbushy727:
@daigof ok i believe i addressed all your concerns. The only thing that isnt 100% exactly is the story for the RoundButton. the layout (horizontal vs vertical) is different than the rest of the button pages, so i opted to conform to the current page style rather than make it appear like the other pages.

Shemoel1:
100% agree with @benkolde on special case use. Looks good to me otherwise @dbushy727

PR #220, opened 2020-04-16
[Icon Updates] Adds CirclePlusIcon and CircleMinusIcon

ipc103:
* [Figma](https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance-UI?node-id=1682%3A1007)

Including in this PR are two additional Icons, which were designed as part of the Duplicate Patient Workflow. 

### CircleAddIcon
Screen Shot 2020-04-16 at 3 28 00 PM

### CircleRemoveIcon
Screen Shot 2020-04-16 at 3 28 11 PM

### IconLibrary
https://curology-radiance-pr-220.herokuapp.com/?path=/story/icon--icon-library


src/svgs/icons/index.js
```diff
@@ -9,7 +9,9 @@ export { default as CameraIcon } from './camera-icon.svg';
 export { default as ChatBubbleIcon } from './chat-bubble-icon.svg';
 export { default as CheckmarkIcon } from './checkmark-icon.svg';
 export { default as ChevronIcon } from './chevron-icon.svg';
+export { default as CircleAddIcon } from './circle-add-icon.svg';
 export { default as CircleCheckmarkIcon } from './circle-checkmark-icon.svg';
+export { default as CircleRemoveIcon } from './circle-remove-icon.svg';
```
dbushy727, line 14: [tiny nit] maybe this should be CircleMinusIcon since its like the minus icon but has a circle around it. Add is more the action and the minus is more of the composition.

Also may apply to CircleAddIcon -> CirclePlusIcon

ipc103: Yeah, that's a good thought. For reference, Semantic UI uses "Plus Circle" and "Minus Circle" to describe similar icons. https://semantic-ui.com/elements/icon.html @Shemoel1 any thoughts?

ipc103:
👍 - will go ahead and make that change since we seem to have consensus. @Shemoel1 let us know if you disagree or have other thoughts and we can discuss further.

Shemoel1:
@ipc103 👍

ipc103:
Thanks gang. @daigof I know you have a release coming up. Am I okay to publish this change as `6.4.1`?

daigof:
yes please release it as 6.4.1 same process thanks

PR #229, opened 2020-04-24
[Alert component] add truncateText prop

daigof:
- https://app.asana.com/0/1126113002004936/1172672399123690/f
- enabled by default, may be disabled per specific case

### DIFF
--- a/docs/alert.md
+++ b/docs/alert.md
@@ -31,8 +31,8 @@ import { Alert } from 'radiance-ui';
   <Alert
     content={
       <Fragment>
-        <strong>Whoops!</strong> There was an error updating your address, pleas
-        try again later
+        <strong>Whoops!</strong> There was an error updating your address,
+        please try again later
       </Fragment>
     }
     type="error"
@@ -61,8 +61,8 @@ import { Alert } from 'radiance-ui';
   <Alert
     content={
       <Fragment>
-        <strong>Whoops!</strong> There was an error updating your address, pleas
-        try again later
+        <strong>Whoops!</strong> There was an error updating your address,
+        please try again later
       </Fragment>
     }
     type="error"
--- a/docs/alert.md
+++ b/docs/alert.md
@@ -8,7 +8,7 @@ We should avoid using alerts to show flow-level success states (i.e. Cancellatio
 
 Alerts contain some visual marker, an icon `16px` or avatar `32px`
 
-Alerts should have a **bolded** portion of the text that helps inform the user what the alert is about. The text should be short and preferably not extend more than 2 lines. Ellipses (...) will be used to truncate past 2 lines by default. Can be turned off via the prop `truncateText`.
+Alerts should have a **bolded** portion of the text that helps inform the user what the alert is about. The text should be short and preferably not extend more than 2 lines. Ellipses (...) will be used to truncate past 2 lines by default. Truncate functionality can be turned off via the prop `truncateText`.
 
 Alerts can contain a CTA section. These should be reserved for really important actions. These alerts persists until an action is taken.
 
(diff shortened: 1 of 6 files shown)


docs/alert.md
```diff
@@ -8,7 +8,7 @@ We should avoid using alerts to show flow-level success states (i.e. Cancellatio
 
 Alerts contain some visual marker, an icon `16px` or avatar `32px`
 
-Alerts should have a **bolded** portion of the text that helps inform the user what the alert is about. The complete text should be short and not extend more than 2 lines. Ellipses (...) will be used to truncate past 2 lines.
+Alerts should have a **bolded** portion of the text that helps inform the user what the alert is about. The text should be short and preferably not extend more than 2 lines. Ellipses (...) will be used to truncate past 2 lines by default. Can be turned off via the prop `truncateText`.
```
michaeljaltamirano, line 11:
```suggestion
Alerts should have a **bolded** portion of the text that helps inform the user what the alert is about. The text should be short and preferably not extend more than 2 lines. Ellipses (...) will be used to truncate past 2 lines by default. Truncate functionality can be turned off via the prop `truncateText`.
```

stories/alert/index.js
```diff
@@ -31,7 +31,7 @@ stories.add(
           content={
             <Fragment>
               <strong>Whoops!</strong> There was an error updating your address,
-              pleas try again later
+              please try again later
```
michaeljaltamirano, line 34: Similar typo on L64

src/shared-components/alert/test.tsx
```diff
@@ -1,65 +1,62 @@
...
 
-const testAlert = {
-  text: 'Your email address was updated successfully!',
-  type: 'success',
-  duration: 'sticky',
-};
+// Note on truncateText prop test: this cannot be tested because element scrollHeight and offsetHeight are not simulated correctly
+
+const alertText =
+  'Lorem ipsum, dolor sit amet consectetur adipisicing elit. Ex quis dicta iusto a placeat!';
```
michaeljaltamirano, line 9: Are we using this because it's longer than `Your email address was updated successfully!`?

stories/alert/index.js
daigof: fixed

michaeljaltamirano:
@daigof I looked at the Asana ticket. Did you have any other conversations with Ben about this? It's a little confusing because he says `I just don't know the specific instances where we have text getting cut off.` but also filed the ticket, so presumably there are some places we want to follow-up in app usage to make the copy more concise, right? Or to add `truncateText={false}`.

src/shared-components/alert/test.tsx
daigof: yes that was the idea, 80% of the time on this PR was trying to make a truncateText test, but found out that those dom props like scrollHeight that we use in the Component are all hardcoded to 0 in tests so desisted (we probably can make it work but not worth the effort since we have a real test example in story anyway)

I can change this lorem text to `Your email address was updated successfully!` now

src/shared-components/alert/test.tsx
```diff
@@ -1,65 +1,62 @@
...
+const CustomContentComponent = (): JSX.Element => (
   <div>
     <strong>Your info is updated!</strong>{' '}
-    <a href="http://google.com"> Click here to see changes</a>
+    <a href="curology.com"> Click here to see changes</a>
   </div>
 );
 
-const createNodeMock = element => {
+const createNodeMock = (element: ReactElement): any => {
```
michaeljaltamirano, line 17:
```suggestion
const createNodeMock = (element: ReactElement): { innerHTML: string } | null => {
```

(I'm also comfortable just leaving returns like this not explicitly typed.)

daigof:
@michaeljaltamirano By Ben's comment this may be just created by him preemptively. I left a comment if they know an example in the app. I can leave a comment in the hax channel for devs to be aware of the alert texts from now on.

 Cases may arise naturally in the near future for the use of this prop anyway.

PR #231, opened 2020-04-28
[Alerts] use dontTrucateText flag

daigof:
- To be more consistent whit our other flags usage it is best to place flags as true when you want to override some default behaviour so I changed the way we check the flag.

When you want full text scenario:

Before `<Alert truncateText={false} ...`

After `<Alert dontTrucateText ....`

This has no impact in Curology since the flag is not being used yet, I was motivated to do this because of the messy code I was doing in https://github.com/curology/PocketDerm/pull/8978 with truncateText={false}


daigof:
I agree it is better `preventTextTruncating `  so changed.

PR #237, opened 2020-05-11
[Inmmersive modal]  Redesign

daigof:
- Cycle 4 Project: https://www.notion.so/curology/Pitch-New-Modal-UX-ec4d81063d13413aa299559df343a313
- This PR covers ImmersiveModals only
- Refactored how keycodes work: removed keyIdentifier (only supported in Safari 5 and below Ancient) in favor of key and keyCode

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyIdentifier


daigof:
@cmgmyr **ok I am glad the tests failed in circleCi**. it looks like I had a corrupted `node_modules`. I removed it and run yarn install and it reverted back the snapshots to they original state. I thought snapshots changed because of the updated transition library, but no they are no longer part of this PR. So no snapshot changes now.

 I wasn't paying much attention to circle CI checks because they were failing like 3 or 4 times each time in build due to `Prospective Merge` or `ENOEM` errors and had to reset each time.

daigof:
It also worth noticing that it seems we run on Node 10 for Radiance. I was using Node 12, we have to use `nvm`  to develop in the correct Prod version.

PR #239, opened 2020-05-12
Exclude deprecated SPACING constant from story

youngmanalive:
This excludes the deprecated `SPACING` constant from the story to avoid confusion.

I also filtered out the `base` property from `SPACER` since we should be only be using `medium`. I've seen some confusion over that.

I haven't looked if `SPACING` or `base` are still in use in the other repos (is/was `base` ever used?), this was just a quick patch so folks aren't using the wrong ones. Might be time to just pull them out completely.

Screen Shot 2020-05-12 at 12 21 12 PM

### DIFF
--- a/stories/constants/index.js
+++ b/stories/constants/index.js
@@ -2,9 +2,9 @@ import React from 'react';
 import { withDocs } from 'storybook-readme';
 import { storiesOf } from '@storybook/react';
 import { css } from '@emotion/core';
-
 import * as CONSTANTS from 'src/constants';
 import { Typography } from 'src/shared-components';
+// eslint-disable-next-line import/extensions
 import ConstantsReadme from 'docs/constants.md';
 
 import renderConstantsMap from './renderConstantsMap';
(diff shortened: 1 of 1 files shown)


stories/constants/index.js
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import { withDocs } from 'storybook-readme';
 import { storiesOf } from '@storybook/react';
 import { css } from '@emotion/core';
-
 import * as CONSTANTS from 'src/constants';
 import { Typography } from 'src/shared-components';
+// eslint-disable-next-line import/extensions
 import ConstantsReadme from 'docs/constants.md';
```
daigof, line 8: `md` as been added as known extension so you can remove the `.md` and remove the eslint disable comment here

youngmanalive: yeah it wouldn't let me commit without it. is it because its not a typescript file?

Screen Shot 2020-05-12 at 1 05 58 PM

youngmanalive:
@daigof I think I'm actually just gonna close this PR and go ahead and remove the constants. Seems silly to patch/hide it instead of just getting rid of them 😅.

 I'll do another check on everything so we're sure they're not used anywhere.

daigof:
AWESOME yeah that sounds good, I dont remeber why I left SPACING as
deprecated, I was new and afraid I think. but I support the idea of getting
rid of it now I dont see any reason why not now

youngmanalive:
closing in favor of #241

PR #244, opened 2020-05-29
Use circle icon for selected radio buttons instead of a checkmark

timmartin19:
A checkmark implies that the radio button can be unselected.  Radio
buttons cannot be unselected.  A white circle is a more consistent
pattern for radio buttons.

Checkboxes will still have a white checkmark.

Also, made some updates to pass the linter.

Screen Shot 2020-05-29 at 10 55 50 AM

### DIFF
--- a/src/shared-components/selectorButton/index.js
+++ b/src/shared-components/selectorButton/index.js
@@ -69,6 +69,8 @@ const SelectorButton = ({
       onKeyPress={onClick}
       tabIndex="0"
       selector={selector}
+      role={selector}
+      aria-checked={checked}
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...rest}
     >
--- a/src/shared-components/selectorButton/index.js
+++ b/src/shared-components/selectorButton/index.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import {css} from '@emotion/core';
+import { css } from '@emotion/core';
 
 import CheckmarkIcon from '../../svgs/icons/checkmark-icon.svg';
-import {COLORS} from '../../constants';
+import CircleSolidIcon from '../../svgs/icons/circle-solid-icon.svg';
+import { COLORS } from '../../constants';
 import {
   OuterContainer,
   Selector,
@@ -26,10 +27,12 @@ const propTypes = {
 };
 
 const defaultProps = {
-  onClick: () => {
-  },
+  onClick: undefined,
   type: 'primary',
   selector: 'radio',
+  icon: undefined,
+  children: undefined,
+  size: undefined,
 };
 
 const SelectorButton = ({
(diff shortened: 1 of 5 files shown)


src/shared-components/selectorButton/index.js
```diff
@@ -41,38 +44,52 @@ const SelectorButton = ({
...
+        `}
+        width={16}
+        height={16}
+      />
+    );
+  return (
+    <OuterContainer
+      onClick={onClick}
+      onKeyPress={onClick}
+      tabIndex="0"
```
ipc103, line 70: This may be out of scope, but it might be a good idea to add a `role` of `checkbox` and an `aria-checked` attribute to help assistive technologies determine the state of the selector. 

```jsx
role={selector} // 'radio' or 'checkbox'
aria-checked={checked}
```

I'm very much a beginner when it comes to accessibility, so definitely possible I'm missing other stuff that would be helpful too.

src/shared-components/selectorButton/test.js
```diff
@@ -175,5 +175,20 @@ describe('<SelectorButton />', () => {
       wrapper.simulate('click');
       expect(spy).toHaveBeenCalled();
     });
+
+    it('Does nothing when no onClick is set', () => {
+      const wrapper = shallow(<SelectorButton checked={false} />);
```
ipc103, line 180: This is helpful - I didn't realize that you could pass `undefined` as a React event handler without breaking stuff.

src/shared-components/selectorButton/index.js
```diff
@@ -26,10 +27,12 @@ const propTypes = {
 };
 
 const defaultProps = {
-  onClick: () => {
-  },
+  onClick: undefined,
   type: 'primary',
   selector: 'radio',
+  icon: undefined,
+  children: undefined,
```
ipc103, line 34: [nit/optional] - TS should now be available in this repo, so if you wanted to, you could convert this file to TS and type the component out as a `React.FunctionComponent`. That should handle the `children` definition so you don't have to default it as `undefined`. Totally your call as converting to TS may cause additional ripple effects.

src/shared-components/selectorButton/test.js
timmartin19: I wasn't positive (hence the test).  But if you don't pass an argument it's undefined so I figured that would be fine here (and eslint was getting angry about an empty arrow function).

src/shared-components/selectorButton/index.js
timmartin19: I'm going to hold off on this, mostly because I don't feel totally comfortable in radiance yet and I don't want to break everything that uses it.

src/shared-components/selectorButton/test.js
ipc103: Yeah, that makes sense. To think of all those hours I've wasted defining empty arrow function 🤦

daigof:
@timmartin19 You can release a new Radiance version, but hold on updating in Curology repo beacuse we are working on adapting the new DialogModal should be finished by next week
@cmgmyr fyi

PR #246, opened 2020-06-19
Support effective tree shaking with ESM

timmartin19:
Webpack was not able to effectively tree shake when everything was
included in one module.  This update makes it easier for webpack to tree
shake.  It _should_ be backwards compatible because we updated the index
for the ESM module in `package.json`.  However, we were packaging our
own version of `React` and `ReactDOM` in the package.  This caused the
bundle size to get quite large.  Moving these to peer dependencies
reduces the size considerably.

Note, I did my best to test this locally with yarn, but I plan on doing
a pre release and testing with that before actually releasing this as a
proper version.


package.json
```diff
@@ -120,11 +120,12 @@
     "react-test-renderer": "^16.8.6",
     "regenerator-runtime": "^0.12.1",
     "rimraf": "^2.6.2",
-    "rollup": "^0.66.6",
+    "rollup": "^2.16.1",
```
michaeljaltamirano, line 123: We’ve held off on upgrading rollup because there’s a bunch of unsupported peerDependencies—did you see anything strange on your end?

package.json
```diff
@@ -120,11 +120,12 @@
...
     "rollup-plugin-babel": "^4.0.3",
     "rollup-plugin-commonjs": "^9.2.0",
     "rollup-plugin-node-resolve": "^3.4.0",
-    "rollup-plugin-size-snapshot": "0.7.0",
+    "rollup-plugin-peer-deps-external": "^2.2.2",
```
michaeljaltamirano, line 127: Perhaps this answers my question—need to look at it

package.json
timmartin19: Unsupported peer dependencies of what exactly?  The plugin I added just requires that a user has `React` and `ReactDom` installed alongside radiance-ui.  e.g. we're not going to bundle it within radiance. (For the ESMs  anyways)

michaeljaltamirano: For example `rollup-plugin-node-resolve ` now exists under the `@rollup` namespace, I think v2 moved most of the supporting modules there.

timmartin19: I see.  I'll take a look and see if there's anything else that needs updating.

michaeljaltamirano:
> Note, I did my best to test this locally with yarn, but I plan on doing
> a pre release and testing with that before actually releasing this as a
> proper version.

Diego and I have made liberal use of `alpha` tags [(example)](https://github.com/curology/radiance-ui/releases/tag/v8.0.0-alpha.0), which you can feel free to copy.

michaeljaltamirano:
It _should_ be backwards compatible because we updated the index
> for the ESM module in `package.json`. 

Here are the outputs, side-by-side:

Screen Shot 2020-06-19 at 4 24 36 PM

I would have expected `bundle-es` to not have `index.js` nested in `src`, but I suppose the inclusion of `node_modules` (and `_virtual`?) might've made that not so straightforward. 

We could just alias our usage of `radiance-ui` to `radiance-ui/dist/src` (at least for testing, to see if it works as expected?), but we might want to try and replicate the simple folder structure as much as possible.

michaeljaltamirano:
@timmartin19 This is the step in the right direction, and sorely needed. Please tag me for re-review when you think it is ready again! (Presumably after pulling into our main repo and playing around with it, but feel free to continue to @ me in comments for review.)

timmartin19:
@michaeljaltamirano  The `module` key in package.json should result in ESM imports being handled the same as before (e.g. you can still do `import {Button} from 'radiance-ui'`).  I tested that locally with `yarn add file:...`

michaeljaltamirano:
> @michaeljaltamirano The `module` key in package.json should result in ESM imports being handled the same as before (e.g. you can still do `import {Button} from 'radiance-ui'`). I tested that locally with `yarn add file:...`

That sounds correct to me. I just didn't register that diff. Thanks!

src/shared-components/container/index.js
```diff
@@ -1,18 +1,3 @@
-import PropTypes from 'prop-types';
-
-import {
-  Container,
-  Section,
-  Divider,
-  Image,
-} from './style';
-
-Container.Section = Section;
```
timmartin19, line 10: This was causing issues because `Rollup` was turning this into

    import {Container, Section Divider, Image } from 'src/shared-components/container/styles';
    import 'src/shared-components/container';

Webpack was tree shaking the `import 'src/shared-components/container'` since we marked the package as side effect free.  Basically, any mutations like this need to happen before exporting.  I went through the rollup code to see if I could find any other potential issues and I couldn't find anything.

rollup.config.js
```diff
@@ -63,15 +39,49 @@ export default {
...
-    '@emotion/styled',
-    'prop-types',
-    'react',
-    'react-modal',
-    'react-slick',
-    'react-toggle-button',
-    'react-transition-group',
-    'tinycolor2',
-  ],
+  external: [...Object.keys(pkg.dependencies), '@emotion/styled-base'],
```
timmartin19, line 42: This should be a bit more sustainable.  We're telling rollup that all of the dependencies will be installed alongside this package.  This prevents the problem of duplicating `react-dom` within the package for the other versions.  Also `@emotion/styled-base` is necessary in addition because `Rollup` effectively bypasses some intermediate imports from `@emotion/styled` to pull directly from `@emotion/styled`.

timmartin19: With `preserveModules` it will try to create a `node_modules` folder in the dist and then import from there (with a relative path).  But those get filtered out in the publish.  Telling rollup that these are all external lets rollup know that it should be able to just use `import ReactDom from 'react-dom'` and that will be available.

timmartin19:
Screen Shot 2020-06-23 at 10 36 10 AM
Screen Shot 2020-06-23 at 10 35 49 AM

timmartin19:
I tested an alpha version against the `PocketDerm` repo here https://github.com/curology/PocketDerm/pull/9664

rollup.config.js
michaeljaltamirano: `Also @emotion/styled-base is necessary in addition because Rollup effectively bypasses some intermediate imports from @emotion/styled to pull directly from @emotion/styled.` Can we memorialize this as a code comment?

src/shared-components/container/style.js
```diff
@@ -2,6 +2,7 @@ import styled from '@emotion/styled';
 import { css } from '@emotion/core';
 
 import { SPACER, BOX_SHADOWS, COLORS, MEDIA_QUERIES } from '../../constants';
+import PropTypes from 'prop-types';
```
michaeljaltamirano, line 5: This is pulling up an `eslint` error for me--did the commit hooks not catch it? We should group it up with the other absolute imports.

src/shared-components/container/index.js
michaeljaltamirano: Yep, this looks like the only instance. Good catch!

src/shared-components/container/style.js
timmartin19: I'm not totally sure what happened there...

rollup.config.js
michaeljaltamirano: @timmartin19 Is your sense that including `react` but not `react-dom` before was an oversight?

timmartin19: Yeah, I think we just started using react-dom for the portal function in the newer dialogs.  I think we already had `react-dom` in the dev dependencies but it wasn't strictly necessary for production code until we added the portal.

PR #248, opened 2020-06-24
[Typescript] Refactor some components

daigof:
First Batch of components updates.
- Disable @typescript-eslint/explicit-function-return-type this removing all return types relying on TS inference
- Component Prop TS type: marking as non optional if they have DefaultProp (should be 100% of cases anyway). Consuming app will not require you to pass these props

### Long Term Objective:

Convert all component (not toooo difficult IMO just need a little time) and then deal with how to integrate full support of Components Types in PocketDerm


src/shared-components/alert/index.tsx
```diff
@@ -46,30 +59,35 @@ class Alert extends React.Component {
...
 
-  constructor(props) {
-    super(props);
-    this.contentText = React.createRef();
-  }
+  static Container = ({ children }: childrenProp): JSX.Element => (
+    <AlertsContainer>{children}</AlertsContainer>
+  );
+
+  contentText = React.createRef<HTMLDivElement>();
```
daigof, line 69: removed constructor in favor of property initialization

src/shared-components/alert/index.tsx
```diff
@@ -25,11 +25,24 @@ const alertIconMapping = {
...
-  static Container = ({ children }) => (
-    <AlertsContainer>{children}</AlertsContainer>
-  );
+type AlertProps = {
+  avatarSrc?: string;
+  content: React.ReactNode;
+  ctaContent?: React.ReactNode;
+  duration?: string | number;
+  truncateText: boolean;
+  type: 'success' | 'error' | 'default';
```
daigof, line 34: TS doesnt recognize defaultProps, so to avoid having to handle undefined and other verbose methods I just mark the prop as required

stories/accordion/index.tsx
```diff
@@ -1,7 +1,5 @@
 import React from 'react';
-// @ts-ignore TODO-TS: remove this when updating storybook version since newest includes types
```
daigof, line 2: not needed anymore!!

src/shared-components/alert/index.tsx
```diff
@@ -78,36 +96,34 @@ class Alert extends React.Component {
...
       window.clearTimeout(this.timer);
     }
   }
 
-  alertExitHandler = () => {
+  alertExitHandler = (): void => {
     const { onExit, ...rest } = this.props;
     this.setState({ exiting: true });
 
-    // eslint-disable-next-line no-undef
```
daigof, line 99: not need anymore, I added window in eslint config as global before

src/shared-components/alert/index.tsx
```diff
@@ -25,11 +25,24 @@ const alertIconMapping = {
...
+  type: 'success' | 'error' | 'default';
+  onExit?: (rest: Omit<AlertProps, 'onExit'>) => void;
+};
+
+type AlertState = {
+  exiting: boolean;
+  exited: boolean;
+};
+
+type childrenProp = { children: React.ReactNode };
```
michaeljaltamirano, line 43: Capitalize `childrenProp` to `ChildrenProp` both in definition and usage. (Alternatively, I would just inline the definition myself.)

src/shared-components/alert/index.tsx
```diff
@@ -78,36 +96,34 @@ class Alert extends React.Component {
...
 
-      onExit({ ...rest });
+      if (onExit) {
+        onExit({ ...rest });
+      }
     }, ANIMATION_DELAY);
   };
 
-  render() {
+  render(): JSX.Element | null {
```
michaeljaltamirano, line 126: My personal preference, in most cases, is the only define the return type if it is unexpected, like this. Did you define the `void` return types above in other lifecycle methods because of eslint errors, warnings, or just a preference for completeness?

src/shared-components/alert/index.tsx
```diff
@@ -46,30 +59,35 @@ class Alert extends React.Component {
...
-      while (contentElement.scrollHeight > contentElement.offsetHeight) {
-        wordsArray.pop();
-        // eslint-disable-next-line prefer-template
-        contentElement.innerHTML = wordsArray.join(' ') + ' ...';
+      if (contentElement) {
+        const wordsArray = contentElement.innerHTML.split(' ');
+        while (contentElement.scrollHeight > contentElement.offsetHeight) {
+          wordsArray.pop();
+          // eslint-disable-next-line prefer-template
+          contentElement.innerHTML = wordsArray.join(' ') + ' ...';
```
michaeljaltamirano, line 89: Should we just change this to using template strings while we're here? contentElement.innerHTML = \`${wordsArray.join(' ')} ...\`;

src/shared-components/alert/index.tsx
daigof: eslint warning. I see in Curology it is turned off. do you think I can apply that rule here as well?
`  "@typescript-eslint/explicit-function-return-type": "off",`

src/shared-components/alert/index.tsx
```diff
@@ -78,36 +96,34 @@ class Alert extends React.Component {
...
 
-    // eslint-disable-next-line no-undef
     window.clearTimeout(this.timer);
 
-    // eslint-disable-next-line no-undef
     window.setTimeout(() => {
       this.setState({ exited: true });
 
-      onExit({ ...rest });
+      if (onExit) {
```
michaeljaltamirano, line 120: If we're doing a truthy check for `onExit` now, due to typing, I think I'd prefer we type it in `defaultProps` as `undefined`. (This is consistent with our TypeScript type of `onExit?` as well.) I didn't see any typing issues in my terminal when doing that.

src/shared-components/alert/index.tsx
michaeljaltamirano: @daigof Yeah we can turn it off 🙂  I think there's a time it's helpful but all the yellow highlight and terminal noise isn't all that helpful.

src/shared-components/alert/index.tsx
michaeljaltamirano: This is better anyway since we should've been consistent and setting `state` in the constructor as well 😅

src/shared-components/alert/style.ts
```diff
@@ -50,20 +50,23 @@ const errorAlertStyles = css`
   box-shadow: 0px 8px 24px rgba(189, 32, 15, 0.05);
 `;
 
-export const AlertContainer = styled.div`
+export const AlertContainer = styled.div<{
+  exiting: boolean;
+  alertType: string;
```
michaeljaltamirano, line 55: It might be preferable to export `type AlertType = 'success' | 'error' | 'default' used both in the props of the component and also here, to be more specific than `string`.

src/shared-components/banner/index.tsx
```diff
@@ -19,7 +19,15 @@ const bannerIconMapping = {
...
-const Banner = ({ content, type, onClick }) => {
+export type BannerTypes = 'default' | 'success' | 'error' | 'danger';
+
+type BannerProps = {
+  content: React.ReactNode;
+  type: BannerTypes;
+  onClick: () => void;
+};
+
+const Banner = ({ content, type, onClick }: BannerProps): JSX.Element => {
```
michaeljaltamirano, line 30: It's a little repetitive but we could probably make this more accurate by adding ES6 defaults here in addition to the defaultProps:

```
type BannerProps = {
  content: React.ReactNode;
  type?: BannerTypes;
  onClick?: () => void;
};

const Banner = ({ content, type = 'default', onClick = () => undefined }: BannerProps): JSX.Element => {
```

Otherwise we'll throw errors when using Banner without providing either of those, which doesn't seem like our intent based on the `defaultProps`.

daigof: Wait, this is important you are right. So my logic of typing props as required to avoid handling undefined state is wrong. I wasn't thinking on this as exported Library. This solves the case for FunctionComponent Ill figure out how to handle class based props

src/shared-components/alert/index.tsx
daigof: This thinking is wrong, If prop is Not required then it must be marked as optional and viceversa in other words I need to honor Proptypes required/non required logic in TS

michaeljaltamirano: You could try setting up a separate method like so:

```
getProps = () => {
 // Return props with return type information that includes those provided by defaultProps
}
```

And then use `const { ... } = this.getProps()` to get the type information we want. We use a similar pattern in PocketDerm for `mobx` `inject` in a few components, look up `injectedProps` in that repo. 

It would be nice to get a method that doesn't add actual code overhead though. 🙈

src/shared-components/alert/index.tsx
daigof: will apply same rule as curology for consistency

src/shared-components/alert/index.tsx
daigof: No worries, just tested another solution: Making all TS prop type that have defaultProps as Not optional in TS and it works. It looks like TS is smart enough that if you use a component it wont throw an error if it has optional defaultProp and is marked as not optional in TS. this is clean and is what it is being used in the wild it seems

daigof: I standarize the code to this solution. On the other hand the tests Im doing in test app are great, it marks the props as required/not required correctly and with intellisense end everything, this is promising

michaeljaltamirano: Yep--however, this means that in consuming components, we will throw errors if those props are not provided, even though there are defaultProps present. 

I think it is actually preferable to have the low-cost conditional checks present, rather than have to update the API of all consuming components that currently rely on defaultProps.

daigof: ok the wording is confusing, but I meant to say I did tested in consumig app (like Curology) and it does not throw an error in fact all Component Props work correctly . Looks like TS figure this out automatically (not optional but has defaultProps)

michaeljaltamirano: Oh cool. Yeah maybe the types for Class components are smart enough to infer from what's provided, along with the defaultProps of the class, to figure that out. Just tested on something in the main app as well. Yay!

src/shared-components/banner/index.tsx
daigof: Our new convention is Optional Prop but has DefaultProp then mark it as not optional in TS. It works for Functional Components as well so I will leave it without the ES6 default and mark all as TS not optional types

daigof: tested with this component and works

PR #249, opened 2020-06-24
[PropTypes] Exclude deprecated values from checker

michaeljaltamirano:
Color warnings are being logged throughout all applications that rely on `radiance-ui` because the PropTypes checker is triggering the warning. This PR sets up a separate prop type to be used in place of iterating over all the possible color values. 

My `radiance-ui` environment isn't currently building storybook, where I would expect the same warnings to trigger, but I pulled in this modified code into an application using it and fixed the warnings.


src/constants/colors/index.ts
```diff
@@ -149,17 +152,34 @@ export const guideColors = {
...
+};
+
 const deprecatedProperties = {
   purple80: 'purple80 is deprecated. Use purple85 instead',
   purple60: 'purple60 is deprecated. Use purple70 instead',
 };
 
+// Do not include deprecated properties in PropTypes or it will trigger warning
+// eslint-disable-next-line
+const { purple80, purple60, ...allOtherLegacyColors } = legacyColors;
```
michaeljaltamirano, line 174: I'm open to setting up this spreading and sharing however reviewers think is clearest.

LilCare: I think it's clear the way it's written.

daigof:
@michaeljaltamirano logic looks legit but it is failing in my storybook, Im looking into it 

> @michaeljaltamirano logic looks legit but it is failing in my storybook, Im looking into it

@michaeljaltamirano the fix is to add
```
  if (category === 'default' || category === 'COLORS_PROP_TYPES') {
        return null;
      }
```
to `colorsStory.js` since it is iterating over all exported members of colors

I will look now in my Radiance test project to see how this new exported member would affect TS types generation

stories/constants/colorsStory.js
```diff
@@ -22,7 +21,7 @@ const ColorsContainer = styled.div`
 const colorsStory = () => (
   <MainContainer>
     {Object.keys(COLORS).map(category => {
-      if (category === 'default') {
+      if (category === 'default' || category === 'COLORS_PROP_TYPES') {
```
daigof, line 24: Pushed this fix. Test in real app works good, we do have the new export `COLORS_PROP_TYPES` in the constants (intellisense, etc) now but it's ok

michaeljaltamirano: @daigof Constants intellisense for `PocketDerm` or this repo?  If PocketDerm we can remove the prop export from the index file, and just import it directly. Would that help?

daigof: as of now it is an exported member so we have it available in PocketDerm. `remove the prop export ` would help but I didnt consider it a big deal, note that if you opt to remove the export then we may not need my fix anymore

PR #252, opened 2020-06-26
[Typescript] Convert Button Components

daigof:
### Changelog
- Reorder propTypes and types alphabetically
- In order to accept ...rest props for buttons components I did:
   - added [key: string]: any; as TS type
   - mark all non required props as TS optional (we said this isn't necessary if they have default props but it is the only way to make rest parameter work in consuming apps, tested and works)
   - put ES6 default values to props because sadly TS doesn't recognise defaultProps and treat them as undefined otherwise
- added TextButton test
- DRY the `isLoadingPropFunction` PropType check
- cleanup old comments and eslint rules no longer needed


src/shared-components/button/index.tsx
```diff
@@ -10,66 +10,74 @@ import LinkButton from './components/linkButton';
...
+  deprecatedProperties,
+  isLoadingPropFunction,
+} from './deprecatedPropsHanlder';
 
-const deprecatedProperties = {
-  loading: "The 'loading' prop is deprecated. Use 'isLoading' instead.",
-};
+export type ButtonType = 'primary' | 'secondary' | 'tertiary' | 'quaternary';
+// This is needed for RoundButton
+export type ButtonTypeWithAction = ButtonType | 'action';
```
michaeljaltamirano, line 20: Shouldn't we define this in the `RoundButton` component? Or were there some complications with that.

daigof: fair comment, I left it here because it is used in main styles and also Loader component, also thought leaving these 2 side by side would be more clear?

michaeljaltamirano: That makes sense. I might rephrase the comment as:

```
/**
 * RoundButton can also accept a buttonType prop of 'action'
 */
```

But I'm also fine if you just resolve this thread without any changes!

daigof: updated the comment, thanks for wording

stories/button/linkButton/index.tsx
```diff
@@ -9,13 +9,20 @@ import { COLORS } from 'src/constants';
...
-/* eslint-disable */
-const MockLink = ({ to, children, ...rest }) => (
+const MockLink = ({
+  to,
+  children,
+  ...rest
+}: {
+  to: string;
+  children: React.ReactNode;
+  rest: any[];
```
michaeljaltamirano, line 19: Since this is a rest operator within named parameters the shape is the remaining props within a single object, rather than an array, so let's re-type this to `{ [key: string]: unknown }`.

stories/button/container.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { css } from '@emotion/core';
+import { Button, Typography } from 'src/shared-components';
+
+// eslint-disable-next-line react/prop-types
+export default ({
```
michaeljaltamirano, line 6: Let's avoid `export default`--it hurts discoverability and also results in anonymous components in the React devtools. 

```
const Container = ({}).

export default Container;
```

(It also looks like we can remove the eslint disable).

src/shared-components/button/style.ts
```diff
@@ -138,14 +147,23 @@ function parseTheme(disabled, buttonType, isLoading, buttonColor) {
   }
 }
 
+type BaseButtonStylesTypes = {
+  disabled: boolean;
+  buttonType: ButtonTypeWithAction;
+  buttonColor: string;
+  isLoading?: boolean;
```
michaeljaltamirano, line 154: Should we enforce this prop more strictly (required) on the dev side to encourage better handling of the different button states?

michaeljaltamirano: Resolving comment--this is potentially options since we're using the same function for LinkButtons too.

michaeljaltamirano:
Will finish reviewing later today.

src/shared-components/button/components/linkButton/index.tsx
```diff
@@ -3,46 +3,59 @@ import PropTypes from 'prop-types';
...
   as: 'a',
+  buttonColor: COLORS.primary,
+  buttonType: 'primary',
+  disabled: false,
   onClick: () => undefined,
   textColor: '',
 };
 
+type LinkProps = {
+  as?: React.ReactNode;
```
michaeljaltamirano, line 35: We can remove the TS-ignore if we replace this type with `as?: 'a' | 'span' | React.ElementType;`. (`ReactNode` is very permissive, and usually only appropriate for things like `children`)

It appears that, with the types, we're able to infer that `'a'` and `'span'` are used as `React.ElementType`. Adding `'a' | 'span'` is helpful, I think, for the API on the consuming end, though.

src/shared-components/button/deprecatedPropsHanlder.ts
```diff
@@ -0,0 +1,15 @@
+export const deprecatedProperties = {
```
michaeljaltamirano, line 1: Typo in filename: `deprecatedPropsHanlder` --> `deprecatedPropsHandler`.

src/shared-components/button/components/linkButton/index.tsx
daigof: I am OK with this change, I knew about `React.ElementType` would solve the issue but wouldnt allow the strings. But we are not going to allow any other html tag. Although I think we will only ever use this for `a` and `Link` comp.
Note that I will have to update the docs accordingly

src/shared-components/button/index.tsx
```diff
@@ -10,66 +10,76 @@ import LinkButton from './components/linkButton';
 import RoundButton from './components/roundButton';
 import TextButton from './components/textButton';
 import { COLORS, COLORS_PROP_TYPES } from '../../constants';
+import {
+  deprecatedProperties,
+  isLoadingPropFunction,
+} from './deprecatedPropsHanlder';
```
michaeljaltamirano, line 16: Will want to rename here too.

src/shared-components/button/components/linkButton/index.tsx
michaeljaltamirano: Yep--that sounds right. Looks like the `as={clickable ? 'a' : 'span'}` I saw was for some other component. Sorry about that!

docs/button/linkButton.md
```diff
@@ -73,15 +73,15 @@ import { COLORS } from 'radiance-ui/lib/constants';
...
-| disabled    | bool            | no       | false    | when disabled, click listener will not be called and the UI will look disabled                                               |
-| onClick     | func            | no       | () => {} | callback function called on click of the button                                                                              |
-| textColor   | string          | no       | ''       | color (as a string) that will override existing text, icon, and loading colors for the button (except when disabled is true) |
+| prop        | propType             | required | default  | description                                                                                                                  |
+| ----------- | -------------------- | -------- | -------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| as          | "a", "span", element | no       | "a"      | Specifies the tag or element to be rendered.                                                                                 |
```
michaeljaltamirano, line 78: Sorry, I think I was wrong about us using `span`.

src/shared-components/button/components/textButton/test.tsx
```diff
@@ -0,0 +1,50 @@
...
+describe('<TextButton />', () => {
+  describe('UI snapshots', () => {
+    it('renders without any props', () => {
+      const tree = renderer
+        .create(<TextButton>Button Text</TextButton>)
+        .toJSON();
+
+      expect(tree).toMatchSnapshot();
+    });
+    it('renders witht diabled props', () => {
```
michaeljaltamirano, line 16: `renders with the disabled prop`?

src/shared-components/button/components/linkButton/index.tsx
daigof: I changed it to `string | React.ElementType;` (I thought I tried this before and didn't work without the tsignore)

michaeljaltamirano: Any reason we shouldn't limit the specific string type to `'a'`? I'd prefer errors be raised dev side immediately, from typing, than pass something in that we didn't really build the component for.

daigof: By the component doc I thought it accepted any html element but we don't have any other case than "a" so we'll do `"a" | React.ElementType;` for now since is the stricter option

michaeljaltamirano: Sorry if I wasn't clear about this! I'm mostly suggesting  we should take advantage of adding TypeScript types to add specificity to our usage. 🙂

PR #256, opened 2020-07-02
Update dropdown component max-height

heyamykate:
I noticed a bug where the dropdown max-height was rendering outside of its container, causing the last item to not be visible. 

Based on feedback, I added a `maxHeight` prop to the dropdown component to make it more flexible. 👍 

Before:
Screen Shot 2020-07-01 at 5 44 37 PM

Screen Shot 2020-07-02 at 10 07 03 AM

After:
Screen Shot 2020-07-02 at 9 42 58 AM

### DIFF
--- a/src/shared-components/dropdown/desktopDropdown.js
+++ b/src/shared-components/dropdown/desktopDropdown.js
@@ -21,10 +21,22 @@ const DesktopDropdown = ({
   onOptionClick,
   onSelectClick,
   isOpen,
+  maxHeight,
 }) => (
-  <OffClickWrapper onOffClick={closeDropdown} css={css`width: 100%;`}>
+  <OffClickWrapper
+    onOffClick={closeDropdown}
+    css={css`
+      width: 100%;
+    `}
+  >
     <DropdownContainer textAlign={textAlign}>
-      <div id="select-input-box" onClick={onSelectClick}>
+      <div
+        id="select-input-box"
+        role="button"
+        onClick={onSelectClick}
+        onKeyDown={onSelectClick}
+        tabIndex="0"
+      >
         <div css={dropdownInputStyle({ textAlign })}>
           {currentOption && currentOption.label}
         </div>
@@ -33,9 +45,9 @@ const DesktopDropdown = ({
         </IconContainer>
       </div>
 
-      <DropdownOptionsContainer isOpen={isOpen}>
+      <DropdownOptionsContainer isOpen={isOpen} maxHeight={maxHeight}>
         {options.map(option => {
-          const { value: optionValue, disabled, label, ...rest } = option;
+          const { value: optionValue, disabled, ...rest } = option;
 
           return (
             <DropdownOption
@@ -44,11 +56,12 @@ const DesktopDropdown = ({
               selected={value === optionValue}
               disabled={disabled}
               onClick={onOptionClick}
+              // eslint-disable-next-line react/jsx-props-no-spreading
               {...rest}
             >
               {option.label}
             </DropdownOption>
-          )
+          );
         })}
       </DropdownOptionsContainer>
     </DropdownContainer>
@@ -60,10 +73,12 @@ DesktopDropdown.defaultProps = {
   options: [{ value: null, label: '' }],
   currentOption: { value: null, label: '' },
   textAlign: 'left',
-  closeDropdown() {},
-  onOptionClick() {},
-  onSelectClick() {},
+  /* eslint-disable  @typescript-eslint/no-empty-function */
+  closeDropdown: () => {},
+  onOptionClick: () => {},
+  onSelectClick: () => {},
   isOpen: false,
+  maxHeight: null,
 };
 
 DesktopDropdown.propTypes = {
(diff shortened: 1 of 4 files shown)


src/shared-components/dropdown/style.js
```diff
@@ -9,7 +9,7 @@ import {
   TYPOGRAPHY_CONSTANTS,
 } from '../../constants';
 
-const optionsContainerMaxHeight = '250px';
+const optionsContainerMaxHeight = '206px';
```
michaeljaltamirano, line 12: I think we should be cautious changing the behavior of other Dropdown usages:

250px max-height:

Screen Shot 2020-07-02 at 1 09 08 PM

208px max-height:

Screen Shot 2020-07-02 at 1 09 15 PM

So I think we should try and scope the change to fixing the particular case we're trying to fix. I don't know if any of the changes might require design resources. Some ideas (list is not: in any particular order, prescriptive, nor exhaustive):

1. Making max-height adjustable via props to the `radiance-ui` component.
1. Scoping some `emotion` css usage in PocketDerm to target the particular `ul` instance we're trying to fix. 
1. Just making the height of the parent container a bit larger to accommodate the 250px max-height. 
1. Allowing the particular Dropdown instance we're concerned with to be positioned `relative` instead of `absolute` so the container height expands and shrinks as the Select expands and shrinks.

heyamykate: Ah nice. I like the idea of making `max-height` adjustable via props. I will refactor 👍

src/shared-components/dropdown/index.js
```diff
@@ -7,7 +7,10 @@ import allowNullPropType from '../../utils/allowNullPropType';
 
 const defaultProps = {
   textAlign: 'left',
+  /* eslint-disable  @typescript-eslint/no-empty-function */
   onChange() {},
```
michaeljaltamirano, line 11: To avoid this (on the other touched files too) you can do:

```suggestion
  onChange: () => undefined,
```

src/shared-components/dropdown/style.js
```diff
@@ -101,10 +99,10 @@ export const DropdownOptionsContainer = styled.ul`
   list-style: none;
   -webkit-overflow-scrolling: touch;
 
-  ${({ isOpen }) =>
+  ${({ isOpen, maxHeight }) =>
```
michaeljaltamirano, line 102: It's verbose, but if you're not opposed I think we might prefer `optionsContainerMaxHeight` instead of `maxHeight`. It's wordy but it's also clearer that it applies to the options container specifically, and makes discoverability a bit easier.

src/shared-components/dropdown/desktopDropdown.js
```diff
@@ -88,6 +103,7 @@ DesktopDropdown.propTypes = {
   onOptionClick: PropTypes.func,
   onSelectClick: PropTypes.func,
   isOpen: PropTypes.bool,
+  maxHeight: PropTypes.string,
```
michaeljaltamirano, line 106: It looks like this component is only consumed by `Dropdown`. Since we're providing `defaultProps` in that component, we should be able to make the height prop `isRequired` here.

src/shared-components/dropdown/desktopDropdown.js
```diff
@@ -44,11 +56,12 @@ const DesktopDropdown = ({
               selected={value === optionValue}
               disabled={disabled}
               onClick={onOptionClick}
+              // eslint-disable-next-line react/jsx-props-no-spreading
               {...rest}
```
michaeljaltamirano, line 60: `...rest` will now pass in `label` since we're no longer destructuring it above. Is that intentional? 

It looks like it was an oversight that the destructured `label` from option wasn't used below in place of `{option.label}`. I think we might more likely avoid regression by continuing to destructure the `label` in the map callback and instead replace `{optional.label}` with just `{label}`.

heyamykate: good call!

src/shared-components/dropdown/style.js
heyamykate: I like that idea. I'm always in favor of verbosity if it increases clarity 👍

src/shared-components/dropdown/index.js
```diff
@@ -7,7 +7,10 @@ import allowNullPropType from '../../utils/allowNullPropType';
...
+  maxHeight: '250px',
+  value: null,
```
michaeljaltamirano, line 13: [Nit] Did you see any issues on your end with this as `undefined` instead of `null`? 

In all likelihood it probably does not matter, but sometimes 3rd party libraries (especially in Type definitions) draw a meaningful distinction, so it might be preferable to keep this as `undefined` (which is what current behavior is, if not provided) instead of null.

src/shared-components/dropdown/desktopDropdown.js
```diff
@@ -21,10 +21,22 @@ const DesktopDropdown = ({
...
+    onOffClick={closeDropdown}
+    css={css`
+      width: 100%;
+    `}
+  >
     <DropdownContainer textAlign={textAlign}>
-      <div id="select-input-box" onClick={onSelectClick}>
+      <div
+        id="select-input-box"
+        role="button"
```
michaeljaltamirano, line 35: ❤️

PR #261, opened 2020-07-10
[a11y] Update dropdown component for keyboard usage

ipc103:
[Asana Task](https://app.asana.com/0/1182623358122135/1183910548728233)

Our `Dropdown` component renders two different components depending on the type of device. For devices with touchscreen enabled, it uses the `MobileDropdown`, which takes advantage of semantic `select` and `option` elements. However, for styling purposes, the `DesktopDropdown` is using a `ul` with behavior modifications to effectively function as a select box. Doing this causes a few different accessibility issues:

  * Keyboard controls do not work as expected
  * Screen reader output does not match what the component is actually doing

In order to preserve the existing visuals, I attempted to fix these issues by adding accessibility information and additional keyboard event handlers to the component. Definitely curious to get other opinions - it may be easier to rebuild this component using semantic HTML, where we'll get most of the correct behavior for free, and then re-style to match instead of what I've done here. Included in this PR:

  * Converts all files in the `Dropdown` directory to TypeScript
  * Adds a role of `listbox` to the `ul` representing the dropdown, and `option` to each `li`. 
  * Each individual `li` is tab-able and select-able.
  * A keyboard event handler in the `OffClickWrapper` to fire the callback when the `Escape` key is pressed. This component is also used by our `ImmersiveModal`, which I see as a benefit (pressing the Escape key should allow for the same functionality as doing an off-click with the mouse)
  * Adds a `useResetFocus` hook to send focus back to the trigger button after an option is selected. 
  * Updates the focus styles to match the hover styles for each option.

Future TODO/Other Potential Improvements:

1. Once a `listbox` has been opened, you should be able to scroll the elements using the arrow keys instead of the `tab` key. Testing in Chrome and Safari with Voiceover turned off, this did not work as implemented - the options are only select-able using the `tab` key. 
2. Testing with Voiceover and Safari, the listbox does scroll correctly using the arrow keys. However, the screen reader just reads aloud the word "text" instead of the actual content of the list item.


src/shared-components/offClickWrapper/index.js
```diff
@@ -23,10 +27,19 @@ class OffClickWrapper extends React.Component {
 
   addOffClickListener = () => {
     document.addEventListener('click', this.handleOffClick, false);
+    document.addEventListener('keydown', this.handleKeyPress, false);
   };
 
   removeOffClickListener = () => {
     document.removeEventListener('click', this.handleOffClick, false);
+    document.removeEventListener('keydown', this.handleKeyPress, false);
```
ipc103, line 35: This is used in a couple of places, including in the `ImmersiveModal` to handle outside clicks. I can't think of a reason not to also respond to `Esc` for those, but I may need to also do a quick search in the PocketDerm repo to see if we are duplicating this logic there.

src/shared-components/dropdown/test.tsx
```diff
@@ -58,30 +59,30 @@ describe('<MobileDropdown />', () => {
...
-        .includes('Test1')
+        .includes('Test1'),
     ).toEqual(true);
   });
 
   describe('onSelectClick callback', () => {
     it('should be invoked onClick', () => {
       const spy = jest.fn();
-      const wrapper = shallow(
+      const wrapper = mount(
```
ipc103, line 80: These tests were failing when using `shallow` - I believe the issue is that the `hook` was still firing, but it seemed like it was coming from the render of the `class` component, which violates the laws of Hooks. Using `mount` does a more full render of the tree and seemed to fix the issue.

src/shared-components/dropdown/index.tsx
```diff
@@ -5,6 +5,20 @@ import MobileDropdown from './mobileDropdown';
 import DesktopDropdown from './desktopDropdown';
 import allowNullPropType from '../../utils/allowNullPropType';
 
+export type OptionType = {
```
ipc103, line 8: This file is just TS conversion changes.

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+import OffClickWrapper from '../offClickWrapper';
+import ChevronIcon from '../../svgs/icons/chevron-icon.svg';
+import {
+  DropdownContainer,
+  dropdownInputStyle,
+  IconContainer,
+  DropdownOptionsContainer,
+  DropdownOption,
+} from './style';
+
```
ipc103, line 15: File CHANGELOG:

1. Convert to TS
2. Add `useResetFocus` hook to move focus back to the button when an option is selected
3. Update roles, tabIndex, and aria-labels for the `DropdownOptionsContainer` ul and `DropdownOption` li. 
4. Add a keydown handler to allow selecting an option using the `enter` key.

ipc103:
I mentioned this to @michaeljaltamirano in Slack, but I'll be away from the office until 7/20. Feel free to pick up where this leaves off, reject this, or ignore it entirely and leave for me when I come back 😄

src/shared-components/dropdown/index.tsx
```diff
@@ -37,10 +55,10 @@ class Dropdown extends React.Component {
     this.setState({ isOpen: !isOpen });
   };
 
-  onSelectChange = event => {
+  onSelectChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
     const { onChange } = this.props;
-
-    const { value, selectedOptions } = event.target;
+    const {target} = event;
```
michaeljaltamirano, line 60: Did this not auto-format to be `{ target }`? I've noticed this before in this repo... wonder if something is up

ipc103: It did not - and apparently I've gotten very dependent on our linting to fix my bad whitespace habits 😅

src/shared-components/dropdown/test.tsx
```diff
@@ -16,9 +16,10 @@ const options = [
 describe('<Dropdown />', () => {
   describe('on touch screen', () => {
     it('renders <MobileDropdown />', () => {
-      global.document.documentElement.ontouchstart = () => {};
+      // @ts-ignore TODO: Add global interface to TS definitions https://stackoverflow.com/questions/40743131/how-to-prevent-property-does-not-exist-on-type-global-with-jsdom-and-t
+      global.document.documentElement.ontouchstart = () => undefined;
```
michaeljaltamirano, line 20: It looks like we can use `window` instead of `global` and the test suite still passes, without the need for `// @ts-ignore` directives. (I couldn't tell you why this is, though. Maybe enzyme v3 changed `window` behavior.)

michaeljaltamirano:
> Definitely curious to get other opinions - it may be easier to rebuild this component using semantic HTML, where we'll get most of the correct behavior for free, and then re-style to match instead of what I've done here.

We could also build off of `reach-ui`'s [`Listbox` component](https://reach.tech/listbox) or [`Menu Button` component](https://reach.tech/menu-button/)

src/shared-components/dropdown/mobileDropdown.tsx
```diff
@@ -4,9 +4,21 @@ import PropTypes from 'prop-types';
 import ChevronIcon from '../../svgs/icons/chevron-icon.svg';
 import { DropdownContainer, dropdownInputStyle, IconContainer } from './style';
 
+import { OptionType } from './index';
+
+type MobileDropdownProps = {
```
michaeljaltamirano, line 9: It's not part of the diff, so I can't comment directly on it, but [L77 (old diff)/L95 (new diff)](https://github.com/curology/radiance-ui/pull/261/files#diff-d73843a82a787f2c9d2271ef4eea7cd7R95) is spreading `{...this.props}` into this component. So we're also passing this component two unused props: `onChange` and `optionsContainerMaxHeight`.

My recommendation would be to pass down `options`, `textAlign`, and `value` explicitly, outside of the spread, on the aforementioned L77/L95.

src/shared-components/dropdown/mobileDropdown.tsx
```diff
@@ -4,9 +4,21 @@ import PropTypes from 'prop-types';
...
+  value: string | null;
```
michaeljaltamirano, line 10: Should we loosen this type to respect the intent of `value: PropTypes.any`? I think `string | number | null` would cover all bases, and we could perhaps change `value: PropTypes.any` to `value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])`.

src/shared-components/dropdown/mobileDropdown.tsx
```diff
@@ -4,9 +4,21 @@ import PropTypes from 'prop-types';
...
+  options: OptionType[];
+  textAlign: 'left' | 'center';
+  onSelectChange: (event: any) => void;
```
michaeljaltamirano, line 13:
```suggestion
  onSelectChange: (event: React.ChangeEvent<HTMLSelectElement>) => void;
```

I think this is only ever being used with `onSelectChange` as defined in the main Dropdown component, and it would be ideal, I think, to raise errors should we ever modify what's going on in the parent component.

src/shared-components/dropdown/index.tsx
michaeljaltamirano: @ipc103 I think we'll also want to convert this to a Function component. We're typing `optionsContainerMaxHeight` as optional in children components, but it's provided via `defaultProps`. If we convert this to a Function component we can rely on ES6 defaults to inform TypeScript that it is optional but, downstream, being required is not type inconsistent.

src/shared-components/dropdown/index.tsx
```diff
@@ -5,6 +5,20 @@ import MobileDropdown from './mobileDropdown';
...
+  label: string;
+  disabled?: boolean;
+};
+
+type DropdownProps = {
+  value?: string;
+  onChange: (option: OptionType) => void;
+  options: OptionType[];
+  optionsContainerMaxHeight?: string;
+  textAlign: 'left' | 'center';
```
michaeljaltamirano, line 19: Based on defaultProps it looks like `textAlign` should be potentially optional.

Also, `onChange` probably should've never been prop-typed as optional. All our usage requires it and this component makes no sense without it, so we can remove the `defaultProps` for it and make the PropTypes `isRequired`.

src/shared-components/dropdown/index.tsx
```diff
@@ -50,15 +68,15 @@ class Dropdown extends React.Component {
     this.closeDropdown();
   };
 
-  onOptionClick = event => {
+  onOptionClick = (event: React.MouseEvent<HTMLLIElement>) => {
     const { onChange } = this.props;
-
-    if (event.target.hasAttribute('disabled')) {
+    const target = event.target as HTMLSelectElement;
```
michaeljaltamirano, line 73: Since we're applying `onOptionClick` to each `option` we map over in the `DesktopDropdown`, we should be able to swap `event.target` with `event.currentTarget` (they should always be the same element) so that we can remove the casting.

Screen Shot 2020-07-20 at 9 38 47 AM

src/shared-components/dropdown/style.ts
```diff
@@ -76,7 +82,10 @@ export const IconContainer = styled.div`
   }
 `;
 
-export const DropdownOptionsContainer = styled.ul`
+export const DropdownOptionsContainer = styled.ul<{
+  isOpen: boolean;
+  optionsContainerMaxHeight?: string;
```
michaeljaltamirano, line 87: This is one of the downstream optional types that is actually always provided (currently, via `defaultProps`).

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+      onOffClick={closeDropdown}
+      css={css`
+        width: 100%;
+      `}
+    >
+      <DropdownContainer textAlign={textAlign}>
+        <div
+          id="select-input-box"
+          onClick={onSelectClick}
+          onKeyDown={(event: React.KeyboardEvent) => {
```
michaeljaltamirano, line 54: Moving this into a separate function might make it more readable for other folks (esp. if paired with a code comment?)

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+          tabIndex={isOpen ? 0 : -1}
+          aria-labelledby="select-input-box"
+        >
+          {options.map(option => {
+            const {
+              value: optionValue, disabled, label, ...rest 
+            } = option;
+
+            return (
+              <DropdownOption
```
michaeljaltamirano, line 87: Did you play around at all with trying to achieve the (kind of) opposite of `useResetFocus`, but rather than resetting focus back to another element, we actually set the focus *ahead* to the first descendent? (Kind of like the menu button when using keyboard navigation here: https://reach.tech/menu-button/#reach-skip-nav). 

I also have not dug into the internals of all this as much as you have, so I'm not use if that kind of functionality is consistent with "best practices" or we'd be potentially introducing hacky behavior that we ought to achieve with other code.

src/utils/accessibility/useResetFocus.ts
```diff
@@ -0,0 +1,18 @@
+import { useRef } from 'react';
+
+const useResetFocus = <T extends HTMLElement>() => {
```
michaeljaltamirano, line 3: For other folks this might benefit from a JSDoc comment explaining usage. 🙂

src/shared-components/dropdown/desktopDropdown.tsx
ipc103: This is a good idea - I'm going to write up a separate ticket for this and will investigate further. Looking at the `reach-ui` implementation, it looks like they generally send focus to the first list element, but it may make sense to focus on the selected element if there is one.

PR #262, opened 2020-07-10
[a11y] Add Button functionality to OptionButton

michaeljaltamirano:
### What and Why

`OptionButton`'s outer-most clickable container is a div rather than a button. This PR updates it to use button semantics instead, and converts all related code for functionality to TypeScript to better safeguard against regression.

### CHANGELOG

1. Updates all `() => {}` instances to be `() => undefined` to future-proof against `eslint` issues for https://eslint.org/docs/rules/no-empty-function.
    - It also prettifies these `.md` files, so you can just mark them viewed to reduce the diff noise.
1. Renames `type` prop, used for styling, to be `buttonType` in `OptionButton`.
    - More verbose naming makes clear the type is for our `primary` and `secondary` button styles.
    - This was required to avoid `type` namespace issues for the HTML button property. 

### TODO

- I'd like to double-check how this looks in `PocketDerm`, but also wanted to get this in front of you for a quick review (time permitting) before your EOD, if only to sanity-check direction next week.


src/shared-components/optionButton/__snapshots__/test.tsx.snap
```diff
@@ -214,19 +212,17 @@ exports[`<OptionButton /> UI snapshots radio unselected, with icon node prop 1`]
...
-  type="clickable"
+  type="button"
 >
   <div
     className="emotion-6 emotion-7"
   >
     <div
       className="emotion-0 emotion-1"
       selected={false}
-      type="secondary"
```
michaeljaltamirano, line 229: This isn't a regression: we currently pass `type` down as a DOM property even though we're only using it for styling.

src/shared-components/optionButton/style.ts
```diff
@@ -2,28 +2,43 @@ import styled from '@emotion/styled';
...
+const getTypeColor = (buttonType: BaseIconWrapperStylesProps['buttonType']) => {
+  if (buttonType === 'secondary') {
     return COLORS.secondary;
   }
+
   return COLORS.primary;
 };
 
-export const ClickableContainer = styled(Container)`
+export const ClickableContainer = styled.button<{
```
michaeljaltamirano, line 38: I'm pretty sure my preference is for us to use `styled.button` directly, for explicitness. However, There is a version of this PR that could rely on [`withComponent`](https://emotion.sh/docs/styled#change-the-rendered-tag-using-withcomponent) instead. The simplicity could be arguably more clear. 

If we prefer to rely on more emotion API to DRY things up on our end, I can explore that in more detail next week.

src/shared-components/optionButton/index.tsx
```diff
@@ -45,19 +66,16 @@ const OptionButton = ({
   </ClickableContainer>
 );
 
+/* eslint-disable react/require-default-props */
```
michaeljaltamirano, line 69: One of the motivations for our more strictly enforcing this rule was to have full documentation of how components were being used. 

As we move to more Function components, which can use ES6 defaults, and especially as we migrate more code over to TypeScript, the usefulness of this rule drops to 0. 

For now, I think it's fine to just in-line disables. I think I will bring up disabling this rule in Demos and Discussions soon.

src/shared-components/container/style.ts
```diff
@@ -0,0 +1,99 @@
...
+  overflow: hidden;
+  object-fit: cover;
+`;
+
+type StyledContainer = StyledComponent<
+  DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>,
+  {
+    type?: 'clickable' | 'message' | undefined;
+  },
+  any
```
ipc103, line 75: [silly question] As a TS novice, I still find some of these type definitions to be difficult to parse. For example, what does the `any` refer to? Is that part of the `DetailedHTMLProps` that we expect? My best attempt at explaining this: 

We're defining a type called `StyledContainer`, which is a `StyledComponent` which accepts `DetailedHTMLProps` as an argument, plus the additional attribute of `type` optionally.

src/shared-components/optionButton/__snapshots__/test.tsx.snap
```diff
@@ -91,19 +91,17 @@ exports[`<OptionButton /> UI snapshots checkbox selected, without custom icon 1`
   line-height: 1.5;
 }
 
-<div
+<button
```
ipc103, line 94: 😍  just seeing this snapshot makes me so happy

src/shared-components/optionButton/index.tsx
```diff
@@ -12,27 +12,48 @@ import {
 } from './style';
 import CheckmarkIcon from '../../svgs/icons/checkmark-icon.svg';
 
+export type OptionButtonProps = {
+  buttonType?: 'primary' | 'secondary';
```
ipc103, line 16: I like the updated naming - this is pretty consistent with other components as well, I believe.

src/shared-components/optionButton/index.tsx
ipc103: I don't have a strong opinion, but to be honest, I wouldn't hate if it was possible to update the rule so that you either use an ES6 default, required prop, or a default prop. This check has been helpful for me to make sure a component has a consistent contract for what it requires. And if you don't need a prop to be required or have any default value, adding a defaultProp of `undefined` isn't too hard, either.

src/shared-components/optionButton/style.ts
```diff
@@ -2,28 +2,43 @@ import styled from '@emotion/styled';
...
-import Container from '../container';
+import {
+  COLORS, ANIMATION, SPACER, BOX_SHADOWS, 
+} from '../../constants';
+import { containerStyles, ContainerType } from '../container/style';
 
-const getOptionTypeStyles = optionType => {
+import { OptionButtonProps } from '.';
+
+type BaseIconWrapperStylesProps = Required<
```
ipc103, line 12: This is just saying that `buttonType`, `optionType`, and `selected` from the `OptionButtonProps` are all required for the `BaseIconWrapperStylesProps`, yes?

src/shared-components/optionButton/style.ts
```diff
@@ -2,28 +2,43 @@ import styled from '@emotion/styled';
...
   }
+
   return css`
     border-radius: 50%;
   `;
 };
 
-const getTypeColor = type => {
-  if (type === 'secondary') {
+const getTypeColor = (buttonType: BaseIconWrapperStylesProps['buttonType']) => {
```
ipc103, line 30: This is cool - I didn't realize you could access properties off of a `type` like this.

src/shared-components/optionButton/style.ts
ipc103: I think I like using the `button` as well. My big takeaway from the past two weeks is that elements should be defined by what they do, not what they look like. Using the `button` here and then applying the styles keeps these concerns well separated.

src/shared-components/optionButton/index.tsx
```diff
@@ -12,27 +12,48 @@ import {
...
   subtext,
-  type,
   ...rest
-}) => (
-  <ClickableContainer onClick={onClick} type="clickable" tabIndex="0" {...rest}>
+}: OptionButtonProps) => (
+  <ClickableContainer
+    onClick={onClick}
+    type="button"
+    containerType="clickable"
```
ipc103, line 39: Unless I'm missing something, I think we also might want to pass `role={buttonType}`. We'll get the correct behavior from using a `button` either way, but I think these always either act as radio or checkbox toggles when they are clickable.

src/shared-components/container/style.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const containerStyles = (type?: ContainerType) => css`
+  background-color: ${COLORS.white};
+  border: 1px solid ${COLORS.border};
+
+  ${getContainerTypeStyles(type)};
+`;
+
+// TODO update desktop padding to 48px
```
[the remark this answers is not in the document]
michaeljaltamirano: It's a 2 year old TODO--in general I am in favor of leaving things as they are as much as possible, but given the age I think we're fine to just remove it.

src/shared-components/container/style.ts
michaeljaltamirano: I've added a code comment in 7926094 but this is just the type that `@emotion/styled` returns 😬 , so I don't think there's a way for people to get comfortable with advanced and complex types beyond working more directly/frequently with these libraries and browsing the type definitions. 

So `any` corresponds to the Theme: `StyledComponent<InnerProps, StyleProps, Theme>`

src/shared-components/container/style.ts
```diff
@@ -0,0 +1,99 @@
...
+export type ContainerType = 'clickable' | 'message';
+
+const getContainerTypeStyles = (type?: ContainerType) => {
+  switch (type) {
+    case 'message':
+      return messageStyle;
+    case 'clickable':
+      return clickableStyle;
+    default:
+      return css``;
```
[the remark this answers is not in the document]
michaeljaltamirano: It does but I'm usually loath to change the API/outcome of things. I couldn't *find* an example of this executing in PocketDerm, but I'd rather not introduce an unrelated regression to the task at hand, so I've added a TODO instead.

src/shared-components/optionButton/style.ts
michaeljaltamirano: Somewhat 😬 

`OptionButtonProps` has `buttonType` and `selected` as potentially optional, so `Required` (a TS built-in Utility: https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredt) casts it as non-optional. 🙂

So the type in this diff is equivalent to:

```
type BaseIconWrapperStylesProps = {
  buttonType: 'primary' | 'secondary';
  optionType: 'radio' | 'checkbox';
  selected: boolean;
}
```

But by using the Utilities with the imported type we can couple the style TypeScript types with the component TypeScript types. It's not necessary--if it were just one prop I wouldn't have done it--but by sharing types this explicitly I think it's easier to visualize what is being shared and spread elsewhere.

src/shared-components/optionButton/style.ts
michaeljaltamirano: Yep--it's how you index into it. 

You probably won't ever need to do it but you could also do it to pull the type of an item in an array, by keying in with `[number]`:

Screen Shot 2020-07-15 at 5 42 29 PM

src/shared-components/optionButton/index.tsx
michaeljaltamirano: I would be open to ignoring this rule for Function components in this repo: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md#ignorefunctionalcomponents

It would just then be incumbent on us to provide sane defaults, which would be easy enough with TypeScript. But the `eslint` rule isn't refined enough to call out about that.

src/shared-components/checkbox/index.js
```diff
@@ -37,6 +39,7 @@ const Checkbox = ({
     type={type}
     icon={icon}
     size={size}
+    // eslint-disable-next-line react/jsx-props-no-spreading
```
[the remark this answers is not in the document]
michaeljaltamirano: I think our philosophy is, for the most part, being annoyed by `eslint-disabling` this rule, but I am very pro it 😅 

That being said, it's probably not all that appropriate in a component library where we're potentially spreading HTML attributes to the DOM. While I think the rule keeps top of mind that we should prefer explicitness, I'm open to removing the rule in this repo.

michaeljaltamirano:
> The functionality is fantastic, and overall I love the approach of using an actual interactive element. There might be a world where we would want this to use either a `radio` or `checkbox` input instead of a `button`, but honestly I think this will meet all of our compliance and usability goals if we add a `role` to indicate the type of input this is representing.
> 
> The button width is behaving strangely on the Storybook for some reason - I imagine that the `button` is inheriting a `width` from somewhere? Adding `width` and `max-width` seemed to fix it...
> 
> Screen Shot 2020-07-10 at 5 18 37 PM
> 
> Once those issues are resolved, though, I think this is good to go. I'm going to go ahead an approve, but obviously feel free to get a second set of eyes before merging if you'd like some additional thoughts for any reason!

It looks like there is div-specific styling in our stories (along with some button-specific changes that needed to be made). This is fixed in d615b7a.

src/shared-components/optionButton/index.tsx
michaeljaltamirano: Updated with `role={optionType}` (`buttonType` is `primary | secondary`, `optionType` is `checkbox | radio`).

src/shared-components/optionButton/style.ts
```diff
@@ -2,30 +2,47 @@ import styled from '@emotion/styled';
...
 
-export const ClickableContainer = styled(Container)`
+export const ClickableContainer = styled.button<{
+  containerType: ContainerType;
+}>`
+  ${({ containerType }) => containerStyles(containerType)};
   padding: ${SPACER.large};
   margin-bottom: ${SPACER.medium};
+  width: 100%;
+  text-align: left;
```
michaeljaltamirano, line 45: `width` and `text-align` are new changes to account for button differences from existing div behavior.

src/shared-components/optionButton/style.ts
michaeljaltamirano: I ended up writing the simpler type because while playing with this in the main app (under `v9.3-beta.0`) I realized that some of these styled components are being used directly, outside of the `OptionButton`. (Thankful for TypeScript for catching this.)

PR #265, opened 2020-07-16
[a11y] Add box-shadow outline to button focus state

michaeljaltamirano:
### What & Why

Keyboard navigation without focus state indication is difficult--adding this style will make things easier. This will apply to *all* buttons. 

boxshadow

I view this as a temporary solution--the outline on our primary buttons is very faint and not ideal, but it's a simple start.


benkolde:
@michaeljaltamirano looks good, we're going to be revisiting the the styling of this in the follow-up pitch

PR #266, opened 2020-07-16
[a11y] Add aria-disabled to disabled usage

michaeljaltamirano:
### What and Why

Our accessibility tooling requires explicit directives to ignore contrast/etc. issues on "disabled" content. The HTML `disabled` attribute is not sufficient, so this PR adds `aria-disabled` alongside our `disabled` prop logic (in the outermost element, in the case of a component with multiple `disabled` prop usages).

### DIFF
--- a/src/shared-components/button/components/roundButton/index.tsx
+++ b/src/shared-components/button/components/roundButton/index.tsx
@@ -77,6 +77,7 @@ const RoundButton = ({
     <RoundButtonWrapper>
       <RoundButtonBase
         onClick={!disabled && !isLoading ? onClick : () => false}
+        aria-disabled={disabled}
         disabled={disabled}
         buttonType={buttonType}
         buttonColor={buttonColor}
--- a/src/shared-components/button/components/textButton/index.tsx
+++ b/src/shared-components/button/components/textButton/index.tsx
@@ -29,6 +29,7 @@ const TextButton = ({
   ...rest
 }: TextButtonProps) => (
   <BaseTextButton
+    aria-disabled={disabled}
     disabled={disabled}
     onClick={!disabled ? onClick : event => event.preventDefault()}
     // eslint-disable-next-line react/jsx-props-no-spreading
--- a/src/shared-components/button/index.tsx
+++ b/src/shared-components/button/index.tsx
@@ -87,6 +87,7 @@ class Button extends React.Component<ButtonProps> {
 
     return (
       <ButtonBase
+        aria-disabled={disabled}
         disabled={disabled}
         onClick={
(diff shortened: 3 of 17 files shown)


src/shared-components/button/components/roundButton/index.tsx
```diff
@@ -77,6 +77,7 @@ const RoundButton = ({
     <RoundButtonWrapper>
       <RoundButtonBase
         onClick={!disabled && !isLoading ? onClick : () => false}
+        aria-disabled={disabled}
```
daigof, line 80: ok so pardon my quick and opinionated investigation but this article: https://stackoverflow.com/questions/38059140/when-to-use-the-disabled-attribute-vs-the-aria-disabled-attribute-for-html-eleme

says we don't need the attribute when it is supported natively by the html component and yes we do need it when is a div for example. Looks like `RoundButtonBase` is a button do you think we still need the attribute for buttons?

src/shared-components/button/components/textButton/index.tsx
```diff
@@ -29,6 +29,7 @@ const TextButton = ({
   ...rest
 }: TextButtonProps) => (
   <BaseTextButton
+    aria-disabled={disabled}
```
daigof, line 32: this is also an `styled.button`

src/shared-components/button/index.tsx
```diff
@@ -87,6 +87,7 @@ class Button extends React.Component<ButtonProps> {
 
     return (
       <ButtonBase
+        aria-disabled={disabled}
```
daigof, line 90: `styled.button`

src/shared-components/dropdown/mobileDropdown.js
```diff
@@ -23,7 +23,12 @@ const MobileDropdown = ({
         }
 
         return (
-          <option key={option.value} value={option.value} disabled={isDisabled}>
+          <option
```
daigof, line 26: looks like this element supports disabled natively as well
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

src/shared-components/button/components/roundButton/index.tsx
michaeljaltamirano: No pardon needed :sweat_smile: it looks like you're right about the buttons. I think I threw it on everything without much discernment, so I'm going to take a second look at this--we probably only need it for divs where we're passing `disabled` to apply styling. We want our accessibility tools to ignore all the content within that markup (which will still be in the DOM) in those cases. Thank you for the critical eye.

michaeljaltamirano:
Closing for future consideration

PR #274, opened 2020-07-21
[Typography] Add focus styling to Link styles

michaeljaltamirano:
### What and Why

1. Adds focus styling to `TYPOGRAPHY_STYLE.link`

Holding off on review because this is perhaps best paired with the [`focus-visible polyfill`](https://github.com/WICG/focus-visible)

### Other

The next `radiance-ui` release that has this in it will need to be a v10 due to removing the deprecating `TYPOGRAPHY` properties. I checked and we do not currently use it, and the TODO was related to v2, so I'm finally getting rid of it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "@svgr/webpack": "^4.0.2",
     "@types/enzyme": "^3.10.3",
     "@types/jest": "^24.0.20",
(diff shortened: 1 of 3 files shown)


michaeljaltamirano:
Closing in favor of #778

PR #275, opened 2020-07-21
[a11y] Fix dropdown accessibility information

ipc103:
The changes introduced in v9.3.0 did not correctly take advantage of `aria-activedescendant`. This property references the id of another element, but I was passing the option value. This PR adds an id to the child elements to make this work properly. 

In addition, I replaced the `listbox` role with a `menu` and `menuitemradio` combination. This seemed to make more sense, since the Dropdown only allows for one element to be selected. Testing with Voiceover and Safari, this gave more consistent behavior for reading off the list items. 

Finally, I updated the focus style of the menu to match the focus state for the rest of the form elements.

### DIFF
--- a/src/shared-components/dropdown/desktopDropdown.tsx
+++ b/src/shared-components/dropdown/desktopDropdown.tsx
@@ -11,7 +11,6 @@ import {
   IconContainer,
   DropdownOptionsContainer,
   DropdownOption,
-  applyFocusState,
 } from './style';
 
 import { OptionType } from './index';
@@ -73,7 +72,6 @@ const DesktopDropdown = ({
           aria-haspopup="menu"
           role="button"
           ref={initialFocus}
-          css={applyFocusState}
         >
           <div css={dropdownInputStyle({ textAlign })}>
             {currentOption && currentOption.label}
--- a/src/shared-components/dropdown/desktopDropdown.tsx
+++ b/src/shared-components/dropdown/desktopDropdown.tsx
@@ -11,6 +11,7 @@ import {
   IconContainer,
   DropdownOptionsContainer,
   DropdownOption,
+  applyFocusState,
 } from './style';
 
 import { OptionType } from './index';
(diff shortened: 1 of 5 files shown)


src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -84,10 +86,10 @@ const DesktopDropdown = ({
         <DropdownOptionsContainer
           isOpen={isOpen}
           optionsContainerMaxHeight={optionsContainerMaxHeight}
-          role="listbox"
+          role="menu"
           aria-activedescendant={value}
-          tabIndex={isOpen ? 0 : -1}
```
ipc103, line 89: Removed the `tabindex` from the `ul` - this felt like a weird intermediate step to focus on the entire menu list instead of simply proceeding to the element in the list.

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -84,10 +86,10 @@ const DesktopDropdown = ({
...
           aria-labelledby="select-input-box"
+          aria-hidden={!isOpen}
```
ipc103, line 92: A `listbox` or `menu` must have focusable content to be valid, so we hide this from the accessibility tree when the menu is not open.

src/shared-components/dropdown/test.tsx
```diff
@@ -104,7 +105,10 @@ describe('<DesktopDropdown />', () => {
         />,
       );
 
-      wrapper.find('#select-input-box').simulate('click');
+      wrapper
+        .find('#select-input-box')
```
ipc103, line 109: After adding the `css` prop to the selector `div`, this query selector began returning multiple nodes. Selecting the first one produced the correct behavior, though I'm not exactly sure why...

michaeljaltamirano: If you add some snapshots (for the before and after) it might be easy to see the change in the diff.

michaeljaltamirano: Additionally, since that looks like an ID, we might definitely want to check that because there should only be one node, right?

ipc103: Good call on the snapshot test - will add that in now. Yeah, I'm kind of confused as to what is happening. It seems like adding the `css` prop (regardless of the content) was causing this to return a second node...

ipc103:
Yeah, looking at this again, I think the most consistent thing might be to remove the changed focus state. The default focus state currently gets applied when you click on the dropdown, so it's not doing anything visually different from what we already do. That also takes care of the test issue (thought I'm still curious as to what that's about). 

Unless you have other thoughts, I'll go ahead and remove the `applyFocusState` style.

PR #283, opened 2020-07-22
[a11y] Trap keyboard focus in DialogModal

ipc103:
[Asana Task](https://app.asana.com/0/1182623358122135/1182893274583740)

When the `DialogModal` is open, we should lock the focus inside of the modal to make it easy for keyboard users to interact with. This PR accomplishes this by using the [FocusScope](https://react-spectrum.adobe.com/react-aria/FocusScope.html) component from react-aria. We automatically trap focus inside the modal when it opens, and return focus to the last focused element when it closes. In addition, if a close icon is present, we allow the user to dismiss the modal using the Escape key. 

Screen Shot 2020-07-22 at 2 47 57 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,6 +132,7 @@
   "dependencies": {
     "@emotion/core": "10.0.10",
     "@emotion/styled": "10.0.10",
+    "@react-aria/focus": "^3.0.2",
     "lodash.round": "^4.0.4",
     "lodash.throttle": "^4.1.1",
     "lodash.uniqueid": "^4.0.1",
--- a/src/shared-components/dialogModal/index.tsx
+++ b/src/shared-components/dialogModal/index.tsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { Transition } from 'react-transition-group';
+import { FocusScope } from '@react-aria/focus';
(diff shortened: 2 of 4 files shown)


michaeljaltamirano:
> In addition, if a close icon is present, we allow the user to dismiss the modal using the Escape key.

Is this informed by any sort of WCAG or WAI-ARIA spec, or just our own flavor?

ipc103:
@michaeljaltamirano - just our own flavor. I haven't seen a spec that explicitly requires that the Escape key dismisses modals, and I can imagine cases where we might require an explicit confirmation before returning to the rest of the page. If that seems too restrictive, definitely happy to apply the escape key behavior for all instances.

michaeljaltamirano:
> @michaeljaltamirano - just our own flavor. I haven't seen a spec that explicitly requires that the Escape key dismisses modals, and I can imagine cases where we might require an explicit confirmation before returning to the rest of the page. If that seems too restrictive, definitely happy to apply the escape key behavior for all instances.

I'm not too familiar with the DialogModal development but I would venture to guess that the more restrictive version requires an explicit yes/no from a patient, so I think it's fine as is. Great use of the tool.

ipc103:
Super interesting! The component felt so intuitive to use and solved such an important problem that I almost cried 😂

PR #284, opened 2020-07-22
[a11y] Add box-shadow to TextButton

michaeljaltamirano:
I've updated most of our button stylings, but saw that our TextButton component has it's own stand-alone focus outline override. This PR adds the box-shadow we've been adding across components for keyboard navigation focus management. 

Screen Shot 2020-07-22 at 3 26 54 PM

Andrew and Ben tagged as a general FYI

### DIFF
--- a/src/shared-components/button/components/textButton/style.ts
+++ b/src/shared-components/button/components/textButton/style.ts
@@ -1,8 +1,7 @@
 import styled from '@emotion/styled';
 
-import { COLORS } from '../../../../constants';
+import { BOX_SHADOWS, COLORS } from '../../../../constants';
 
-/* eslint-disable-next-line import/prefer-default-export */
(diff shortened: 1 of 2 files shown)


src/shared-components/button/components/textButton/style.ts
```diff
@@ -19,5 +18,6 @@ export const BaseTextButton = styled.button<{ disabled: boolean }>`
   &:active,
   &:focus {
     outline: none;
+    box-shadow: ${BOX_SHADOWS.focusSecondary};
```
ipc103, line 21: Just a note that I'm going to add a note to our pitch to look at potentially defining `FOCUS.secondary` - might be helpful to remove some of the duplication and make things a bit more expressive. Definitely not needed for now though.

PR #285, opened 2020-07-22
[a11y] Add keyboard control to Accordion component

ipc103:
[Asana Task](https://app.asana.com/0/1182623358122135/1182893274583743)

In this PR:

  * Adds a keyDown listener so that you can expand or close the Accordion using the Enter key
  * Provides tabIndex, role, and our custom focus state to the Accordion title to make the button focusable
  * Uses `aria-expanded`, `aria-disabled`, and `aria-hidden` to provide additional info to screen readers and to ensure that disabled components will not raise color-contrast issues


ipc103:
Going to merge this in now, but will wait until tomorrow or Friday to cut another Radiance release as I think we have a few other in-flight changes.

PR #303, opened 2020-08-10
Add disabled prop to Checkbox component

AlyssaKirstine:
This PR adds a disabled prop to the Checkbox component and disables the checkbox when this prop is true by changing the background color of it and preventing the user from clicking the checkbox. This is needed more immediately for the CCPA Opt Out project.

Note: This is no longer needed for the CCPA Opt Out project (we decided to use a button instead), but I might as well still publish this in Radiance since I did the work for it.

Screen Shot 2020-08-10 at 4 14 16 PM

### DIFF
--- a/src/shared-components/selectorButton/style.js
+++ b/src/shared-components/selectorButton/style.js
@@ -51,8 +51,8 @@ const secondarySelectorStyle = checked => css`
 `;
 
 const disabledSelectorStyle = css`
-  background-color: ${COLORS.purple30};
-  border-color: ${COLORS.purple30};
+  background-color: ${COLORS.disabled};
+  border-color: ${COLORS.disabled};
   cursor: not-allowed;
 `;
 
--- a/src/shared-components/selectorButton/style.js
+++ b/src/shared-components/selectorButton/style.js
@@ -51,8 +51,8 @@ const secondarySelectorStyle = checked => css`
 `;
 
 const disabledSelectorStyle = css`
-  background-color: ${COLORS.disabled};
-  border-color: ${COLORS.disabled};
+  background-color: ${COLORS.purple30};
+  border-color: ${COLORS.purple30};
   cursor: not-allowed;
 `;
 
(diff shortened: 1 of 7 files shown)


stories/checkbox/index.js
```diff
@@ -1,10 +1,12 @@
...
 import { storiesOf } from '@storybook/react';
 import { withDocs } from 'storybook-readme';
-import { withKnobs, text, boolean, select } from '@storybook/addon-knobs';
+import {
+  withKnobs, text, boolean, select, 
+} from '@storybook/addon-knobs';
 import { action } from '@storybook/addon-actions';
 import { css } from '@emotion/core';
-
+// eslint-disable-next-line import/extensions
```
AlyssaKirstine, line 9:
```suggestion
```

src/shared-components/selectorButton/style.js
```diff
@@ -30,12 +30,14 @@ export const OuterContainer = styled.div`
...
 
 export const SelectorIcon = styled.div`
-  cursor: pointer;
   fill: currentColor;
   left: 50%;
   position: absolute;
   top: 50%;
   transform: translate(-50%, -50%);
+  ${({ disabled }) => css`
+    cursor: ${disabled ? 'not-allowed' : 'pointer'};
```
ZeMunchkin, line 39: 💯

stories/checkbox/index.js
```diff
@@ -24,6 +25,9 @@ stories.add(
       <Checkbox type="secondary" checked>
         This is a secondary checkbox
       </Checkbox>
+      <Checkbox checked disabled>
+        This is a disabled checkbox
```
ZeMunchkin, line 29: Should also add the Storybook knob below to control the disabled prop (see lines 39-51)

AlyssaKirstine: Good point. Thanks. Added [here.](https://github.com/curology/radiance-ui/pull/303/commits/3566deec38ff7f37bfbb1d454476490d45654dd7)

PR #316, opened 2020-08-20
[TS conversion] - Progress Bar and update dependencies

daigof:
- Convert ProgressBarComponent add 1 more test

- TS flagged we had a typo in ANIMATION.defaultTiming was `defaultIming` before for so long

- change `@babel/env` to `@babel/preset-env` in babel configs to be consistent with official docs

- Add  TSX snapshots exclude rule for  test Coverage calculation (long overdue)

- removed plain old `babel-core` npm library

- Update Babel and related presets (not mayor version)

- Update Jest, Typescript and babel-jest to latest version. 

### Dependencies Update Reason
I had a test error where Jest/Babel didnt like `as const` added in `PROGRESS_BAR_STATUS` const.


package.json
```diff
@@ -53,14 +53,14 @@
   "homepage": "https://github.com/PocketDerm/radiance-ui#readme",
   "devDependencies": {
     "@babel/cli": "^7.1.2",
-    "@babel/core": "^7.1.2",
+    "@babel/core": "^7.11.1",
     "@babel/plugin-proposal-class-properties": "^7.1.0",
```
michaeljaltamirano, line 57: We should bump all the babel packages in lock-step: @babel/cli, @babel/plugin-proposal-class-properties, @babel/plugin-proposal-export-default-from, @babel/plugin-proposal-export-namespace-from, @babel/plugin-transform-parameters, @babel/preset-react, @babel/register

src/.babelrc
```diff
@@ -1,6 +1,6 @@
 {
   "presets": [
-    ["@babel/env", {
+    ["@babel/preset-env", {
```
michaeljaltamirano, line 3: FWIW the babel docs use both: https://babeljs.io/docs/en/usage

src/constants/progressBarStatus/index.ts
```diff
@@ -1,5 +1,7 @@
-export default {
+const PROGRESS_BAR_STATUS = {
```
michaeljaltamirano, line 1: 🙏

src/shared-components/progressBar/style.ts
```diff
@@ -3,35 +3,42 @@ import { css, keyframes } from '@emotion/core';
...
+  height: number;
+  status: ProgressBarStatusType;
+}>`
   position: absolute;
   top: 0;
   left: 0;
   right: 0;
   background: ${props => props.backgroundColor};
   overflow: hidden;
-  transition: opacity ${ANIMATION.defaultIming} ease-in-out 500ms;
```
michaeljaltamirano, line 18: Catching the typos might be my favorite part of TypeScript 🙂

src/shared-components/progressBar/__snapshots__/test.tsx.snap
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  to {
+    -webkit-transform: translateX(-5%);
+    -ms-transform: translateX(-5%);
+    transform: translateX(-5%);
+  }
+}
+
+@keyframes animation-0 {
```
michaeljaltamirano, line 18: Do we know why this is showing two separate keyframes with the same name? Is it fixing the typo? Are we duplicating the style somehow (maybe because there was a typo, we didn't see it working, and so activated it some other way?)

src/shared-components/progressBar/index.tsx
```diff
@@ -4,18 +4,34 @@ import PropTypes from 'prop-types';
 import { COLORS, PROGRESS_BAR_STATUS } from '../../constants';
 import { OuterContainer, InnerBar } from './style';
 
+export type ProgressBarStatusType =
```
michaeljaltamirano, line 7: I think this might be easier if we pull over the `valueof` utility into this repo.

```ts
type valueof<T> = T[keyof T];
type ProgressBarStatusTypes = valueof<typeof PROGRESS_BAR_STATUS>;
```

src/shared-components/progressBar/index.tsx
```diff
@@ -4,18 +4,34 @@ import PropTypes from 'prop-types';
...
+  // ...rest type
+  [key: string]: any;
+};
+
 const ProgressBar = ({
-  backgroundColor,
-  barColor,
-  height,
-  loadingTime,
+  backgroundColor = COLORS.background,
```
michaeljaltamirano, line 23: We can just go ahead and remove the `defaultProps` and rely on ES6 defaults alone. I think that should be our preference going forward.

src/shared-components/progressBar/style.ts
```diff
@@ -3,35 +3,42 @@ import { css, keyframes } from '@emotion/core';
...
+  transition: opacity ${ANIMATION.defaultTiming} ease-in-out 500ms;
   width: 100%;
   z-index: 2;
   height: ${props => `${props.height}px`};
 
   ${({ status }) =>
-    [PROGRESS_BAR_STATUS.success, PROGRESS_BAR_STATUS.error].includes(status) &&
+    (status === PROGRESS_BAR_STATUS.success ||
```
michaeljaltamirano, line 30: Was the change from using the array a TypeScript thing, or personal preference?

src/shared-components/progressBar/test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    it('renders with default props', () => {
+      const component = renderer.create(
+        <ProgressBar status={PROGRESS_BAR_STATUS.loading} />,
+      );
+
+      const tree = component.toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('renders with custom props', () => {
```
michaeljaltamirano, line 19: 🙏

src/shared-components/progressBar/__snapshots__/test.tsx.snap
daigof: the typo influenced the `transition` css not the animation.  I analyzed and see no apparent issues. Remeber we updated Jest to a new mayor version so that could be it also the default props test looks the same (except for the defaultTiming fix) and only defines it once I think is just a snapshot creation thing

package.json
daigof: piushing this soon

src/shared-components/progressBar/index.tsx
daigof: it worked with `export type ProgressBarStatusType = keyof (typeof PROGRESS_BAR_STATUS);`
thanks I was looking for an example to reduce the lines

michaeljaltamirano: @daigof That works because the keys happen to match the values, but we should be sure to target the literal values instead of the keys. So you can just in-line the `valueof` definition: `typeof PROGRESS_BAR_STATUS[keyof typeof PROGRESS_BAR_STATUS]` should work.

src/shared-components/progressBar/index.tsx
daigof: eslint will protest that non required proptypes requires a default Proptype

src/shared-components/progressBar/style.ts
daigof: TS protested that `loading` could never be assigned to the ['success','error'] array I decided to refactor instead of ts-ignoring

src/shared-components/progressBar/index.tsx
michaeljaltamirano: @daigof We can eliminate that rule for function components by changing the `eslint` rules:

```
"react/require-default-props": [
      "error",
      { "ignoreFunctionalComponents": true }
    ],
```

package.json
daigof: pushed all updated dependecies. Bundle size matches what we have in curology now

src/shared-components/progressBar/index.tsx
daigof: Yeah I understand about key thing. 
I was honestly looking for a better way to define my type, Ill push the version you provided but it looks kind of weird sort of for readability purpose or short list of props the first way I coded look better maybe?

I want to have this at hand for future uses so I want to commit your version

src/shared-components/progressBar/index.tsx
daigof: done. nice, Thanks for this standarization, Ill start applying in Radiance

src/shared-components/progressBar/index.tsx
michaeljaltamirano: I agree it's not very readable, so we should probably add the `valueof` to be able to be used everywhere like we do in our main repo

PR #319, opened 2020-08-24
[Hotfix Storybook] - Downgrade dependency that is causing issues

daigof:
- We have to downgrade `@babel/preset-react` cause it is causing issued transpiling for storybook and messing up with emotion css prop
- took me a while to pinpoint what was causing this, had this issue before it is related to emotion `css=` prop in elements and the usage of `@emotion/babel-preset-css-prop` which was updated to latest here but still does not work
- tried the next version to 7.0.0  (7.6.3) and that didnt work either so we need to lock to 7.0.0 https://www.npmjs.com/package/@babel/preset-react

- tried some options passed to `@babel/preset-react` but didn't work in first attempts. May look into this in next cooldown or just analyze the migration to Storybookv6

I think it is best to release this fix now and start thinking on migrating to Storybook v6 which will fix all this old dependecies issues since it has more built integration and doesnt rely much on babel. See daylight which is in v6 I dont have to define a babel config file at all and TS integration is built in


babel.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   presets: [
     '@babel/preset-env',
-    '@babel/react',
+    '@babel/preset-react',
```
daigof, line 4: just standarization here. I prefer this way so when you look at package.json file and try to find where in the project a dependecy is used it appear in search result

daigof:
@michaeljaltamirano ok here is more detail why I think we should keep this fix and the latest PR which incidentally was not only related to TS convert ProgressBar
- the babel dependencies were not exactly ideal in the previous version, I think what we have now is more in sync (some were in 7.0 some in 7.1 some in 7.2)
image

- The underlying issue updating `@babel/preset-react` to latest was a css prop issue But you can see in console right now in storybook prod that it says React is not defined. You can import React in preview.js like some pages says but then you see that css prop styles are not applied (you dont even get this an error you have to digg in the inspector). Then I tried to add the `jsx` pragma to preview.js at this point I knew this is just patch that solves nothing, You start realizing that you need the file as they were and try to fix elsewhere, thats when I detected it was dependecies thing.

- The old PR needed to update babel and jest because there were some error when tryin to use this files `as const`. This syntax cannot be used with old dependecies 
```
const PROGRESS_BAR_STATUS = {
  loading: 'loading',
  success: 'success',
  error: 'error',
} as const;
```

- If we are rolling back, I would rather remove the `as const` syntax and revert all dependecies to what they were so the test can pass and we can release the version mantaining some refactor from old PR. But ultimately I dont think it matters much, also the `babel-preset-css-prop` upgrade was not necessary I just threw it in here in hope of fixing the bug because that is the preset that is not working because @babel/preset-react is doing something that the emotion preset cannot process or viceversa these 2 packages are clearly out of sync

michaeljaltamirano:
@daigof I haven't dug into the internals of this as much as you have, but it looks like if we convert `.storybook/preview.js` to `.storybook/preview.tsx`, we fix the behavior. Can you double-check this is correct?

daigof:
@michaeljaltamirano that has the same effect of adding `import React from...` you still have css prop issues in other pages like Checkbox
image

Which would be solvable by converting Checkbox story to TSX so on and so forth.

I propose 2 things: get this PR published or revert all package versions to what they were and remove `as const` syntax.
A third proposition would be convert now Checkbox story to TSX (or add the jsx pragma has the same effect) but I know even not noticeable there are css prop errors in all other pages. This has no effect to the library build.

Thanks to your finding it looks like converting all stories files to TSX would fix the issue with all latest babel packages, So taht can be a task for cooldown too.

michaeljaltamirano:
@daigof I see. Thank you for pointing out that screenshot. My preference is to revert 6b295d67c3c4a53402c5fd5f7325cfb4a16c1e84 (#316) so that the repo is back in sync with the `10.3.0` release.

daigof:
@michaeljaltamirano  Repo revert back to commit before merged PR. Closing this one and opening another ProgressBar PR that doesnt messup with dependecies

PR #336, opened 2020-09-08
Update Carousel component to have bottom right aligned arrows

AlyssaKirstine:
The [new LP pricing module](https://www.notion.so/curology/Pitch-New-Landing-Page-Pricing-Module-412f1153db264ea8a1d33c7416aa56d1) has a different positioning for the carousel arrows. They are aligned beneath the carousel cards on the bottom right side. The coloring of the arrows is also a bit different. Because this component will only be used in the pricing module on the site, rather than creating a new `arrowColor` prop, I just added both of these things (new alignment and new coloring) as a `bottomRightAlignedArrows` prop to the Radiance Carousel component.

Screen Shot 2020-09-08 at 5 51 55 PM

### DIFF
--- a/src/shared-components/carousel/arrow/style.ts
+++ b/src/shared-components/carousel/arrow/style.ts
@@ -25,14 +25,12 @@ export const BottomRightAlignedArrowContainer = styled.div<{
   next: boolean;
   disabled: boolean;
 }>`
-  position: absolute;
-  top: 50%;
-  z-index: ${Z_SCALE.e2000};
-  transform: translate(0%, -50%);
   display: block;
+  align-self: flex-end;
+  margin: ${SPACER.small};
 
-  ${({ prev }) => prev && `left: ${SPACER.small};`};
-  ${({ next }) => next && `right: ${SPACER.small};`};
+  ${({ prev }) => prev && `order: 2; margin-left: auto;`};
+  ${({ next }) => next && `order: 3;`};
 
   ${({ disabled }) =>
     disabled &&
@@ -41,6 +39,10 @@ export const BottomRightAlignedArrowContainer = styled.div<{
         background-color: none;
         border-color: ${COLORS.primary};
         color: ${COLORS.primary};
+
+        &:hover {
+          color: ${COLORS.primary};
+        }
       }
     `};
 `;
--- a/src/shared-components/carousel/arrow/style.ts
+++ b/src/shared-components/carousel/arrow/style.ts
@@ -1,8 +1,9 @@
 import styled from '@emotion/styled';
+import { css } from '@emotion/core';
 
-import { SPACER, Z_SCALE } from '../../../constants';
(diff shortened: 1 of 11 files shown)


src/shared-components/carousel/style.ts
```diff
@@ -191,6 +191,10 @@ export const InnerContainer = styled.div<{
 
   ${reactSlickStyles};
 
+  .slick-slider {
+    flex-wrap: wrap;
+  }
+
```
AlyssaKirstine, line 197: Note: adding this in didn't alter any of the current Carousels

Yihwan (approved):
code looks fine to me, but i didn't know this carousel component existed and am probably not the best person to speak to that for continuity. 

i think this could be fine to merge for now, but would require additional monitoring and updates if the new carousel usage changed, or if we for whatever reason decided not to use it at all. 

another approach in the future could be to essentially duplicate radiance components in code, and then make any necessary one-off changes. once we're reasonably confident that the changes are something we want to keep around for a the long-term, we could update radiance accordingly and communicate changes to design as well.

src/shared-components/carousel/arrow/style.ts
```diff
@@ -19,4 +20,29 @@ const ArrowContainer = styled.div<{
...
+export const BottomRightAlignedArrowContainer = styled.div<{
+  prev: boolean;
+  next: boolean;
+  disabled: boolean;
+}>`
+  display: block;
+  align-self: flex-end;
+  margin: ${SPACER.small};
+
+  ${({ prev }) => prev && `order: 2; margin-left: auto;`};
```
daigof, line 32: Im not sure about this flex-wrap thing and it wouldn't work with dots. Could you do me a favor and try with 

```
position:absolute;
bottom: -32px;
right: 32px;
```

32 is just a random number to get you an idea and the prev button would need more than next. Please try this and let me know (also remove flex:wrap)

AlyssaKirstine: I tried this and it doesn't work because of the height of the Carousel component. Half of the arrows end up being hidden. It would only work if I add `overflow: visible` to `InnerContainer` or increase the `padding-bottom` of the `InnerContainer` or `slick-slider`, In my opinion, these seem like worse alternatives to `flex: wrap`. To remedy the dot problem, I could just force dots to be hidden when using this new prop?

daigof: ok, nvm then. you can also add that comment to the docs explainng this wont work with dots

AlyssaKirstine: Yup. I added it in [here](https://github.com/curology/radiance-ui/pull/336/commits/f796aba9a110453b6ad5930c34b1aadc409ef29e)!

AlyssaKirstine:
Merging this in so that I can move forward with the Pricing Module. I can come back and make any needed updates in the future.

PR #338, opened 2020-09-09
[Dropdown] Apply box-shadow changes 

michaeljaltamirano:
### What and Why

1. Updates our `clickable` and `clickableHover` styles to match what design has in Figma. 
1. Updates our `border-radius` styling for the Dropdown component, which previously had no `border-radius`.
    - Our Mobile dropdown will always have `border-radius: 4px` since we rely on native select dropdown for functionality. 
    - Our Desktop dropdown will have `border-radius: 4px` *unless* it is open, at which point the following occurs:
        - The Dropdown select will have `border-radius: 4px 4px 0 0`, shorthand for setting the top-left and top-right border radius.
        - The last child of the Dropdown options will have `border-radius: 0 0 4px 4px`, shorthand for setting the bottom-left and bottom-right border radius. 
1. Updates our Dropdown component to accept an optional `borderRadius` prop, so that the above `border-radius` values are programmable, with `4px` as the default. 

**Clickable changes**

Before:

Screen Shot 2020-09-09 at 2 30 40 PM

After:

Screen Shot 2020-09-09 at 2 30 33 PM

**Action button changes (as a result of `clickable` changes)``

Before:

button-before

After:

button-after

**Dropdown changes**

Before:

Screen Shot 2020-09-09 at 2 40 05 PM

After:

Screen Shot 2020-09-09 at 2 40 38 PM

### Next Steps

1. These `clickable` and `clickableHover` style changes will also be used in future PRs updating our `Accordion` and `OptionButton` component.
    - I plan on releasing all these changes together.

### DIFF
--- a/src/shared-components/dropdown/desktopDropdown.tsx
+++ b/src/shared-components/dropdown/desktopDropdown.tsx
@@ -16,27 +16,29 @@ import {
 import { OptionType } from './index';
 
 type DesktopDropdownProps = {
-  value?: string;
-  options: OptionType[];
-  currentOption: OptionType;
-  textAlign: 'left' | 'center';
+  borderRadius: string;
   closeDropdown: () => void;
+  currentOption: OptionType;
+  isOpen: boolean;
   onOptionClick: (event: any) => void;
   onSelectClick: () => void;
-  isOpen: boolean;
+  options: OptionType[];
   optionsContainerMaxHeight: string;
+  textAlign: 'left' | 'center';
+  value?: string;
 };
 
 const DesktopDropdown = ({
-  value,
-  options,
-  textAlign,
-  currentOption,
+  borderRadius,
   closeDropdown,
+  currentOption,
+  isOpen,
   onOptionClick,
   onSelectClick,
-  isOpen,
+  options,
   optionsContainerMaxHeight,
+  textAlign,
+  value,
 }: DesktopDropdownProps) => {
   const { initialFocus, resetFocus } = useResetFocus<HTMLDivElement>();
 
@@ -73,8 +75,9 @@ const DesktopDropdown = ({
         >
           <div
             css={dropdownInputStyle({
-              textAlign,
+              borderRadius,
               shouldBeFullyRounded: !isOpen,
+              textAlign,
             })}
           >
             {currentOption && currentOption.label}
@@ -85,6 +88,7 @@ const DesktopDropdown = ({
         </div>
 
         <DropdownOptionsContainer
+          borderRadius={borderRadius}
           isOpen={isOpen}
           optionsContainerMaxHeight={optionsContainerMaxHeight}
           role="menu"
@@ -131,6 +135,7 @@ DesktopDropdown.defaultProps = {
 };
 
 DesktopDropdown.propTypes = {
+  borderRadius: PropTypes.string.isRequired,
   // eslint-disable-next-line react/forbid-prop-types
   value: PropTypes.any,
   options: PropTypes.arrayOf(
--- a/src/shared-components/dropdown/desktopDropdown.tsx
+++ b/src/shared-components/dropdown/desktopDropdown.tsx
@@ -71,7 +71,12 @@ const DesktopDropdown = ({
           role="button"
           ref={initialFocus}
         >
-          <div css={dropdownInputStyle({ textAlign })}>
+          <div
+            css={dropdownInputStyle({
+              textAlign,
+              shouldBeFullyRounded: !isOpen,
+            })}
+          >
             {currentOption && currentOption.label}
           </div>
           <IconContainer>
--- a/src/shared-components/dropdown/index.tsx
+++ b/src/shared-components/dropdown/index.tsx
@@ -6,27 +6,29 @@ import DesktopDropdown from './desktopDropdown';
 import allowNullPropType from '../../utils/allowNullPropType';
 
 export type OptionType = {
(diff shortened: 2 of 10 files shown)


michaeljaltamirano:
> **Action button changes (as a result of `clickable` changes)``
> 
> Before:
> 
> button-before
> 
> After:
> 
> button-after

@benkolde Tagging you for a design review before an engineer for a code review because this behavior in particular seems suboptimal. Since our focus outline is smaller than the new `box-shadow` size the keyboard effect is somewhat jarring.

benkolde (approved):
Looks great, thanks @michaeljaltamirano

src/shared-components/dropdown/test.tsx
```diff
@@ -49,7 +49,13 @@ describe('<MobileDropdown />', () => {
   describe('UI snapshots', () => {
     it('renders correctly', () => {
       const tree = renderer
-        .create(<MobileDropdown options={options} />)
+        .create(
+          <MobileDropdown
+            onSelectChange={() => undefined}
+            borderRadius="4px"
```
michaeljaltamirano, line 55: It's a little annoying to have to provide these for the Jest tests--however, I thought it was cleaner to add the optional argument w/ default at the parent component `Dropdown`, instead of keep the value optional there (without a default) and setting the default on both `MobileDropdown` and `DesktopDropdown`.

src/shared-components/dropdown/style.ts
```diff
@@ -9,57 +9,73 @@ import {
...
+    box-shadow: ${BOX_SHADOWS.clickable};
+    background: ${COLORS.white};
+    background-image: none;
+
+    width: 100%;
+    min-height: ${SPACER.x4large};
+    max-height: ${SPACER.x4large};
+
+    border: 1px solid ${COLORS.border};
+    ${dropdownBorderRadius};
```
michaeljaltamirano, line 48: This line is the only change--the diff is larger due to indentatio.

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -16,27 +16,29 @@ import {
...
   onOptionClick: (event: any) => void;
   onSelectClick: () => void;
-  isOpen: boolean;
+  options: OptionType[];
   optionsContainerMaxHeight: string;
+  textAlign: 'left' | 'center';
+  value?: string;
 };
 
 const DesktopDropdown = ({
```
michaeljaltamirano, line 31: Larger diff is from alphabetizing--only change in this file is adding `borderRadius` (and `shouldBeFullyRounded`) logic.

src/shared-components/dropdown/index.tsx
```diff
@@ -6,27 +6,29 @@ import DesktopDropdown from './desktopDropdown';
...
-  value?: string;
+  borderRadius?: string;
   onChange: (option: OptionType) => void;
   options: OptionType[];
   optionsContainerMaxHeight?: string;
   textAlign?: 'left' | 'center';
+  value?: string;
 };
 
-const Dropdown = (props: DropdownProps) => {
```
michaeljaltamirano, line 22: CHANGELOG:

1. Move `defaultProps` to ES6 defaults. 
    - Improved type-inference from relying on language built-ins without working out `defaultProps` limitations.
        - e.g., the compiler can infer that even though `borderRadius` is optional, since we're providing it with defaults when we pass it into `DesktopDropdown` and `MobileDropdown` we can type it there as required. 
    - There's still some inconsistencies: we're passing an ES6 default in `MobileDropdown` even though we typed it as a required prop. I'm not sure if that's just due to making things easier to test (related comment: https://github.com/curology/radiance-ui/pull/338/files#r488011625) but this PR already has enough ancillary changes so I didn't want to change too much. 
1. Modify type of `value` in `OptionType`
    - We rely on a falsy default for our dropdown: changing the type from `string` to `string | undefined` reflects this. 
    - There's some inconsistency in our falsy default--`null` vs. `undefined`. The type definition for `<option>` prefers `undefined` instead of `null` (`OptionHTMLAttributes<HTMLOptionElement>.value?: string | number | readonly string[] | undefined`) so this PR prefers that when making changes.

src/shared-components/dropdown/mobileDropdown.tsx
```diff
@@ -7,21 +7,27 @@ import { DropdownContainer, dropdownInputStyle, IconContainer } from './style';
...
-  options: OptionType[];
-  textAlign: 'left' | 'center';
+  borderRadius: string;
   onSelectChange: (event: React.ChangeEvent<HTMLSelectElement>) => void;
+  options: OptionType[];
+  textAlign?: 'left' | 'center';
+  value?: string | number | undefined;
 };
 
 const MobileDropdown = ({
```
michaeljaltamirano, line 17: CHANGELOG:

1. Change default `value` and `options[number].value` type from `null` to `undefined` to conform to HTML attribute types. 
1. Remove defaultProps for `onSelectChange`
    - I'm assuming this was included to make the tests easier--it's a required TypeScript type so this makes things more consistent. 
1. Sets `textAlign` TypeScript type as optional to reflect that we have an ES6 default. 
    - This isn't consistent with parent `Dropdown` behavior but I didn't want to remove this functionality. All sub-component usage goes through `Dropdown`.

ZeMunchkin (approved):
💯

cmgmyr (approved):
👏

PR #342, opened 2020-09-10
[CI] Expand Build Step

michaeljaltamirano:
TODO


michaeljaltamirano:
Closing for now since we'd need to do more purposeful tool updating (set higher node version) + orchestration (way to auto-opt-in on size changes). Potential cooldown project.

PR #351, opened 2020-09-10
[CI] Expand Build Step

michaeljaltamirano:
Our `build` step in CI just installs dependencies, so it doesn't really test much. This PR adds our `yarn build` command to log the generated output as well.

### Looking Forward

We should incorporate some size checks into our CI pipeline to track how our dist bundle grows/shrinks with changes. It will also be easier to track dependency changes we're making to see if anything is inadvertently ballooning filesize. (See #342)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,8 +48,10 @@ jobs:
             - <<: *yarn-cache-restore
             - run:
                 name: Install
-                command: |
-                    yarn install
+                command: yarn install
+            - run:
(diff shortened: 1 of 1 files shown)


michaeljaltamirano:
The build work here will likely be made redundant by #409 but I will keep this draft open as a reminder to get the build size metrics in.

michaeljaltamirano:
Closing, added `bundlewatch` in #667

PR #352, opened 2020-09-11
[Dependencies] Upgrade Storybook from v5 to v6

michaeljaltamirano:
### What & Why

This PR upgrades Storybook from v5 to v6. This branch started by running `npx sb upgrade` and then running through the migration guides to get a working build:

Medium migration guide: https://medium.com/storybookjs/storybook-6-migration-guide-200346241bb5
Comprehensive migration guide: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-53x-to-60x

Fortunately (or unfortunately) most of the changes do not apply to us because we're using the old `storiesOf` syntax instead of the [Component Story Format](https://storybook.js.org/docs/react/api/csf). However, this is fine (for now) since `storybook-readme` does not yet support v6. I plan to fully deprecate `storybook-readme` in a follow-up PR (#416) but will maintain compatibility for the main `@storybook` package major updates from 5 to 6 in this PR. 

The vast majority of the diff is from two files: `yarn.lock` and `.size-snapshot.json`. There are not that many actual changes. 

### Outcome

This PR updates the underlying Storybook dependencies from v5 to v6, allowing us to access now functionality as we build in `radiance-ui`. There are no regressions with the published Storybook, and no changes to the built component library imported via `node_modules`. 

### CHANGELOG:

1. Expands `.babelrc` for additional TypeScript functionality within `./storybook`. (originally added in #334. Removes TypeScript webpack loader (since Storybook supports it by default).
    - This seemed to be the arrangement to get type-checking + build working without issue. 
1. Adds `.storybook/main.ts`
    - This is the new preferred method for loading stories, not `.storybook/preview.js`
    - It's also preferred to include `addons` in this file instead of `/register` imports in `.storybook/manager.js`
1. Removes `stories.addDecorator(withKnobs)` usage in individual stories.
    - New preset does this automatically for it. (You can opt-out at the story level via parameters. Not too worried about this since we might want to consider moving to `addon-controls` instead of knobs in the future.)
1. Updates some settings in `./storybook/preview.ts`. 
1. General TypeScript conversion. 

I've added a couple comments in the PR, as well.

### Next Steps

1. Complete component story + documentation overhaul (https://github.com/curology/radiance-ui/pull/416) in anticipation of setting up visual regression testing for components (#409)


michaeljaltamirano:
@daigof Going to keep working on this branch, FYI. If you've got other work in `radiance-ui` coming up feel free to let me know if it blocks you. I don't think it should, though.

package.json
```diff
@@ -66,15 +66,15 @@
     "@rollup/plugin-babel": "^5.2.1",
     "@rollup/plugin-commonjs": "^15.0.0",
     "@rollup/plugin-node-resolve": "^9.0.0",
-    "@storybook/addon-a11y": "5.3.21",
-    "@storybook/addon-actions": "5.3.21",
-    "@storybook/addon-centered": "5.3.21",
```
michaeljaltamirano, line 71: Removes `@storybook/addon-centered`, now deprecated, as new core Storybook settings suffice, as adds `@storybook/theming` explicitly, since we use it in `.storybook/theme.ts`.

.storybook/preview.tsx
```diff
@@ -71,8 +71,6 @@ const InjectGlobalStyles = (storyFn) => (
 );
 
 addDecorator(InjectGlobalStyles);
-addDecorator(centered);
-addDecorator(withA11y);
```
michaeljaltamirano, line 75: a11y options now set in `addParameters`.

.size-snapshot.json
```diff
@@ -0,0 +1,1356 @@
...
+      "rollup": {
+        "code": 433,
+        "import_statements": 105
+      },
+      "webpack": {
+        "code": 1501
+      }
+    }
+  },
+  "shared-components/dropdown/index.js": {
```
michaeljaltamirano, line 502: This is a useful file that I think we should be committing. Since it's not currently in the repo, it's all new, but I used it to try and track down why there was a very tiny change in bundle size, and it's due to `prettier` auto-formatting in `src/dropdown`. 

I will look into a way to more seamlessly integrate this into our git workflow. It only matters for releases so we don't need to enforce it at the individual PR level.

daigof (approved):
Im getting a small console error while running storybook  but it does work correctly anyway.
The build doesn't seem to be affected. 
Yeah great job on this upgrade I see there is more in the pipeline so I guess is safe to merge this now to get the ball rolling

PR #371, opened 2020-09-15
[Accordion] Update Accordion.Container-based Styling

michaeljaltamirano:
This PR sets out to make the following changes to the Accordion component:

1. Update the `box-shadow` property.
1. Add a default `4px` border-radius.
1. Add the following border-radius behavior: only the top-most and bottom-most accordion elements should be rounded.

The first two updates are relatively trivial. The third update required expanding the styling functionality of `Accordion.Container`, which has additional CSS selectors that handle the `border-radius` & focus outline requirements. 
`border-radius` is set via long-hand properties `border-top-left-radius`, `border-top-right-radius`, `border-bottom-left-radius`, and `border-bottom-right-radius`. 

This PR also updates all documentation instances of `<Accordion>` to be wrapped with `<Accordion.Container>`. 

You can play with the Review App here: https://curology-radiance-pr-371.herokuapp.com/

**Before**:

before

**After**:

after


src/shared-components/accordion/style.ts
```diff
@@ -42,6 +52,8 @@ export const AccordionBox = styled.div<{
 }>`
   ${({ noBorder, isOpen }) => (!noBorder ? getBorderStyle(isOpen) : '')};
 
+  position: relative;
```
michaeljaltamirano, line 55: You can play with this in the dev tools but without it the box-shadow in the focus state is clipped by the border. It's not a total improvement but it's a small thing before I get to the accessibility updates this cycle, which is changing out focus state anyway.

benkolde (approved):

cmgmyr (approved):

src/shared-components/accordion/style.ts
```diff
@@ -2,16 +2,26 @@ import styled from '@emotion/styled';
...
 
+const getTopBorderRadius = (borderRadius: string) => ({
+  borderTopLeftRadius: borderRadius,
+  borderTopRightRadius: borderRadius,
+});
+
+const getBottomBorderRadius = (borderRadius: string) => ({
+  borderBottomLeftRadius: borderRadius,
+  borderBottomRightRadius: borderRadius,
+});
```
ZeMunchkin, line 21: I'm on the fence about these functions, since they seem to only give `borderRadius` two additional names without any modifications. It seems like it might be adding some unnecessary complexity, since `borderRadius` could be used in place of these to the same effect.

ZeMunchkin (commented):
Had a couple questions I left comments on. One might simplify the issue of enforcing the use of Grouping in the codebase, depending on how it shakes out.

docs/accordion.md
```diff
@@ -112,26 +126,50 @@ The disabled accordion should only be used when the user is taking a clearly def
...
+    <Accordion title={<Accordion.Content>First Title</Accordion.Content>}>
+      <Accordion.Content>First Expansion</Accordion.Content>
+    </Accordion>
+  </Accordion.Container>
+  <Accordion.Container>
+    <Accordion title={<Accordion.Content>Second Title</Accordion.Content>}>
+      <Accordion.Content>Second Expansion</Accordion.Content>
+    </Accordion>
+  </Accordion.Container>
+</Accordion.Grouping>;
```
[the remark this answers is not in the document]
michaeljaltamirano: Oh interesting. Since there were some usages of `<Accordion>` without the container I had assumed that it just wraps a single `<Accordion>`. I should have considered more strongly that the docs read `Container to hold instances of <Accordion>` even though all the code examples only had one `<Accordion>` present. Let me refactor.

src/shared-components/accordion/style.ts
michaeljaltamirano: The long-form properties are used because the `:first-of-type` and `:last-of-type` conflict when there is only one `<Accordion>` item. Using a single `border-radius` would override one or the other. 

I can remove the function (or rename suffix to `Radiuses`?) if it's unclear.

docs/accordion.md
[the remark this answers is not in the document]
michaeljaltamirano: @ZeMunchkin Thank you again for catching this. Updated in https://github.com/curology/radiance-ui/pull/371/commits/03af86c950f0805b6c9939d3f5323253bd55e300 (and story markup formatting fixed in https://github.com/curology/radiance-ui/pull/371/commits/7d95b2b3a04dbea5bf8dcfd1c5fd223cdb5cea06). I've also updated the PR description.

src/shared-components/accordion/style.ts
michaeljaltamirano: @ZeMunchkin Opted to just remove the helpers entirely in https://github.com/curology/radiance-ui/pull/371/commits/ac26593096f3a93d5643702a403b503c80b458e4 🙂

michaeljaltamirano:
Need to refactor the styling since the box-shadows are not applying accurately anymore in `TitleWrapper`.

ZeMunchkin (approved):
LGTM! 🎊

PR #386, opened 2020-09-21
Move Arrow to exported icons directory

ipc103:
As a part of our Messages V1 updates, we want to render this arrow above our Alert component in a similar fashion to the tooltip. This PR simply exports the Arrow icon so that we can render it alongside the Alert. 

This seemed a little bit more flexible than trying to extend the existing `Alert` component, but I'm open to other approaches if folks disagree. 

Screen Shot 2020-09-21 at 4 36 19 PM


jashieh (approved):
Maybe the arrow could be included as an optional flag? Like add pointer and location top/bottom

ipc103:
> Maybe the arrow could be included as an optional flag? Like add pointer and location top/bottom

Yeah, that could work, but we also need a way to adjust how far to the left or right the renders, which in our case will vary by the size of the display. For now, I *think* it's a little bit more flexible to just render the pointer separately. It will be easier to delete the implementation in PocketDerm and replace it with an implementation here than it will to try to rip that behavior out once we add it.

PR #390, opened 2020-09-23
[a11y] Update focus states

michaeljaltamirano:
### What and Why

This PR adds two new properties to our `BOX_SHADOWS` constants:

1. `focus`
1. `focusInner`. 

The behavior of `focus` is two separate box-shadows to simulate an outline with margin between the element and the outline. Here is the change of behavior in the `Button` component:

Before:

Screen Shot 2020-09-23 at 4 34 30 PM

After:

Screen Shot 2020-09-23 at 4 34 44 PM

The behavior of `focusInner` is an `inset` box-shadow, such that the desired outline appears within the given element. Here is the change of behavior in the `Accordion` component:

Before:

Screen Shot 2020-09-23 at 4 36 16 PM

After:

Screen Shot 2020-09-23 at 4 36 08 PM

This PR also adds a polyfill for `focus-visible`, a CSS pseudo-class that effectively allows us to apply custom styling to keyboard navigation that does not take effect for mouse navigation. ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible)). This allows us to more gracefully introduce accessibility styling, such as helpful keyboard focus states, and limit the styling to keyboard navigation, keeping the existing mouse behavior. 

This PR piggy-backs off of our existing global `resetStyles` usage, adding `focusStyles` to it, such that our usage is automatically opt-in without concerning ourselves with adding a third global usage, with `brandStyles` being the other global usage. We want `focus-visible` to apply regardless of brand. 

### Other

1. Adds a simple `src/utils/styles/buttonReset` export. Our transitioning from `div` to `button` use for accessibility means we write a lot of `background` and `border` overrides. 
1. Updates some `eslint` configuration to remove noisy rules. 
1. Fixes the mobile dropdown story, which did not reflect the border-radius changes made in #338. 
1. Cleans up outdated `yarn.lock` packages from `dependabot` PR merges. 
1. Updates Dropdown tests that were failing due to relying on a styled component to help apply the inset `box-shadow` to the DesktopDropdown component. Uses `shallow` instead of `mount`. Presumably, using a styled component instead of a div is creating some prop-forwarding issues that was resulting in there being two buttons we were finding/simulating clicking. I didn't investigate it deeper, because it will be obvious to us if click behavior breaks. 

### Next Steps

1. Update `toggle` component.
    - It uses a dependency that hasn't been updated in 3 years. I will update that component in a separate PR
    - We want to apply the `focus` styling to it. 
1. Update `alert` component
    - We want to apply the `focus` styling to it.

Once those two items are complete, I will release a new beta version of this library for QA, in anticipation of a new major release.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,6 +36,7 @@
     "@typescript-eslint/naming-convention": "off",
     "@typescript-eslint/no-shadow": "error",
     "@typescript-eslint/no-use-before-define": "error",
+    "arrow-parens": "off",
     "no-shadow": "off",
     "no-use-before-define": "off",
     "react/jsx-filename-extension": ["warn", { "extensions": [".js", ".tsx"] }],
@@ -56,7 +57,8 @@
         "emotion/jsx-import": "error",
         "emotion/no-vanilla": "error",
         "emotion/import-from-emotion": "error",
-        "emotion/styled-import": "error"
+        "emotion/styled-import": "error",
+        "react/jsx-props-no-spreading": "off"
       }
     },
     {
--- a/package.json
+++ b/package.json
@@ -131,6 +131,7 @@
     "@emotion/core": "^10.0.35",
     "@emotion/styled": "^10.0.27",
     "@react-aria/focus": "^3.0.2",
+    "focus-visible": "^5.1.0",
     "lodash.round": "^4.0.4",
     "lodash.throttle": "^4.1.1",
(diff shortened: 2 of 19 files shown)


cmgmyr (approved):

ZeMunchkin (approved):
🥇

benkolde (approved):
These look incredible, @michaeljaltamirano !

PR #394, opened 2020-09-24
[a11y] Add focus to Toggle

michaeljaltamirano:
### What & Why 

This PR continues the work in #390 to add a focus state around our Toggle component. "Button," used in the markup (and underlying dependency) is a misnomer because it's really a checkbox component, so keyboard navigation requires a `space` to toggle rather than `enter`. 

Before:

Screen Shot 2020-09-24 at 10 37 49 AM

After: 

Screen Shot 2020-09-24 at 10 39 10 AM

### Other

1. This PR converts all of the files to TypeScript. I will lint the PR with the specific changes in the component.
1. I noticed in #390 that `react-toggle-button` is a dependency that hasn't been updated in 3+ years. It might be worth switching to an alternative in the future (https://react-spectrum.adobe.com/react-spectrum/Switch.html, https://chakra-ui.com/switch) but it wasn't worth the scope creep right now. It's worth noting we only use this component in one place.


src/shared-components/toggle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+ * toggleInputStyles overrides the defaults.
+ */
+const toggleInputStyles = {
+  margin: '0',
+  height: '24px',
+  width: '40px',
+  position: 'absolute',
+};
+
+const Toggle = ({ checked = false, label = '', onChange }: ToggleProps) => (
```
michaeljaltamirano, line 38: `CHANGELOG`:

1. Update typing: make `onChange` required and rely on ES6 defaults instead of `defaultProps`. 
    - This required passing an `onChange` prop in the tests, too, which is the only changes there. 
1. Use `passThroughInputProps` option to override in-line styles on underlying `input` component. 
    - This was required to achieve the desired behavior via selectors in `ReactToggleButtonContainer`.

cmgmyr (approved):

benkolde (approved):

PR #397, opened 2020-09-25
[a11y] Alert Changes

michaeljaltamirano:
### What & Why

Uses a `button` as the underlying element instead of a `div` for the AlertContainer. This has the consequences of an extremely minor shift in text, one that I couldn't identify the cause of when looking at the computed styles. Here's a gif to show what I mean:

text

Ultimately I don't think it's consequential so I'm leaving just documenting this change as sufficient 🙂 

It also applies our new `focus` style:

Screen Shot 2020-09-25 at 2 06 48 PM

### Other
There's an open question of *how* we want the focus for an alert to behave. Should we move the focus immediately to the alerts? Should we trap the focus? After dealing with the alerts, should we restore the focus to the item that was previously focused? I tried looking through other component libraries but our alert seems particular in that it wants to draw the user's attention but not so much that it is blocking, so containing the focus seems extreme. 

Ultimately making a decision on stricter focus behavior seems like something we can figure out with greater adoption of keyboard usage + specific UX research around it. Maybe we trap focus but only for high-priority alerts. In any case, it is premature to make those decisions at this point in development.

### DIFF
--- a/src/shared-components/alert/__snapshots__/test.tsx.snap
+++ b/src/shared-components/alert/__snapshots__/test.tsx.snap
@@ -54,6 +54,11 @@ exports[`Alert UI snapshots renders a default alert 1`] = `
   box-shadow: 0px 8px 24px rgba(51,46,84,0.05);
 }
 
+.emotion-6:focus {
+  outline: none;
+  box-shadow: 0px 0px 0px 2px #ffffff,0px 0px 0px 4px #332e54;
+}
+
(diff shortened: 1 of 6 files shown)


cmgmyr (approved):
This looks great to me! Not sure about the potential focus changes, but I think how you have it now will be fine and we can revisit in the future once there is some more data like you mentioned

PR #400, opened 2020-09-25
[a11y] Other Focus States

michaeljaltamirano:
### What & Why

This PR contains the final changes made to our accessibility/keyboard focus states before cutting a major release. Many of changes are simply replacing `BOX_SHADOWS.focusSecondary` for `BOX_SHADOWS.focus` and then updating the snapshots. The below notes cover the other changes:

1. Remove `BOX_SHADOWS.focusSecondary`
    - `BOX_SHADOWS.focus` and `BOX_SHADOWS.focusInner` will be the preferred focus states going forward. 
1. Makes `Banner` a button instead of div. 
    - `Banner` optionally accepts an `onClick` prop. When it is present, we'll make it keyboard focusable. Otherwise, we'll treat it as a div. (No interactivity.)
    - Removes `defaultProps` in favor of ES6 defaults. 
1. Converts `src/shared-components/immersiveModal/index.tsx` to a function component.
    - If we want to keep the diff smaller, I am pretty sure I can set this back to the old class component code. I converted it because I thought I had to in order to take advantage of `useFocusManager`, but it became clear while working on it that concern was better separated into a separate component, which I've named `CrossIconComponent`. I'm open to changing it 😓 
        - My feeling is: if it works, might as well keep the newer code. 
        - I've left an additional comment on that specific diff with a `CHANGELOG` to make reviewing it easier.

### Components updated:

**Banner**: This was not scoped but since we allow onClick functionality we need to support it: 

image

**DialogModal** close icon before and after:
 
Screen Shot 2020-09-28 at 3 35 22 PM

Screen Shot 2020-09-28 at 3 35 31 PM

**Field** before and after:

Screen Shot 2020-09-28 at 3 36 45 PM
Screen Shot 2020-09-28 at 3 36 32 PM

**ImmersiveModal**: same as DialogModal, but there was previously no outline. 

**OptionButton** before and after:

Screen Shot 2020-09-28 at 3 38 59 PM

Screen Shot 2020-09-28 at 3 38 41 PM

**SelectorButton** (underlying component for Checkbox and RadioButton) befores and afters:

Screen Shot 2020-09-28 at 3 39 29 PM
Screen Shot 2020-09-28 at 3 39 36 PM
Screen Shot 2020-09-28 at 3 39 22 PM
Screen Shot 2020-09-28 at 3 39 14 PM

**TextButton**:

Screen Shot 2020-09-28 at 3 38 23 PM

### Other

I did not add transitions to components that did not already have them, so tabbing between the dedicated Buttons story is a very different experience than Checkbox/Option/Radio button components. I am okay with that, but this probably represents an area where Design and Eng might have a different understanding of behaviors, so I'm calling it out. cc @benkolde for visibility/FYI on the Design side.

### Release Notes

- We will be removing `BOX_SHADOWS.focusSecondary` and cutting a new major release after merging this in. We will need to replace it with `BOX_SHADOWS.focus` in direct `BOX_SHADOWS.focusSecondary` usages. 
    - It does not exist in our Design assets anymore, so we should not keep it around.

### DIFF
--- a/src/shared-components/banner/style.ts
+++ b/src/shared-components/banner/style.ts
@@ -1,7 +1,9 @@
 import styled from '@emotion/styled';
 import { css } from '@emotion/core';
+import { buttonReset } from 'src/utils/styles/buttonReset';
 
 import {
+  BOX_SHADOWS,
   COLORS,
   MEDIA_QUERIES,
   SPACER,
@@ -25,7 +27,17 @@ const errorAlertStyles = css`
   box-shadow: 0px 8px 24px rgba(189, 32, 15, 0.05);
 `;
 
-export const BannerContainer = styled.div<{ bannerType: BannerType }>`
+/**
+ * width: 100% to match previous <div> behavior
+ */
+export const BannerContainer = styled.button<{ bannerType: BannerType }>`
+  ${buttonReset}
+  width: 100%;
+  &:focus {
+    outline: none;
+    box-shadow: ${BOX_SHADOWS.focus};
+  }
+
   cursor: pointer;
   position: relative;
   margin: 0 auto ${SPACER.small};
--- a/src/shared-components/banner/style.ts
+++ b/src/shared-components/banner/style.ts
@@ -27,9 +27,6 @@ const errorAlertStyles = css`
   box-shadow: 0px 8px 24px rgba(189, 32, 15, 0.05);
 `;
 
-/**
- * width: 100% to match previous <div> behavior
- */
 export const BannerContainer = styled.button<{
   bannerType: BannerType;
   onClick?: () => void;
--- a/src/shared-components/banner/style.ts
+++ b/src/shared-components/banner/style.ts
@@ -30,7 +30,10 @@ const errorAlertStyles = css`
 /**
  * width: 100% to match previous <div> behavior
  */
-export const BannerContainer = styled.button<{ bannerType: BannerType }>`
+export const BannerContainer = styled.button<{
+  bannerType: BannerType;
+  onClick?: () => void;
+}>`
   ${buttonReset}
   width: 100%;
   &:focus {
@@ -38,7 +41,7 @@ export const BannerContainer = styled.button<{ bannerType: BannerType }>`
     box-shadow: ${BOX_SHADOWS.focus};
   }
 
-  cursor: pointer;
+  cursor: ${({ onClick }) => (onClick ? `pointer` : `inherit`)};
   position: relative;
   margin: 0 auto ${SPACER.small};
   padding: ${SPACER.medium};
--- a/src/shared-components/button/components/textButton/style.ts
+++ b/src/shared-components/button/components/textButton/style.ts
@@ -1,10 +1,11 @@
 import styled from '@emotion/styled';
 
-import { BOX_SHADOWS, COLORS } from '../../../../constants';
+import { BOX_SHADOWS, COLORS, SPACER } from '../../../../constants';
 
 export const BaseTextButton = styled.button<{ disabled: boolean }>`
   border-color: transparent;
   background-color: transparent;
+  border-radius: ${SPACER.xsmall};
 
   color: ${({ disabled }) =>
     `${disabled ? COLORS.textDisabled : COLORS.primary}`};
@@ -18,6 +19,6 @@ export const BaseTextButton = styled.button<{ disabled: boolean }>`
   &:active,
   &:focus {
     outline: none;
-    box-shadow: ${BOX_SHADOWS.focusSecondary};
+    box-shadow: ${BOX_SHADOWS.focus};
   }
 `;
--- a/src/shared-components/immersiveModal/crossIconComponent/index.tsx
+++ b/src/shared-components/immersiveModal/crossIconComponent/index.tsx
@@ -5,7 +5,7 @@ import CrossIcon from '../../../svgs/icons/cross-icon.svg';
 import { CrossIconContainer } from '../style';
 
 type CrossIconComponentProps = {
-  showDesktopHeaderBar: boolean;
+  isVisible: boolean;
   onClick: () => void;
 };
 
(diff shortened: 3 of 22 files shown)


src/shared-components/button/components/textButton/style.ts
```diff
@@ -1,10 +1,11 @@
 import styled from '@emotion/styled';
 
-import { BOX_SHADOWS, COLORS } from '../../../../constants';
+import { BOX_SHADOWS, COLORS, SPACER } from '../../../../constants';
 
 export const BaseTextButton = styled.button<{ disabled: boolean }>`
   border-color: transparent;
   background-color: transparent;
+  border-radius: ${SPACER.xsmall};
```
michaeljaltamirano, line 8: Added for rounded focus state.

stories/immersiveModal/immersiveModalExamples.tsx
```diff
@@ -131,10 +131,12 @@ const ImmersiveModalExamples = (): JSX.Element => {
       </Button>
       {withButtons && (
         <ImmersiveModal
-          onClose={(): void => setWithButtons(false)}
+          onClose={() => setWithButtons(false)}
           footerContent={
             <Button.Container>
-              <Button isFullWidth>cta content</Button>
+              <Button isFullWidth onClick={() => setWithButtons(false)}>
```
michaeljaltamirano, line 137: Just makes the storybook example more correct--CTAs should have onClicks that does an action.

src/shared-components/immersiveModal/index.tsx
```diff
@@ -22,193 +23,177 @@ import {
...
 } from './style';
+import CrossIconComponent from './crossIconComponent';
 
 type ImmersiveModalProps = {
   children: React.ReactNode;
   headerImage?: React.ReactNode;
   footerContent?: React.ReactNode;
   onClose: () => void;
   title?: string;
+  [key: string]: unknown;
```
michaeljaltamirano, line 34: `CHANGELOG`:

1. Expands `ImmersiveModalProps` since it appears we spread `...rest` into `<Overlay>`. 
1. Adds `FocusScope` to contain keyboard navigation inside the modal.
    - Inspiration drawn from `DialogModal` configuration, though this implementation required some additional hand-holding of the focus state (via `useFocusManager` in `CrossIconComponent`. The component has comments explaining the what/why). 
    - (The majority of the `ImmersiveModal` changes are, basically, kind of overkill to get the `contain` behavior working "optimally".)
1. Updates ID usage into `SCREAMING_CASE` variables, `getElementById` instead of `querySelector`, etc. 
1. Uses `useRef` as replacement for previous lifecycle method reliance to set instance variables.
1. Uses ES6 defaults in favor of `defaultProps`. 
1. Updates `CrossIconContainer` to be a `button` instead of a `div`. 
1. Uses `tabIndex` to remove the hidden button from keyboard navigation.

benkolde (approved):
I think these look good!

src/shared-components/banner/style.ts
```diff
@@ -25,8 +27,21 @@ const errorAlertStyles = css`
   box-shadow: 0px 8px 24px rgba(189, 32, 15, 0.05);
 `;
 
-export const BannerContainer = styled.div<{ bannerType: BannerType }>`
-  cursor: pointer;
+/**
+ * width: 100% to match previous <div> behavior
```
ipc103, line 31: [super-nit, sorry!] I think this comment actually makes things more confusing, because you need to look at the history to see that this was previously a `div`. I think it will be clear that removing the `width` would cause a style regression.

src/shared-components/immersiveModal/crossIconComponent/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * We wrap the second of the CrossIcons we render as a React component
+ * in order to use the @react-aria/focus useFocusManager hook.
+ *
+ * This allows us to automatically toggle focus state between the two separate
+ * CrossIcons present in the markup. (We keep both rendered for our transitions.)
+ */
+const CrossIconComponent = ({
+  showDesktopHeaderBar,
```
ipc103, line 20: [super-nit] - I might name this `isVisible` or some such. I think that's a little bit easier to understand (when the prop becomes not visible, it passes focus back to another focusable thing) by reading this file alone, instead of needing to pass back and forth to the `immersiveModal` file. May not make any difference, as folks are likely to start at the ImmersiveModal anyway, so feel free to disregard.

src/shared-components/immersiveModal/crossIconComponent/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+type CrossIconComponentProps = {
+  showDesktopHeaderBar: boolean;
+  onClick: () => void;
+};
+
+/**
+ * We wrap the second of the CrossIcons we render as a React component
+ * in order to use the @react-aria/focus useFocusManager hook.
+ *
+ * This allows us to automatically toggle focus state between the two separate
```
ipc103, line 16: Super cool - very well done 💯

src/shared-components/immersiveModal/index.tsx
```diff
@@ -22,193 +23,177 @@ import {
...
-    if (this.modalMobileScrollingElement && this.modalDesktopScrollingElement) {
-      this.modalMobileScrollingElement.addEventListener(
-        'scroll',
-        this.handleScroll,
-      );
-      this.modalDesktopScrollingElement.addEventListener(
-        'scroll',
-        this.handleScroll,
-      );
+  const handleEscapeKey = (event: KeyboardEvent) => {
```
ipc103, line 90: During our last cycle, I updated the `OffClickWrapper` to treat `Escape` key presses as off-clicks, but I clearly forgot to remove the logic from this component 🙃  We should be able to remove this logic from the `ImmersiveModal` if you like.

ipc103 (approved):
This looks great - really nice use of the `useFocusManager` hook from `react-aria`. Great to have more examples of thinking through keyboard usage. Passing the focus responsively is particularly cool - nice work!

All of my comments are nits - feel free to disregard entirely.

src/shared-components/immersiveModal/crossIconComponent/index.tsx
michaeljaltamirano: I'm happy to rename things. I just don't typically rename things that are already in place, but when a reviewer suggests it, that's all the approval I need. 😈

cmgmyr (approved):

PR #405, opened 2020-09-28
[Components] Add SegmentedControl

### LINKED ISSUE #873: Add Segmented Control
In Figma but not yet implemented (see #405)

heyamykate:
Fixes #873

[Figma](https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance-UI?node-id=3685%3A0)

This PR adds a new component to the component library - `<SegmentedControl />`

It is very similar to the `Tabs` component we have - it accepts an array of objects to render into segments, and has an optional `onClick` handler prop as well. When a user clicks on one of the buttons, an "active state" button slides over to wherever the new active button resides in the DOM.

To run it in storybook format, pull down the branch and run `yarn && yarn run storybook:dev` and click on the `SegmentedControl` sidebar menu item.

Screen Shot 2020-09-29 at 3 21 31 PM

Screen Shot 2020-09-28 at 11 49 40 AM


src/shared-components/segmentedControl/style.ts
```diff
@@ -0,0 +1,65 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  cursor: pointer;
+  margin: 0;
+  border-radius: 80px;
+  background-color: ${COLORS.purple10};
+  width: ${({ width }) => `${width}%`};
+  &:focus {
+    outline: none;
```
michaeljaltamirano, line 44: Ben will probably be able to provide more guidance re: styling for different states, but there's a new `focus` variable you can use here to start playing around with how the component would work with only keyboard controls:

```
&:focus {
  outline: none;
  box-shadow: ${BOX_SHADOWS.focus};
}

Comment (benkolde) [reply src/shared-components/segmentedControl/style.ts:44]: ah yeah, thanks for calling that out @michaeljaltamirano but we should apply the new focus styling to this: unsure on how it will react between the two options `focus` and `focusInner` but would it be possible to get a screenshot of both with the two scenarios below. These are how it looks with designs but unsure how it will with live code

image

**Review by benkolde — COMMENTED** [review]
Looks good, once we figure out the focus piece it's good to go

**Comment by benkolde** [comment]
Adding comment here for posterity: 

The whole control should be `40px` tall and should use the `caption` typography sizing

File: src/shared-components/segmentedControl/style.ts (hunk 1 above)
Comment (michaeljaltamirano) [reply src/shared-components/segmentedControl/style.ts:44]: @heyamykate I ended up getting out of being underwater this cycle--can you re-tag me for review once the keyboard controls are in place?

File: stories/segmentedControl/index.tsx [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+  { id: 2, text: 'Option 2' },
+  { id: 3, text: 'Option 3' },
+];
+
+const twoItems = [
+  { id: 1, text: 'Option 1' },
+  { id: 2, text: 'Option 2' },
+];
+
+stories.addDecorator(withKnobs);
```
michaeljaltamirano, line 25: We can remove `knobs` from the story since we're not adding one that uses them.

src/shared-components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const SegmentedControl: React.FC<SegmentedControlProps> = ({
+  segmentItems,
+  initialActiveId = 1,
+  onClick,
+}) => {
+  const initialActiveItem = segmentItems.find(
+    (item: SegmentItemType) => item.id === initialActiveId,
+  );
+  const initialActiveIndex = initialActiveItem
```
michaeljaltamirano, line 14: Are there any instances in which the `id` and the `index` of the active segment differ by more than one? Could we reduce the state management overhead by just relying on one of them?

src/shared-components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    ? segmentItems.indexOf(initialActiveItem)
+    : 0;
+
+  const targetRef = useRef<HTMLDivElement>(null);
+  const [activeSegmentId, setActiveSegmentId] = useState(initialActiveId);
+  const [activeSegmentText, setActiveSegmentText] = useState(
```
michaeljaltamirano, line 20: Independent of the changes to working with the `initialActiveIndex` (though things are made easier if we're no longer concerned with that), there's a couple changes here that I think would be beneficial. Here's the code block I came up with (not yet finalized):

```tsx
const getInitialSegmentText = (
  segmentItems: SegmentItemType[],
  initialActiveId: number,
) => {
  const segmentItem = segmentItems.find((item) => item.id === initialActiveId);

  if (!segmentItem) {
    // TODO: How should we handle this if segmentItems is an empty array?
    return segmentItems[0].text;
  }

  return segmentItem.text;
};

const SegmentedControl: React.FC<SegmentedControlProps> = ({
  segmentItems,
  initialActiveId = 1,
  onClick,
}) => {
  const [activeSegmentText, setActiveSegmentText] = useState(() =>
    getInitialSegmentText(segmentItems, initialActiveId),
  );
```

1. We have `initialActiveItem?.text` as the initial value of `activeSegmentText`, which seems like a code smell. I think we should follow the TypeScript compiler's lead on `.find` always potentially returning undefined and handle a few more cases:
    - If the `initialActiveId` we are providing does not match anything in the `segmentItems`, should we default to the first item in `segmentItems`? Are there any scenarios in which the `id`s and expected sort order wouldn't align that we would need to address?
    - What should the behavior of the component be if we pass in an empty array? The component functionality all depends on there being at least one value. Should we raise an error? We should just `return null` at the top of the function if the length is 0? (Or even 1? Does a SegmentedControl make sense with fewer than 2 options?)
1. Less important--but maybe more fun!--is that we can pass a function into `useState` to allow us to do the calculate the initial state only once. ([Here's the article where I read about it](https://www.benmvp.com/blog/four-characters-optimize-react-component/)). Right now, every time the component re-renders (which is any time we select a new option), we're still doing the calculations to find the `initialActiveItem`, even though we only need to do that once. Using a callback in `useState` limits that code execution to just when it is setting the initial state.

src/shared-components/segmentedControl/style.ts
michaeljaltamirano: @heyamykate It looks like there's some CSS finessing to match the above:

Active item has focus state:
Screen Shot 2020-10-07 at 10 20 55 AM

Inactive item has focus state:
Screen Shot 2020-10-07 at 10 20 48 AM

src/shared-components/segmentedControl/index.tsx
heyamykate: I read that same article! I'll take a look 👍

heyamykate: I streamlined the component by removing the active ID logic in favor of using only the index, as you suggested 👍 

I also added a check to return `null` if the length of `segmentItems` is zero. It doesn't make a lot of sense to have only one item, but I wouldn't rule out the possibility.

heyamykate:
@benkolde how is this for focus styling?

Here's the indicator/segment with focus styling applied
Screen Shot 2020-10-08 at 3 19 38 PM

Here's the inactive segment with focus styling
Screen Shot 2020-10-08 at 3 24 01 PM

michaeljaltamirano:
@heyamykate I ran out of time this week before going OOO--if you could get one more person to look at this it would be appreciated.

src/shared-components/segmentedControl/style.ts
LilCare: @heyamykate The focused state looks perfect for the active item.  However, if an inactive item adjacent to the active item, it still has a small piece cutoff:
Screen Shot 2020-10-20 at 8 58 35 AM

If you add a z-index, it should look like below:
Screen Shot 2020-10-20 at 9 01 36 AM

src/shared-components/segmentedControl/style.ts
```diff
@@ -0,0 +1,77 @@
...
+  cursor: pointer;
+  margin: 0;
+  position: absolute;
+  border-radius: 80px;
+  width: ${({ width }) => `${width}%`};
+  top: 0;
+  left: 0;
+  bottom: 0;
+  background-color: white;
+  transition: transform 0.22s cubic-bezier(0.22, 1, 0.36, 1);
```
LilCare, line 66: The animation looks sooo good.

LilCare (approved):
I added to the thread about focus states, but this looks great otherwise!
celebration

michaeljaltamirano:
@heyamykate I updated the story after merging in #416, heads up!

src/shared-components/segmentedControl/style.ts
```diff
@@ -0,0 +1,78 @@
...
+  border: none;
+  ${TYPOGRAPHY_STYLE.caption};
+  color: ${COLORS.purple70};
+  border-radius: 80px;
+  background-color: ${COLORS.purple10};
+  width: ${({ width }) => `${width}%`};
+  &:focus {
+    outline: none;
+    box-shadow: 0px 0px 0px 2px ${COLORS.primary};
+    z-index: 3;
```
michaeljaltamirano, line 49: We have a z-index constants file that might be worth using here, just to avoid hard-coded values: `src/constants/zScale/index.ts`

src/shared-components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const onSegmentClick = (segment: SegmentItemType, index: number) => {
+    setActiveSegmentText(segment.text);
+    setActiveSegmentIndex(index);
+    setTransform(`translate3d(${targetWidth * index}px, 0, 0)`);
+    return onClick ? onClick(segment) : null;
+  };
+
+  return (
+    <SegmentsContainer>
+      <Indicator width={segmentWidth} transform={transform} ref={targetRef}>
```
michaeljaltamirano, line 55: Is there a way to consolidate `Indicator` within the `segmentItems` map function to prevent adding an additional button to the DOM? It looks like there's always one more than there exists `segmentItems`:

Screen Shot 2020-10-20 at 7 53 45 PM

Also, since Indicator precedes all the segment items, the current keyboard navigation behave is to immediately tab to the currently active segment. Is this intended? E.g., if Option 2 of 3 options is active, tabbing will go to 2, then to 1, then to 3. My expectation is that it would go in order.

michaeljaltamirano (changes requested):
I'm currently opening issues for UI toolkit / implementation discrepancies and our `secondaryTheme` was added after this PR was initially reviewed by folks so it looks like we will need to support the spec for the `secondaryTheme` version of this component now:

Screen Shot 2021-02-11 at 12 56 22 PM

heyamykate:
@michaeljaltamirano sorry I sort of forgot about this 😬 . I can work on this over cool-down 👍

winstonkim:
hi @michaeljaltamirano, I took a stab at making the styling updates for the secondary theme. let me know what you think!

also is there something specific i need to do so that the chromatic test passes?

michaeljaltamirano:
> hi @michaeljaltamirano, I took a stab at making the styling updates for the secondary theme. let me know what you think!
> 
> also is there something specific i need to do so that the chromatic test passes?

The initial build will fail if there's new visual regression tests that haven't yet been vetted as the baseline, I went ahead and approved the changes and sent you an invite to join as a collaborator on Chromatic. I'll review this PR later--thanks for carrying the torch!

michaeljaltamirano (approved):
@winstonkim Approving to unblock your work. 

I noted a [usability TODO in my Refinements PR](https://github.com/curology/radiance-ui/pull/1177#:~:text=selected%20focus%20state.-,TODO%20for%20future%20PRs,-The%20current%20functionality), but it is not blocking for getting this merged in, releasing a new minor version, and unblocking your work. We can continue to refine it later. 

Once the dust settles on what the DOM/UX looks/feels like, we should have it properly documented in Figma, [akin to Ben's comment with some focus state designs](https://github.com/curology/radiance-ui/pull/405#discussion_r496193333), but official, based on whatever those future refinements end up being. We'll want to loop in design once we get a better sense of what is feasible. 

However, as the TODO mentions, it might also be worth working with Design to determine whether or not this control is better represented as separate components: a [TabList](https://react-spectrum.adobe.com/react-aria/useTabList.html) which would seem to operate like how we seem to be needing it for some new Site pages being worked on (assuming I looked at the correct LP), and/or a [RadioGroup](https://react-spectrum.adobe.com/react-aria/useRadioGroup.html) if the selections correspond to some form submissions. If they're purely presentational then I think a `TabList` may be preferable to our current `button` implementation.

PR #409, opened 2020-10-01
Add Chromatic Integration

michaeljaltamirano:
### What and Why

This PR adds a [Chromatic integration](https://www.chromatic.com/) to our Storybook. If you would like to be added as a collaborator to this project, please DM me for the appropriate link. **Once this is merged in I will take steps to have individuals added, accordingly.**

It expands on our existing `circle.yml` file by adding a `yarn build` step and a `Publish to Chromatic` step that publishes the stories to a Chromatic CDN, and also reports any regressions if there are changes in our stories:

Screen Shot 2020-10-01 at 2 38 52 PM

This will allow us to report visual regressions in our stories when making component changes. Below is an example not of a component but changing documentation:

image

**I've linted the PR with relevant comments to make review easier, but the following is a summary of changes made across the story files:**

1. Moves the "Primary" (the first) story of all the component stories from the top of the docs page to be co-located with the Stories, which allows the title of the story to auto-populate (with less code).

Before:

Screen Shot 2020-10-22 at 1 33 34 PM

After:

Screen Shot 2020-10-22 at 1 33 52 PM

2. Ignores the "With Controls" (aka Knobs) stories from visual regression tests by using the following custom Chromatic syntax via `parameters`:

```tsx
WithControls.parameters = {
  chromatic: { disable: true },
};
```

3. Adds configuration to regression test at both **mobile** (`330px`) and **desktop** (`768px`) dimensions. The default is to test at Desktop, but the majority of web traffic we see is mobile, so the tests should reflect that.


nic-git (approved):

package.json
```diff
@@ -34,7 +34,8 @@
     "build:svgs:index": "create-index src/svgs/ --recursive --update --extensions svg",
     "publish-package": "run-s build && yarn publish",
     "heroku-postbuild": "yarn run build:icons && yarn run build:icons:index && yarn run build-storybook",
-    "tsc": "tsc"
+    "tsc": "tsc",
+    "chromatic": "chromatic --skip '@(master|dependabot/**)' --project-token"
```
michaeljaltamirano, line 38: We want to run chromatic on feature branches *prior* to being merged into master. When `master` rebuilds we do not want to trigger another build on chromatic. 

I've also set it up to ignore dependabot PRs for now. Once we settle on billing (which might be free, since the project is open source), we can adjust this so we can rely on Chromatic to help us sanity-check dependency updates as well. 

I previously [included the `--exit-on-upload` flag](https://github.com/curology/radiance-ui/pull/409/commits/f112b6e2e31e115c0a1211586fe2e5092d3064bb) here as well, but for whatever reason all subsequent commits after [changing the build step order in CI](https://github.com/curology/radiance-ui/pull/409/commits/fe8513d3c09b6f4e04be58d7950b7a5a9f2f1f4f) has meant we haven't triggered it, so for now we will have to rely on the `build` CI step itself failing if there are visual regressions as our check, instead of the Chromatic-specific reporter.

.storybook/preview.tsx
```diff
@@ -93,6 +94,12 @@ addParameters({
     options: {},
     manual: false,
   },
+  /**
+   * Defaults to smallest mobile and smallest desktop breakpoints for visual regression testing.
+   * Override on a per-story basis if component stories only need to test one breakpoint,
+   * typically small components that are the same on all views (e.g. Chip, Indicator)
+   */
+  chromatic: { viewports: [BREAKPOINTS.xs, BREAKPOINTS.md] },
```
michaeljaltamirano, line 102: I had originally considered three breakpoints, adding `xl` for a more typical desktop experience instead of `md` which is our bare minimum pixel width breakpoint, but I didn't think it added much value.

Some of our designs don't really respect the xs breakpoint (e.g. the breakpoint is `330px` width but some components have set widths like `327px`, before margin/padding) so the Chromatic screenshots might not be perfect, but it suffices in terms of testing.

.storybook/preview.tsx
```diff
@@ -71,7 +72,7 @@ addDecorator(InjectGlobalStyles);
 const ADDONS_REQUIRED_IN_OPTIONS = {
   isFullscreen: false,
   isToolshown: true,
-  panelPosition: 'right',
+  panelPosition: 'bottom',
```
michaeljaltamirano, line 75: Pretty sure this is the standard, and with all the addons this causes less crowding on the canvas.

stories/alert/index.stories.tsx
```diff
@@ -87,14 +87,18 @@ export const WithImage = () => (
...
+    <Alert duration="sticky" content="Alert.Container fixes my position" />
   </Alert.Container>
 );
 
 export default {
   title: 'Components/Alert',
   component: Alert,
   parameters: {
+    /**
+     * TODO: Figure out how to get Alerts to show up on Chromatic
```
michaeljaltamirano, line 99: I was unable to get Alerts to show up, despite my best efforts (https://www.chromatic.com/snapshot?appId=5f6e46f1e7081000221523cf&id=5f909030409e4d002292f8a0) and being able to solve ostensibly trickier components like modals. I think it's fine for now.

stories/accordion/subcomponents/index.stories.tsx
```diff
@@ -134,44 +134,75 @@ export const Truncate = () => {
...
-    <Accordion
-      onClick={() => undefined}
-      title={<Accordion.Content>Title (open)</Accordion.Content>}
-      isOpen
-    >
-      <Accordion.Content>Expansion with content padding</Accordion.Content>
-    </Accordion>
-  </Accordion.Container>
-);
+export const Content = () => {
```
michaeljaltamirano, line 162: The diff in this file is just making these two stories interactive.

stories/dialogModal/index.stories.tsx
```diff
@@ -99,21 +179,21 @@ export default {
           <ArgsTable />
           <Heading>Stories</Heading>
           <Description>
-            Source code for the stories can be found in the individual story
-            canvas tabs. Keeping the modals closed-by-default on the docs page
-            (unlike the open-by-default configuration on the canvas pages)
-            interferes with Storybook source inference.
+            The below stories are the not-open-by-default versions of the
```
michaeljaltamirano, line 182: Hopefully the description here (and in ImmersiveModal) adequately explains the duplication. The stories do not accept props easily, and so when trying to integrate the stories manually in the Docs page there were either: (1) Auto-open issues, or (2) The usage of the component directly outside of the Story broke styling. I decided this was the easiest way to work around the issue.

stories/accordion/subcomponents/index.stories.tsx
nic-git: useReducer 🙌

stories/button/linkButton/index.stories.tsx
```diff
@@ -89,6 +88,10 @@ export const WithControls = () => (
   </LinkButton.Container>
 );
 
+WithControls.parameters = {
+  chromatic: { disable: true },
```
nic-git, line 92: does disablement of the controls add on serve the purpose of making the snapshots cleaner, in that there are no additional UI elements on screen that could interfere with the chromatic checks?

michaeljaltamirano: parameters that include `{ chromatic: { disable: true } }` means we do not take a snapshot of it in Chromatic: https://www.chromatic.com/docs/ignoring-elements#ignore-stories

Essentially--with this LinkButton code as an example--the [LinkButton story](https://radiance-ui.curology.com/?path=/story/components-button-linkbutton--with-controls) is entirely duplicative of the default primary button captured in the [Default story](https://radiance-ui.curology.com/?path=/story/components-button-linkbutton--default). Since visual changes will require manual QA of the diff, I consider stories that are created solely to use add-ons (whether Knobs or Controls, which is what Storybook recommends going forward) not appropriate candidates for the visual regression testing, since they're pretty much just overhead without any precise visual distinction from the other stories. Since we had them before, though, I wanted to keep them in Storybook, just not include them in Chromatic.

Let me know if this doesn't answer your question! I still need to update the documentation in Notion once this is merged in, but I wanted to make sure it was ready to show (even if not *fully* complete, i.e. documentation is updated) in Demos and Discussions.

nic-git (approved):
Nice! can't wait to see this Visual Regression feature in action as part of the CI process.

stories/accordion/subcomponents/index.stories.tsx
michaeljaltamirano: Yeah I originally set it up when it was 3 separate accordions but kept it after reducing it to 2 because I'd like for our code to use more examples--we use a fair amount of coupled `useState` functions and it might be nice to get more comfortable with other hooks API.

stories/button/linkButton/index.stories.tsx
nic-git: Yup makes total sense - thank you

nic-git (approved):

package.json
```diff
@@ -124,7 +126,6 @@
     "rimraf": "^3.0.2",
     "rollup": "^2.26.11",
     "rollup-plugin-size-snapshot": "0.12.0",
-    "tinycolor2": "^1.4.1",
```
michaeljaltamirano, line 127: This is already in `dependencies`.

daigof (approved):
This is Great, I immediately will benefit and learn the process when I do the Components refactors to FC and other standarizations. got a couple of comments may be addressed later

stories/loadingSpinner/index.stories.tsx
```diff
@@ -53,6 +56,7 @@ export default {
   title: 'Components/LoadingSpinner',
   component: LoadingSpinner,
   parameters: {
+    chromatic: { disable: true, viewports: [BREAKPOINTS.xs] },
```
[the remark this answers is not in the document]
michaeljaltamirano: We're setting `{ disable: false }` on just one story but want to disable the others:

```
Default.parameters = {
  chromatic: { disable: false },
};
```

I can co-locate viewports on the specific story if it's clearer.

stories/callout/index.stories.tsx
```diff
@@ -45,6 +44,10 @@ export const WithoutContainer = () => (
   </Callout>
 );
 
+WithoutContainer.parameters = {
+  chromatic: { viewports: [BREAKPOINTS.xs, BREAKPOINTS.md] },
```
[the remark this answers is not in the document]
michaeljaltamirano: The `export default` has `chromatic: { viewports: [BREAKPOINTS.xs] },` so this overrides the story-level definition. I can just set `chromatic: { viewports: [BREAKPOINTS.xs] },` on every individual story instead if that would be clearer.

.storybook/preview.tsx
michaeljaltamirano: My preference is to respect the smallest breakpoint we define without creating new stand-alone constants. If something is compatible with `BREAKPOINTS.xs` it would mean support for iPhone 5, which some (albeit a small amount, 1-2%?) of our users have.

I'm aware that Design uses an iPhone X specification but I think it's a reasonable compromise to test at lower dimensions to see that we're not relying on too many hard-coded dimensions. I'm pretty sure the only thing that needs to change for support <375px width is just using `max-width` instead of `width`.

.circleci/config.yml
```diff
@@ -36,29 +41,42 @@ aliases:
                 fi
 
 jobs:
-    build:
+    build-js:
         <<: *defaults
-
         docker:
             - image: circleci/node:10-stretch
```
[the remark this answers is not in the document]
michaeljaltamirano: I think that's a good idea for something to do while we're waiting for specifications on dark mode (if we're out of theming things to work on) or as a cool down item later on. 

10 is maintenance mode, though, and 12 will be in a few months. My preference would just be to go all the way up to 14, barring technical difficulties: https://nodejs.org/en/about/releases/

michaeljaltamirano:
Merging to get this config into mater and see how it behaves throughout today--I am happy to make any follow-up changes, I just request that you @ me explicitly so I see it with visibility.

.circleci/config.yml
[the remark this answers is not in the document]
daigof: forcing to 12 or 14 is not a good idea, I tried that in Daylight and it also forces your consumer app to be in like 14 for example. My suggestion was to all apps be in the same engine requirement `"node": ">=10.0"`, will try this later

PR #416, opened 2020-10-05
[Storybook] Replace storiesOf API with Component Story Format + MDX

michaeljaltamirano:
### What and Why

This PR overhauls our story setup for dual purposes:

1. The primary purpose is to take advantage of Chromatic visual regression testing (WIP PR: #409). 
    - Currently, our stories consistent of one large README document with some components interspersed. Our visual regression testing setup would be most precise by conforming to the Storybook convention of having separate stateful components as individual stories.
        - **For reviewing the PR: Comments on the way in which the stories are split up would be helpful feedback!** In some cases--like the `Chip` story--I opted to put all usage on one story for simplicity. For others--like the `Accordion`--they seemed significant enough that I put them in separate stories. This initial PR need not set these components up perfectly, but if anything seems glaringly sub-optimal I would appreciate the call-out. 
1. The secondary purpose is to improve the manner in which we document components.
    - Currently, we write our components in a `src` folder, the stories in a `stories` folder, and the markdown file with which we document our component usage in `docs`. This PR consolidates `docs` information into `stories` in one file, in the `stories` directory. 
        - In most cases we were able to achieve this with plain TypeScript files, using [Storybook's Component Story Format (CSF), "the recommended way to write stories"](https://storybook.js.org/docs/react/api/csf).
            - In order to provide as much detail as our existing stories, this required some special markup in the story files by importing `@storybook/addon-docs/blocks` components and constructing the Docs page manually. I think it is a fair trade-off but does add some overhead to how to write stories.
                 - Relying on defaults means we could not include our `Usage` section that shows how to import it. If we decide to fashion our stories differently, we could simplify the majority of our stories, even though a small handful still require the manual docs construction since we pass in custom props to have the Docs components behave differently than the story components. **This mostly applies to modals and tooltips, which we want to visually regression test in the exposed `open` form.** 
        - In some cases, though, the Documentation was so extensive that it was preferable to make use of [MDX](https://mdxjs.com/).
            - Those files are almost more markdown than JavaScript, like `Accordion` and `Icon`. **I'm open to figuring out how we setup a rule of thumb for which to use.**

Ignoring the `yarn.lock` diff, these changes are nearly a net-negative 1,000 line change. 🌈 

### Outcome

Here is an example of how the Banner story is changing.

#### Before:

1. The `Usage` markdown is a hard-coded codeblock within `jsx` template tags. As such, it's divorced from actual usage, and so if the API changes the developer has to remember to also change the markdown documentation.
    - I've encountered more than a few examples of incomplete/incorrect code while converting these files over.
1. All the different states are part of the same large story: I cannot link someone to a clickable banner story, directly: they have to look for it on the page. There's no helpful sub-page navigation links. 
1. We manually add the `prop-types`, which suffers from the same problem as the markdown code: it requires adjustment when the API changes. 

image

#### After:

1. The only hard-coded value is the import Usage. 
    - The `prop-types` are automatically generated from the component. 
1. The Banner description text is pulled from a JSDoc comment on the component definition to automatically populate the docs. 
    - In the `Before` example you can see the `Notes` section at the bottom of the markdown file. Developers should see those comments while looking at the source, too, so that they don't have to pull up a separate page during development.
1. Each separate `Canvas` with a `show code` block is a separate story, but the `Docs` page helpfully combines everything into one file. 
    - This version no longer suffers from the hard-coded markdown example code problem that the other screenshot had. That code is the rendered component. 

localhost_9001__path=_docs_components-banner--default (5)
localhost_9001__path=_docs_components-banner--default (6)

### Summary of Intent

1. Establish more granular stories paradigm to make use of Chromatic visual regression testing (#409). 
1. Removes standalone `docs/` directory: all documentation should be added via JSDoc comments in `src/` components, or directly in `stories/` (via `<Description>` helper components) if required. 
1. Preserve current standard of documentation: this means keeping all knobs stories, etc., unless the examples were essentially non-functional.
1. Use most up-to-date Storybook best practices to improve the Developer Experience of making changes in `radiance-ui`. 

**My recommendation for reviewing this PR would be to compare the version of this storybook (https://curology-radiance-pr-416.herokuapp.com) with http://radiance-ui.curology.com/ before looking at the code,** to confirm that the final product is acceptable from a documentation perspective before focusing on the code changes that create it. 

### Future Changes

1. Replace `knobs` with new `controls` functionality: https://storybook.js.org/docs/react/essentials/controls


package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "test": "jest",
     "storybook": "run-s build:svgs:index storybook:dev",
-    "storybook:dev": "start-storybook --port 9001 --config-dir .storybook --static-dir ./public",
+    "storybook:dev": "start-storybook --docs --port 9001 --config-dir .storybook --static-dir ./public",
```
michaeljaltamirano, line 20: `--docs` starts storybook in `Documentation Mode`, which means that when you load a Storybook it defaults to opening the `Docs` tab of the first story it opens instead of the `Canvas` tab. Since this PR moves the Documentation info from the Story itself to the official `Docs` page, this effectively maintains the current behavior (otherwise the Storybook would just load a single story of a stateful component.)

src/shared-components/accordion/index.tsx
```diff
@@ -27,7 +27,12 @@ type AccordionProps = {
   title: React.ReactNode;
 };
 
-class Accordion extends React.Component<
+/**
+ * A list of items that allows each item's content to be expanded and collapsed by clicking its title bar.
+ *
+ * The accordion component expands to reveal hidden information. They should be used when you need to fit a large amount of content but don't want to visually overwhelm the user.
+ */
+export class Accordion extends React.Component<
```
michaeljaltamirano, line 35: One detail I noticed while working on this is that, for TypeScript files in particular, only Named Exports allowed us to actually document the type/prop information within a given story. Because of this, this PR **overhauls our existing import/export structure for components, and prefers named exports in almost every instance**. The sole exception is sub-component usage that we do not document (e.g. the `Arrow` component for an `Accordion`) or exports that make use of `withDeprecationWarning` like `Button` and `RoundButton`. 

These changes are reflected in `src/shared-components/index.ts`, and I updated the `.size-snapshot.json` file to show that the effect on our bundle is neglible. The majority of the size changes comes from shipping the expanded JSDoc comments, which wouldn't matter when passing it through a minifier that removes comments anyway. Additionally, this PR also removes `defaultProps` in favor of ES6 defaults for function components, so some components actually shrink. Those changes may conflict with #423, but since this PR probably won't get merged until sometime next week, I will just deal with the merge conflicts that result from those changes (if any).

src/shared-components/avatar/__snapshots__/test.tsx.snap
```diff
@@ -12,7 +12,6 @@ exports[`<Avatar /> UI snapshot render with size large styles 1`] = `
 <img
   alt="avatar"
   className="emotion-0 emotion-1"
-  size="large"
```
michaeljaltamirano, line 15: `Avatar` was one of the few components that has a somewhat-breaking change: previously we just used a `styled.img` export directly, but this PR turns it into a react component that receives props and then passes it onto a `styled.img` component. The difference is that the Story can receive type enforcement. I also renamed the prop we pass to the `styled.img` component `avatarSize` so that we wouldn't also pass it to the DOM--this pattern (see `inline` in SVG components) is one of our biggest offenders when it comes to browser console warnings/errors.

src/shared-components/checkbox/index.js
```diff
@@ -1,45 +1,47 @@
...
-  disabled,
-  onClick,
-  children,
-  type,
-  icon,
-  size,
+  children = null,
+  disabled = false,
+  icon = undefined,
+  onClick = undefined,
```
michaeljaltamirano, line 35: Some files not yet converted to JS require this because TS picks up that we're typing this as a function, but if we do not say it is undefined this way there is a TS error about the prop not being provided. Confusing, but #423 will remove this type of error from `radiance-ui` anyway, so it is temporary.

src/shared-components/container/style.ts
```diff
@@ -24,10 +24,6 @@ const messageStyle = css`
 
 export type ContainerType = 'clickable' | 'message';
 
-/**
- * TODO: Determine if we should return a style by default
- * instead of no style at all.
- */
 const getContainerTypeStyles = (type?: ContainerType) => {
```
michaeljaltamirano, line 27: After looking at the story it is clear we support containers without a type, even if it isn't standard in our usage.

src/shared-components/avatar/style.ts
```diff
@@ -10,21 +9,21 @@ const SIZES = {
...
+const AvatarImage = styled.img`
   ${determineSize}
   background-color: ${COLORS.secondary};
   border-radius: 50%;
   overflow: hidden;
 `;
 
-Avatar.propTypes = {
-  size: PropTypes.oneOf(['small', 'medium', 'large']),
+export default {
```
michaeljaltamirano, line 27: While making changes, decided to conform to the un-named default export convention we're currently exploring.

src/shared-components/button/components/linkButton/index.tsx
```diff
@@ -22,27 +22,40 @@ const propTypes = {
...
-
 type LinkProps = {
+  /**
+   * Specifies the tag or element to be rendered
+   */
   as?: 'a' | React.ElementType;
+  /**
+   * TODO-TS: Limit type from string to COLORS constants options
+   */
   buttonColor?: string;
```
michaeljaltamirano, line 33: The diff is already very large, so I largely kept the API of the `src` files the same.

stories/accordion/index.stories.mdx
```diff
@@ -0,0 +1,141 @@
...
+import { TruncatedAccordion } from './truncatedAccordion.stories';
+import { DisabledAccordion } from './disabledAccordion.stories';
+import { NoBorderAccordion } from './noBorderAccordion.stories';
+import { RightAlignedArrowAccordion } from './rightAlignedArrowAccordion.stories';
+import { ThumbnailsAccordion } from './thumbnailsAccordion.stories';
+import { useAccordionState } from './utils';
+import accountImage from 'shared/account.png';
+import personImage from 'shared/person.jpg';
+
+import * as AccordionStories from './index.stories.tsx';
```
michaeljaltamirano, line 23: For more context on this convention: https://github.com/storybookjs/storybook/pull/11752

stories/button/index.stories.tsx
```diff
@@ -108,33 +109,46 @@ import { COLORS } from 'radiance-ui/lib/constants';
...
+      onClick={action('button clicked')}
+      textColor={text('textColor', '')}
+    >
+      {text('children', 'Click me!')}
+    </Button>
+  </Button.Container>
+);
+
+export default {
+  title: 'Components/Button/Button',
```
michaeljaltamirano, line 134: In order for us to nest buttons the way we currently do, we needed to add a base `Components` root node, since otherwise the `Buttons` would be separated from everything else. 

Since this was required, I also decided to split up the `constants` and the `icons` usages, since they're not `components`. 

Before:

Screen Shot 2020-10-11 at 8 26 42 PM

After (not a 1:1 screenshot, since it takes up more space now):

Screen Shot 2020-10-11 at 8 30 40 PM

src/shared-components/checkbox/index.js
daigof: Im gonna merge my final TS convertion PR that covers alll components converted to TS then solve all merge conflicts for this PR so Reviewers can ignore components in JS for now if they want

nic-git:
**In general:**

- The more granular format works well, especially with the addition of the "show code" control.
- The high-level layout choice for each component made sense to me. 
- The closer coupling of code artifacts to documentation adds value and saves effort over the decoupled stand-alone docs.

**One advantage of the previous style?**

I like that the previous documentation form had a standard "Usage" section for each component. It appeared that the documentation had greater uniformity. Not the primary concern of this PR, but agreeing to have a standard "Basic Usage" story per component appeals to me.

**Some notes when reviewing**

Accordion
when selecting a variation such as "No Border Accordion" I am automatically scrolled to the top edge of the example component, but the title for that component example is off screen (above). This seems to be the case for all the Accordion stories.

DialogModal - With Close Icon
the modal shown is missing the close icon

ImmersiveModal
the same modal is shown for all stories, for example, "With Header Image" example shows the vanilla version of the modal and not the variation with the header image.

Tooltip
The jump links don't work, i.e. clicking on Small Tooltip has no effect on the Scroll Position of the page (you are not moved to the Small Tooltip section and will have to manually scroll).

src/shared-components/avatar/style.ts
nic-git: Is the un-named default export intended to more closely match the fact a default export allows the importing code site to use whichever name for the imported entity as the author wishes?

In other words if I can rename the entity during import why is a name specified in the export itself?

nic-git: Do you have any links or info about "un-named default export" that I could read?

nic-git (commented):
This looked like a big effort 🙏

I left a general comment outside of the review noting some observations when clicking through the heroku deploy. 

I would approve when you are ready and other feedback is in.

daigof (approved):
ok this must have been a huge effort, to put all this in one PR so congrats.
 

**My suggestion would be to to merge this asap** so we avoid any merge conflict that will occur if any work is done in Radiance. Having said that there are couple of items I would like to record here:


- I solved the merge conflict on those components that were in JS, I think I got them right but in case these were the components for you to take a final look: Typography,Tooltip,Tabs,Transition,SelectorButton/Checkbox/RadioButton
- Both Modals open as soon as you open the canvas, that may be confusing Nic if you click the trigger it actually display the Modal with Icons,HeaderImage, etc. 
- There are example code that were lost, there is no way for a developer to copy for example Tooltip, we need to tackle that in future PR


Thanks for putting this together, this is solid base and I think we would benefit from merging this as is and iterate over in future PR.

src/shared-components/avatar/style.ts
michaeljaltamirano: @nic-git https://www.notion.so/curology/styles-ts-5d6c113fa66240829be7d9e7f8eacea0. You can also see the Slack room #mtg-eng-demos-styles-followup for the follow-up discussion.

nic-git:
@daigof - for the modals, I re-checked and the canvas is working great for me, but the docs tab shows the standard modal (and not the modal variant, like a modal with an image header) after clicking one of the variants in the sub-page links. I don't think this is a big deal though and could be intentional.

michaeljaltamirano:
> **One advantage of the previous style?**
> 
> I like that the previous documentation form had a standard "Usage" section for each component. It appeared that the documentation had greater uniformity. Not the primary concern of this PR, but agreeing to have a standard "Basic Usage" story per component appeals to me.

@nic-git I think we should introduce a new convention of looking at the Docs tab for usage details. Having a separate Usage tab would be a non-functional story that we would also then need to omit from the visual regression testing, since Chromatic invests all of the individual component stories. The “Show Code” functionality of the actual components in the Docs is an improvement over Markdown code that has no guarantee of being correct.

I think a helpful way to address this comment would be if you could elaborate on what is meant by “uniformity”. Most of the stories still have the same Docs structure, so there should be uniformity, but if they appear to be missing something that the current Storybook has, we should address that.

michaeljaltamirano:
> Accordion
> when selecting a variation such as "No Border Accordion" I am automatically scrolled to the top edge of the example component, but the title for that component example is off screen (above). This seems to be the case for all the Accordion stories.

Since this story uses MDX very heavily, I think we might be able to fix this by wrapping our built-in documentation details with a `Header` component imported from storybook/addon-docs/blocks. Good catch.

> DialogModal - With Close Icon
> the modal shown is missing the close icon
> 
> ImmersiveModal
> the same modal is shown for all stories, for example, "With Header Image" example shows the vanilla version of the modal and not the variation with the header image.

These might’ve been experienced if you were clicking the different components while keeping the Docs page open. On the Docs page the normal dialog modal automatically opens up (in order to preserve the source on the Docs page). ImmersiveModal I was admittedly lazy and all three automatically open up, but only one ever shows. We might want to sacrifice the source availability on the Docs page (like we do for Tooltip) in order to improve the UX here.

> Tooltip
> The jump links don't work, i.e. clicking on Small Tooltip has no effect on the Scroll Position of the page (you are not moved to the Small Tooltip section and will have to manually scroll).

I covered this in my other comment but i recommend looking at the source code that creates this Docs page to better understand why this doesn’t work. There might also just be something wrong there but I wanted to address this again!

ZeMunchkin (approved):
What a tremendous and awesome change! I'm a big fan of consolidating / automating the documentation. Huge thank you for taking this on 🙌 

One thing that might be nice to add in the new docs (if possible) is a subheader before the props chart so it stands out and is clear what it is. Not sure of the complexity to do that, so feel free to skip if it's difficult to implement. 

In the following components' docs, the `show code` option is sometimes not available. For more basic components that's less of an issue, but for more complex ones like Accordion, being able to see the source code is pretty helpful:
- Accordion
- Dialog Modal
- Immersive Modal
- Tooltip

IMO these aren't blocking issues, so approving in the interest of getting this merged in and avoiding conflicts. 

Really awesome job on this!

michaeljaltamirano:
Updated according to most (if not all) of the review comments: 

Added a Props heading to the component prop type information for visual clarity in https://github.com/curology/radiance-ui/pull/416/commits/754b465a9519e831f3de402c4d2d5d64dc163097.

Fixed the Accordion anchoring/source code availability in https://github.com/curology/radiance-ui/pull/416/commits/31a27ad57465e52851708337b56b734f1b680075.

Refined DialogModal auto-open behavior on Docs page in https://github.com/curology/radiance-ui/pull/416/commits/115377a5846aea411d72c4ef98905fd68c40b032.

Refined ImmersiveModal auto-open behavior on Docs page in https://github.com/curology/radiance-ui/pull/416/commits/7c371bcd332acad372a03c5092145917a80f5ca6.

For both DialogModal and ImmersiveModal there is a more explicit disclaimer about pulling the source code from the Canvas stories.

Decided to have the Tooltip stories render two components: an always-open tooltip and an interactive tooltip, in order to get around the issues working around having the Canvas and Docs pages behave differently cause: https://github.com/curology/radiance-ui/pull/416/commits/b48cf320e6e3beb5bf139408944c981e3fb62251

.storybook/manager.ts
```diff
@@ -1,2 +0,0 @@
-import '@storybook/addon-viewport/register';
```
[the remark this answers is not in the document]
michaeljaltamirano: Deleted! Thanks for catching this.

michaeljaltamirano:
> > **One advantage of the previous style?**
> > I like that the previous documentation form had a standard "Usage" section for each component. It appeared that the documentation had greater uniformity. Not the primary concern of this PR, but agreeing to have a standard "Basic Usage" story per component appeals to me.
> 
> @nic-git I think we should introduce a new convention of looking at the Docs tab for usage details. Having a separate Usage tab would be a non-functional story that we would also then need to omit from the visual regression testing, since Chromatic invests all of the individual component stories. The “Show Code” functionality of the actual components in the Docs is an improvement over Markdown code that has no guarantee of being correct.
> 
> I think a helpful way to address this comment would be if you could elaborate on what is meant by “uniformity”. Most of the stories still have the same Docs structure, so there should be uniformity, but if they appear to be missing something that the current Storybook has, we should address that.

Merging in with this being the only remaining open question, I believe.

PR #421, opened 2020-10-07
[TS Conversion] - Field and VerificationMessages

daigof:
Basically title, added commnent in lines for context


michaeljaltamirano:
@daigof Just merged in the Storybook 5 --> 6 branch, let me know if any of the conflicts are confusing!

src/shared-components/field/__snapshots__/test.tsx.snap
```diff
@@ -114,15 +289,12 @@ exports[`<Field /> UI Snapshot renders with label and labelFor 1`] = `
       className="emotion-2 emotion-3"
       disabled={false}
     />
-    <ul
-      className="emotion-4 emotion-5"
-    />
   </div>
```
daigof, line 292: <ul> is remove cus we return null from VerificationMessage now if message is empty {}

src/shared-components/field/index.tsx
```diff
@@ -43,13 +59,13 @@ class Field extends React.Component {
...
     const {
       children: inputChild,
-      disabled,
-      messages,
-      messagesType,
-      hideMessagesIcon,
-      hintMessage,
-      label,
-      labelFor,
+      disabled = false,
```
daigof, line 62: I think this is how we can handle default Values for Class components. We may need to agree on a convention and standarize for all components. The end goal I think it is to mark the prop as optional if it's optional for future usage in curology

src/shared-components/field/test.tsx
```diff
@@ -32,35 +32,59 @@ describe('<Field />', () => {
...
+        <Field messages={messages} messagesType="success">
           <Field.Input />
         </Field>,
       );
+
       const li = wrapper.find(MessageItem);
 
       expect(li).toHaveLength(1);
-      expect(li.text().match(messages.maxLength)).toBeTruthy();
+      expect(li.text().match(messages.success)).toBeTruthy();
```
daigof, line 87: Ts catch. I honestly dont know how this was working along with type/messagesType above

src/shared-components/verificationMessages/formatMessage/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  if (Array.isArray(message)) {
+    const mapResult = message.map((value) => (
+      <React.Fragment key={uniqueid()}>{value}</React.Fragment>
+    ));
+
+    return (
+      <React.Fragment>
+        {mapResult.reduce((prev, curr) => (
+          <React.Fragment>
+            {prev}
```
daigof, line 16: the old syntax had TS errors, I tested this in the Story and works also added a test.

src/utils/helperTransition/index.tsx
```diff
@@ -38,16 +40,16 @@ const getStyleForTransitionState = (transitionState: string): object => {
 
 type HelperTransitionParamsType = {
   children: JSX.Element;
-  props: Array<any>;
+  [key: string]: any;
```
daigof, line 43: how we handle  spread   ...props

src/utils/helperTransition/index.tsx
```diff
@@ -38,16 +40,16 @@ const getStyleForTransitionState = (transitionState: string): object => {
...
 };
 
 const HelperTransition = ({
   children: child,
   ...props
-}: HelperTransitionParamsType): React.ReactNode => (
+}: HelperTransitionParamsType) => (
```
daigof, line 49: actually removing return types and let TS infer them solved a bunch of TS errors

src/shared-components/field/test.tsx
michaeljaltamirano: It's still not really working: the return value is an array, which is truthy.

You can change these two tests to the following: 

`expect(li.text()).toEqual(messages.maxLength);`

and

`expect(li.text()).toEqual(messages.success);`.

src/shared-components/field/index.tsx
michaeljaltamirano: Since this is just a class component that only uses render, should we just go ahead and convert it to a function component? We can still apply "static" properties to it when the component is just a function, too. It shouldn't break functionality.

As far as a convention: that would work for simple components that only destructure once, but what if a class has 15 methods on it? We'd probably want to define something like:

```
getProps = () => {
  const {
    somePropValue = true,
  } = this.props;

  return {
    somePropValue,
  }
}

render() {
  const { somePropValue } = this.getProps();
}
```

You could also look at Matt's `get injectedProps()` function in our main app for inspiration on how to work around this.

src/utils/helperTransition/index.tsx
```diff
@@ -3,7 +3,9 @@ import { Transition } from 'react-transition-group';
 
 import { ANIMATION } from '../../constants';
 
-const getStyleForTransitionState = (transitionState: string): object => {
+const getStyleForTransitionState = (
+  transitionState: string,
+): Record<string, unknown> => {
```
michaeljaltamirano, line 8: This is another one where we can remove the type definition of the return value--it infers all of the possible returns:

Screen Shot 2020-10-08 at 8 19 46 AM

src/shared-components/verificationMessages/index.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { TransitionGroup } from 'react-transition-group';
+
+import { CenteredMessageList, MessageList, MessageItem } from './style';
+import HelperTransition from '../../utils/helperTransition';
+import formatMessage from './formatMessage';
+
+export type messagesTypes = 'error' | 'success';
```
michaeljaltamirano, line 9: [Nit] We should rename these MessageTypes and MessageType so it's clear just from looking at the name that these are types and not actual values.

src/shared-components/field/index.tsx
daigof: Field refactored to FC.
ok so I have this idea when all components are converted, go to each one and standardise some things like defaultProps and convert to FC all than can be converted. I wanted to get an example of this convention so I applied it to `DialogModal` it worked fine and it even has spread props so it's a complete example. Can u review DialogModal so we have it as model for future work

src/shared-components/field/style.ts
```diff
@@ -17,7 +20,7 @@ export const FieldContainer = styled.div`
   margin-bottom: 22px;
 `;
 
-export const Label = styled.label`
+export const Label = styled.label<{ disabled?: boolean }>`
```
michaeljaltamirano, line 23: There might've been a typing error before, but since we're relying on default props in a function component now this prop is always being passed to Label

src/shared-components/field/test.tsx
daigof: Fixed, yeah I havent paying attention of whats going on with some code just convert them to TS

src/utils/helperTransition/index.tsx
daigof: yep removed thanks

src/shared-components/verificationMessages/index.tsx
daigof: fixed

src/shared-components/field/style.ts
daigof: fixed, yeah this is another one of the things I want to review and standarize after all is converted

src/shared-components/verificationMessages/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+  );
+};
+
+VerificationMessages.propTypes = {
+  centered: PropTypes.bool,
+  messages: PropTypes.objectOf(PropTypes.node),
+  type: PropTypes.oneOf(['error', 'success']),
+};
+
+VerificationMessages.defaultProps = {
```
michaeljaltamirano, line 58: We can remove the `defaultProps` since we have the ES6 defaults present. (Though it looks like we need to update `messages` in the function definition.)

src/shared-components/field/index.tsx
michaeljaltamirano: What you committed works fine. I honestly don't have super strong preferences, so I think it's better to survey the team and see (especially amongst those who have a harder time with TypeScript) what they might prefer. 

It seems like in addition to handling it via ES6 defaults in a separate method, we could also just define a separate TypeScript type, and cast any class components w/ `defaultProps` defined with that type, e.g.

```tsx

type DialogModalProps = {
  children: React.ReactNode;
  onCloseIconClick?: () => void | null;
  title?: string;
};


/**
 * Equivalent to:
 * type DialogModalPropsWithDefaults = {
 *   children: React.ReactNode;
 *   onCloseIconClick: () => void | null;
 *   title: string;
 * }
 */
type DialogModalPropsWithDefaults = Required<DialogModalProps>;

// We could also just inline `Required<DialogModalProps>` but some folks aren't familiar with the built-in utility methods.
getProps = () => this.props as DialogModalPropsWithDefaults;
```

Right now we have both the ES6 defaults *and* the `defaultProps` defined, which some people might find very confusing. 

I also think getting folks comfortable converting class components over to function components is key. The large majority of our lifecycle method usage is so simple that the mental model of how they map over to `useEffect` is *usually* straightforward. But some cases are not. In the case of `dialogModal`, it's a pretty simple conversion, and I think 99 times out of 100 it's better to just do the work to convert something over, rather than debate around that work, but that speaks more to my comfort with doing that work (especially in `radiance-ui`, with very sufficient test coverage) than the actual practicality of that ratio 😅 

Anyway, I encourage you to take the lead on this--perhaps draft up an RFC and share with the team?

src/shared-components/verificationMessages/test.tsx
```diff
@@ -18,12 +18,37 @@ describe('<VerificationMessages />', () => {
...
       };
+
       const component = renderer.create(
         <VerificationMessages messages={messages} />,
       );
 
+      const tree = component.toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+    it('renders with custom props', () => {
```
michaeljaltamirano, line 39: What does `custom props` refer to?

src/shared-components/field/test.tsx
```diff
@@ -32,35 +32,59 @@ describe('<Field />', () => {
       const tree = component.toJSON();
       expect(tree).toMatchSnapshot();
     });
+
+    it('renders with custom props', () => {
```
michaeljaltamirano, line 36: `renders with an error message`?

michaeljaltamirano (approved):
Left a couple questions about test name descriptiveness, but otherwise LGTM! As always, thanks for your hard work in converting a bunch of these components over.

One final request: could you merge in `master`, install all the latest update-to-date dependencies, run `yarn build`, and then commit the `.size-snapshot.json` diff? Still need to figure out a way to automate this but any time we're modifying `src` files it's helpful to trick if/how things are changing, even minor changes. For example, converting Field to a function component? 16% reduction in component (gzipped) size 💪

PR #423, opened 2020-10-09
[TS conversion] - Final components

daigof:
- Typography
- Tooltip (converted to FC)
- Tabs
- Transitions
- SelectorButton/RadioButton/Checkbox
- Added some tests
- Fixed/Refactor some things: tabIndex as number / added disabled support for Radio / null instead of undefined for consistency for Nodes


src/shared-components/tooltip/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const open = defaultOpen || clicked || hovered;
+
+  return (
+    <OffClickWrapper onOffClick={closeTooltip}>
+      <Global
+        styles={{
+          'body.cursor-pointer': {
+            cursor: 'pointer',
+          },
```
youngmanalive, line 84: i never knew we added cursor pointer to the whole body 🤔

youngmanalive (approved):
💯 lgtm!

src/shared-components/tooltip/index.tsx
daigof: yeah this is old logic, what happens is that if u click the trigger element the tooltip stays opened until u click somewhere else so the cursor pointer maybe symbolise that UX that u need click somewhere in the page for the tooltip to go away. Also I think when I reworked the tooltip we were just adding the class and relying on the app to add that class, I think I added the GlobalStyle here so it decouple this logic from the app

PR #437, opened 2020-10-22
[Hotfixes] -  Verification messages animation and modal

daigof:
- Modal: this is pending fixes explained here: https://app.asana.com/0/1173345595414308/1197707937611097/f   
- VerificationMessages: While updating Radiance in Curology, I was checking the snapshot differences and something didnt look right. So I tested the email field in the login page comparing side y side with Production (old radiance) and noticed the initial error animation was lost. Bug was introduced here: https://github.com/curology/radiance-ui/pull/421/files#diff-e9498b132d50a72d80dfb59c0cc8e675150eaeb28909ea7819c319d3f10f8756. If no error messages I returned null but we need the Transition structure present for the inital animation to work. So this fix revert back that functionality to the way it was in the old JS file.

BUG no animation
Screen Recording 2020-10-22 at 16 28 02

FIX animation restored
Screen Recording 2020-10-22 at 16 30 00

### DIFF
--- a/src/shared-components/verificationMessages/index.tsx
+++ b/src/shared-components/verificationMessages/index.tsx
@@ -59,7 +59,7 @@ export const VerificationMessages = ({
             </HelperTransition>
           ))
       ) : (
-        <li />
+        <React.Fragment />
       )}
     </TransitionGroup>
   );
--- a/src/shared-components/verificationMessages/index.tsx
+++ b/src/shared-components/verificationMessages/index.tsx
@@ -38,28 +38,29 @@ export const VerificationMessages = ({
   type = 'error',
 }: VerificationMessagesProps) => {
   const messageKeys = Object.keys(messages);
-
-  if (messageKeys.length === 0) {
-    return null;
-  }
+  const showMessages = messageKeys.length > 0;
 
   return (
     <TransitionGroup component={centered ? CenteredMessageList : MessageList}>
-      {messageKeys
-        .filter((key) => {
-          const message = messages[key];
-          if (!Array.isArray(message)) {
-            return true;
-          }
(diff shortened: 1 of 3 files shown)


src/shared-components/verificationMessages/index.tsx
```diff
@@ -38,28 +38,29 @@ export const VerificationMessages = ({
...
+          })
+          .map((key) => (
+            <HelperTransition key={key}>
+              <MessageItem type={type}>
+                {formatMessage(messages[key])}
+              </MessageItem>
+            </HelperTransition>
+          ))
+      ) : (
+        <li />
```
michaeljaltamirano, line 62: Can we return `null` here instead of an empty list item? It looks like we were returning `false` before, so I imagine it should work.

michaeljaltamirano (approved):
Left one comment that would also require updating the snapshots, but otherwise LGTM

src/shared-components/verificationMessages/index.tsx
daigof: tried false and null but TS yells is not compatible with children type which is Element[]

daigof: tested and it does prints the ul with one empty li and height of 0. I guess is safe, otherwise we could jsut add the ts-ignore and return false like before

michaeljaltamirano: What about `<React.Fragment />`? I just pulled that in without errors.

daigof: yeah I forgot.  Fragment looks better I compared to Prod login screen and they both prints empty ul,.
 Changed, updated snapshot and merging soon

PR #451, opened 2020-10-23
[Components Standarization] Part 1

daigof:
- Lock emotion packages versions
- Convert classes to FC
- Follow the Standarization defined here : https://www.notion.so/curology/Radiance-Conventions-for-future-cleanup-PRs-a2a526b6f11d4efd852c562387bc0ae2

- Part 2 will tackle the rest


package.json
```diff
@@ -131,8 +131,8 @@
     "yarn-run-all": "^3.1.1"
   },
   "dependencies": {
-    "@emotion/core": "^10.0.35",
-    "@emotion/styled": "^10.0.27",
+    "@emotion/core": "10.0.35",
```
michaeljaltamirano, line 134: Should we peg `@emotion/babel-preset-css-prop` too?

src/shared-components/accordion/index.tsx
```diff
@@ -32,131 +32,97 @@ type AccordionProps = {
...
+  onClick,
+  rightAlignArrow = false,
+  title,
+}: AccordionProps) => {
+  const [contentHeight, setContentHeight] = useState('0px');
+
+  const contentRef = React.createRef<HTMLDivElement>();
+
+  useEffect(() => {
+    const nextHeight = `${
```
michaeljaltamirano, line 50: It's less DRY but I think we should just make this more readable instead of continuing to use what was present before:

```tsx
const nextHeight = isOpen && contentRef.current ? `${contentRef.current.clientHeight}px` : '0px'
```

src/shared-components/accordion/index.tsx
```diff
@@ -32,131 +32,97 @@ type AccordionProps = {
...
+Accordion.Container = Container;
+
+Accordion.Content = Content;
+
+Accordion.Thumbnails = Thumbnails;
+
+Accordion.Truncate = Truncate;
+
+Accordion.propTypes = {
+  /** Sets the border-radius of Accordion.Container, AccordionBox, and TitleWrapper */
```
michaeljaltamirano, line 112: These were originally set on `propTypes` because it was required for class components to populate the Docs page. The storybook for this PR has lost these comments (https://curology-radiance-pr-451.herokuapp.com/?path=/docs/components-accordion), so we'll have to move the JSDoc comments to the Type definition instead of `propTypes`.

daigof: Oh these needs to go to the AccordionProps yes

src/shared-components/alert/index.tsx
```diff
@@ -28,18 +28,14 @@ const alertIconMapping = {
...
-};
-
-type AlertState = {
-  exiting: boolean;
-  exited: boolean;
+  ctaContent?: React.ReactNode;
+  duration?: string | number;
+  truncateText?: boolean;
+  type?: AlertType;
+  onExit?: () => void;
```
michaeljaltamirano, line 37: [nit] Alphabetically, this should come after `duration` 🙂

src/shared-components/alert/index.tsx
```diff
@@ -61,45 +57,37 @@ type AlertState = {
...
+    window.clearTimeout(timer);
 
-  timer: number | undefined;
-
-  state = {
-    exiting: false,
-    exited: false,
+    window.setTimeout(() => {
+      setExited(true);
+      onExit();
```
michaeljaltamirano, line 83: The old diff has it like so: `onExit({ ...rest });`.

src/shared-components/alert/index.tsx
```diff
@@ -114,77 +102,65 @@ export class Alert extends React.Component<AlertProps, AlertState> {
...
-
-  alertExitHandler = () => {
-    const { onExit, ...rest } = this.props;
-    this.setState({ exiting: true });
-
-    window.clearTimeout(this.timer);
-
-    window.setTimeout(() => {
-      this.setState({ exited: true });
+    // ComponentWillUnmount cleanup
```
michaeljaltamirano, line 110: We can remove this note--we'll have this repo fully converted over so I don't think we don't need to keep referencing class components.

src/shared-components/alert/index.tsx
```diff
@@ -114,77 +102,65 @@ export class Alert extends React.Component<AlertProps, AlertState> {
...
+);
+
+Alert.propTypes = {
+  avatarSrc: PropTypes.string,
+  content: PropTypes.node.isRequired,
+  ctaContent: PropTypes.node,
+  duration: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  onExit: PropTypes.func,
+  truncateText: PropTypes.bool,
+  type: PropTypes.oneOf(['success', 'error', 'default']),
```
michaeljaltamirano, line 165: [Nit]: We have `danger` in `AlertType` but not here, we should add it here as well.

src/shared-components/banner/index.tsx
```diff
@@ -34,7 +34,11 @@ type BannerProps = {
  *
  * Banners are not dismissable.
  */
-export const Banner = ({ content, type = 'default', onClick }: BannerProps) => {
+export const Banner = ({
+  content,
+  type = 'default',
```
michaeljaltamirano, line 39: [Nit] Let's alphabetize args to match newly alphabetized `BannerProps`.

src/shared-components/banner/index.tsx
```diff
@@ -55,6 +59,6 @@ export const Banner = ({ content, type = 'default', onClick }: BannerProps) => {
 
 Banner.propTypes = {
   content: PropTypes.node.isRequired,
-  type: PropTypes.oneOf(['default', 'success', 'error']),
   onClick: PropTypes.func,
+  type: PropTypes.oneOf(['default', 'success', 'error']),
```
michaeljaltamirano, line 63: [Nit] Looks like `danger` is also in `BannerType` but not here.

src/shared-components/banner/index.tsx
```diff
@@ -34,7 +34,11 @@ type BannerProps = {
  *
  * Banners are not dismissable.
  */
-export const Banner = ({ content, type = 'default', onClick }: BannerProps) => {
+export const Banner = ({
+  content,
+  type = 'default',
+  onClick = undefined,
```
michaeljaltamirano, line 40: [Nit] Some not-yet-converted-to-TypeScript files needed to mark optional functions as undefined due to TypeScript inference of JavaScript files--it looks like we don't need it here, though. Can we remove the explicit undefined?

src/shared-components/button/index.tsx
```diff
@@ -63,99 +63,83 @@ type ButtonProps = {
...
+export const Button = ({
+  buttonColor = COLORS.primary,
+  buttonType = 'primary',
+  children,
+  disabled = false,
+  icon = null,
+  isFullWidth = false,
+  isLoading = false,
+  loading = undefined,
+  onClick = () => undefined,
```
michaeljaltamirano, line 75: [Nit] This is out of the scope for this PR but I don't think a Button without an `onClick` makes sense. If we want I think it's within our purview to remove this for improved type sensibility.

src/shared-components/button/index.tsx
```diff
@@ -63,99 +63,83 @@ type ButtonProps = {
...
-          >
-            {children}
-          </ButtonText>
-        </ButtonContents>
-        <Loader
-          buttonColor={buttonColor}
-          buttonType={buttonType}
-          disabled={disabled}
-          isFullWidth={isFullWidth}
           isLoading={loadingVal}
```
michaeljaltamirano, line 104: [Nit] It looks like we re-arranged some props in alphabetical order, but not entirely: `isFullWidth` after `isLoading` in `ButtonContents`, `css` last instead of first in `ButtonText`.

src/shared-components/button/index.tsx
```diff
@@ -63,99 +63,83 @@ type ButtonProps = {
...
+        <ButtonText
           hasIcon={!!icon}
-          isFullWidth={isFullWidth}
-        >
-          {icon}
-          <ButtonText
-            isLoading={loadingVal}
-            hasIcon={!!icon}
-            css={css`
-              padding-top: 2px;
```
michaeljaltamirano, line 141: [Not originated from this PR] `ButtonText` is already a styled component with `padding-top: 2px;`--do we know why this was added, too? Seems like we should get rid of it.

michaeljaltamirano (approved):
Love to see Chromatic report `No visual changes were found in this build.` :tada:

Only blocking changes are the `onExit` API (e.g. passing in `...rest` as an arg) and fixing the Storybook Docs page for Accordion!

package.json
daigof: I guess yes since it is following same version. I also fixed `eslint-plugin-emotion` for the same reason

src/shared-components/accordion/index.tsx
daigof: done

src/shared-components/alert/index.tsx
daigof: I swear I passed over the files before pushing like 5 times and I was sure I would have missed some

src/shared-components/accordion/index.tsx
daigof: done

src/shared-components/alert/index.tsx
daigof: done

src/shared-components/alert/index.tsx
daigof: Yeah there was a reason for that, danger is not supported anymore but there are still alerts populating from Backend that may have that type (that I couldnt fix) So danger was left there as a failsafe mechanism since the default alert is default type. I guess I agree to add it now since when we rely on TS people will see the danger type anyway. This needs to be solved in a different PR along with the other components like Banner

src/shared-components/banner/index.tsx
daigof: done

src/shared-components/banner/index.tsx
daigof: done

src/shared-components/banner/index.tsx
daigof: done

src/shared-components/button/index.tsx
daigof: So You are saying to make onClick required? I changed to required

src/shared-components/button/index.tsx
daigof: done

src/shared-components/button/index.tsx
daigof: removed, Snapshot test confirms there should not be any visual regression

src/shared-components/alert/index.tsx
daigof: Yeah that was on purpose, This logic didnt make sense  to me and is not used anywhere else in Radiance. Then going to Curology I couldnt find any reference that they are using the onExit params. I have a commit ready to revert back this change and leave it as it was, what do you think we should remove the onExit params or restore?

daigof: I just restored the old signature.

michaeljaltamirano: @daigof I'm okay if API changes are made after investigating and seeing that we're not using it, but in general I'd prefer more functional changes (like this one, compared to removing an optional onClick that we provide that does nothing and instead making it require) be done in separate PRs so the change is not lost in the noise, and so we can track it more clearly in the `CHANGELOG`. You can merge in this PR either way, just want to make sure we're documenting changes like this.

PR #461, opened 2020-10-26
[Cleanup] - Part 2

daigof:
- Convert Carousel,DialogModal and OffClickWrapper to FC
- Other Minor ordering stuff. There will be a final PR part 3 that cover all the Alphabetically Ordering and all checks of Props and Refactor of Colors constants usage

### DIFF
--- a/src/shared-components/carousel/arrow/index.tsx
+++ b/src/shared-components/carousel/arrow/index.tsx
@@ -8,65 +8,51 @@ import { ArrowContainer, BottomRightAlignedArrowContainer } from './style';
 
 type ArrowProps = {
   bottomRightAlignedArrows?: boolean;
-  prev?: boolean;
-  next?: boolean;
   disabled?: boolean;
-  onClick: () => void;
+  next?: boolean;
+  onClick?: () => void;
+  prev?: boolean;
 };
 
-class Arrow extends React.Component<ArrowProps> {
-  static propTypes = {
-    bottomRightAlignedArrows: PropTypes.bool,
-    prev: PropTypes.bool,
-    next: PropTypes.bool,
-    disabled: PropTypes.bool,
-    onClick: PropTypes.func.isRequired,
-  };
+const Arrow = ({
+  bottomRightAlignedArrows = false,
+  disabled = false,
+  next = false,
+  onClick = () => undefined,
+  prev = false,
+}: ArrowProps) => {
+  const ArrowContainerComponent = bottomRightAlignedArrows
+    ? BottomRightAlignedArrowContainer
+    : ArrowContainer;
 
-  static defaultProps = {
-    bottomRightAlignedArrows: false,
-    prev: false,
-    next: false,
-    disabled: false,
-  };
+  return (
+    <ArrowContainerComponent prev={prev} next={next} disabled={disabled}>
+      {prev && (
+        <RoundButton
+          buttonType={bottomRightAlignedArrows ? 'primary' : 'action'}
+          disabled={disabled}
+          icon={<ArrowLeftIcon />}
+          onClick={onClick}
+        />
+      )}
+      {next && (
+        <RoundButton
+          buttonType={bottomRightAlignedArrows ? 'primary' : 'action'}
+          disabled={disabled}
+          icon={<ArrowRightIcon />}
+          onClick={onClick}
+        />
+      )}
+    </ArrowContainerComponent>
+  );
+};
 
-  arrowClickHandler = () => {
-    const { onClick } = this.props;
-    onClick();
-  };
-
-  render() {
-    const {
-      bottomRightAlignedArrows = false,
-      prev = false,
-      next = false,
-      disabled = false,
-    } = this.props;
-    const ArrowContainerComponent = bottomRightAlignedArrows
-      ? BottomRightAlignedArrowContainer
-      : ArrowContainer;
-
-    return (
-      <ArrowContainerComponent prev={prev} next={next} disabled={disabled}>
-        {prev && (
-          <RoundButton
-            buttonType={bottomRightAlignedArrows ? 'primary' : 'action'}
-            disabled={disabled}
-            icon={<ArrowLeftIcon />}
-            onClick={this.arrowClickHandler}
-          />
-        )}
-        {next && (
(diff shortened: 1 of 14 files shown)


src/shared-components/offClickWrapper/index.tsx
```diff
@@ -71,13 +40,30 @@ export class OffClickWrapper extends React.Component<OffClickWrapperProps> {
...
-      <div ref={this.containerRef} className={className}>
-        {children}
-      </div>
-    );
-  }
-}
+  useEffect(() => {
+    addOffClickListener();
+    return () => removeOffClickListener();
+  });
```
michaeljaltamirano, line 56:
```suggestion
  }, []);
```

src/shared-components/carousel/arrow/index.tsx
```diff
@@ -8,65 +8,51 @@ import { ArrowContainer, BottomRightAlignedArrowContainer } from './style';
...
-      disabled = false,
-    } = this.props;
-    const ArrowContainerComponent = bottomRightAlignedArrows
-      ? BottomRightAlignedArrowContainer
-      : ArrowContainer;
+const Arrow = ({
+  bottomRightAlignedArrows = false,
+  disabled = false,
+  next = false,
+  onClick = () => undefined,
```
michaeljaltamirano, line 21: I think we should keep this as required, and without default args, no?

src/shared-components/carousel/index.tsx
```diff
@@ -45,195 +46,140 @@ type CarouselProps = {
...
-  }
-
-  componentDidUpdate() {
-    if (this.shouldReplay) {
-      this.replay();
-    }
-
-    this.updateLastIndex();
-  }
+  const [currentIndex, setCurrentIndex] = useState(0);
```
michaeljaltamirano, line 73:
```suggestion
  const [currentIndex, setCurrentIndex] = useState(FIRST_INDEX);
```

src/shared-components/carousel/index.tsx
```diff
@@ -45,195 +46,140 @@ type CarouselProps = {
...
+  const [lastIndex, setLastIndex] = useState(getLastIndex());
+  const [hasUserInteracted, setHasUserInteracted] = useState(false);
+  const [timeoutId, setTimeoutId] = useState<number | null>(null);
```
michaeljaltamirano, line 76: This was previously defined as an instance property--I think we should implement this with `useRef` to reflect that.

src/shared-components/carousel/index.tsx
```diff
@@ -45,195 +46,140 @@ type CarouselProps = {
...
 
-  get shouldReplay() {
-    const { autoplay, infinite } = this.props;
-    const { currentIndex, lastIndex } = this.state;
+  const slider: React.RefObject<Slider> = React.createRef();
```
michaeljaltamirano, line 78: Should we be using `useRef` here? I also think we should try and use built-in Generics typing as much as possible to avoid in-line typing:

image

src/shared-components/carousel/index.tsx
```diff
@@ -45,195 +46,140 @@ type CarouselProps = {
...
-
-  componentDidUpdate() {
-    if (this.shouldReplay) {
-      this.replay();
-    }
-
-    this.updateLastIndex();
-  }
+  const [currentIndex, setCurrentIndex] = useState(0);
+  const [lastIndex, setLastIndex] = useState(getLastIndex());
```
michaeljaltamirano, line 74:
```suggestion
  const [lastIndex, setLastIndex] = useState(getLastIndex);
```

It's minor, but `useState` also accepts a function so that we only invoke the method once, instead of every time the component re-renders: https://www.benmvp.com/blog/four-characters-optimize-react-component/

src/shared-components/carousel/index.tsx
```diff
@@ -45,195 +46,140 @@ type CarouselProps = {
...
+  const [hasUserInteracted, setHasUserInteracted] = useState(false);
```
michaeljaltamirano, line 75: This was also previously an instance method--do you think we should keep it as a `useRef` value instead?

michaeljaltamirano (approved):
Some open questions on state vs. refs in Carousel, but otherwise LGTM!

src/shared-components/carousel/arrow/index.tsx
daigof: I initially left it as it was but the way the slider loads the Arrows makes it throw an error passing null to onClick (where we clearly define the prop in the Slider settings) You can check this in PROD in the console https://radiance-ui.curology.com/?path=/story/components-carousel--one-card

since we have no control of Slider but total control of Arrow I decided to change it with defauld value for the console error to disappear

src/shared-components/carousel/index.tsx
daigof: fixed here https://github.com/curology/radiance-ui/pull/461/commits/95492678284de09c1f4baf05ab2b951aa5be6d2a

src/shared-components/carousel/index.tsx
daigof: Yeah I agree, thanks for the push on this I was lazy to investigate further, I heard the useRef as constant before (instead of DOM usage) but never implemented one. Now I know how to replace those class instance members.

fixed here https://github.com/curology/radiance-ui/pull/461/commits/95492678284de09c1f4baf05ab2b951aa5be6d2a

src/shared-components/carousel/index.tsx
daigof: Yeah totally, I replace also Accordion and Alert  React.createRef usage
fixed here https://github.com/curology/radiance-ui/pull/461/commits/95492678284de09c1f4baf05ab2b951aa5be6d2a

src/shared-components/carousel/index.tsx
daigof: fixed here https://github.com/curology/radiance-ui/pull/461/commits/95492678284de09c1f4baf05ab2b951aa5be6d2a

src/shared-components/carousel/index.tsx
daigof: yep same comment as timeout

PR #436, opened 2020-10-21
[Theme setup] Initial setup and Accordion example

daigof:
- This is mostly a copy of the work from this PR https://github.com/curology/radiance-ui/pull/387

- Converted Accordion to FC. My main concern was defining the theme as AccordionProp while using the HOC, this will prompt consumers of Radiance to pass a theme prop. With hooks this is solved automatically. Tested with console logs the setContentHeight logic and works correctly with the new FC version.

- Wanted to contain the changes to Accordion component, next PR will continue porting the changes from https://github.com/curology/radiance-ui/pull/387 and convert more components to FC

PR #463, opened 2020-10-27
[Theming] Add built-in TypeScript support

michaeljaltamirano:
This PR adds resolution via TypeScript to allow us to continue importing `emotion-theming` and `@emotion/styled` normally but, through path aliasing, decorates those imports to use provided built-in types. This gives us automatic access to our ThemeType without needing to import it in every file/component.

I had also tested out replacing all our instances of `import styled from '@emotion/styled'`, but ultimately this bit of magic seemed preferable to breaking our convention/expectation. We'll have to set up similar resolution magic in consumer apps, but that would be required regardless of whatever configuration we chose in `radiance-ui`.

`.size-snapshot` is so large because I forgot to re-build in #436. There's no regression with `epic/add-theming-functionality`, and it's not caused by these changes.


daigof (approved):
nice, understood, I wonder if we can export some of the definitions here for the consumer apps. We would need to add types folder to package.json files array for example?

michaeljaltamirano:
> nice, understood, I wonder if we can export some of the definitions here for the consumer apps. We would need to add types folder to package.json files array for example?

Which definitions?

PR #464, opened 2020-10-27
Cleanup and  refactors part 3

daigof:
Left comments in code. This could be the final cleanup. Anything u want throw in here let me know


src/shared-components/accordion/thumbnails/index.tsx
```diff
@@ -57,8 +56,4 @@ const Thumbnails = ({ photoSrcs }: ThumbnailsProps) => {
   return <Container>{renderThumbnails()}</Container>;
 };
 
-Thumbnails.propTypes = {
-  photoSrcs: PropTypes.arrayOf(PropTypes.string).isRequired,
-};
-
 export default Thumbnails;
```
daigof, line 59: no proptypes for internal components we rely on TS types now

src/shared-components/callout/index.tsx
```diff
@@ -39,10 +39,10 @@ export const Callout = ({
   </CalloutContainer>
 );
 
+Callout.Container = ParentContainer;
+
```
daigof, line 43: PropTypes goes last

src/shared-components/dropdown/index.tsx
```diff
@@ -30,7 +30,7 @@ type DropdownProps = {
   optionsContainerMaxHeight?: string;
   textAlign?: 'left' | 'center';
   /**
-   * The currently selected option. Can mount as `null`
+   * The currently selected option. Can mount as `undefined`
```
daigof, line 33: there was never a default null, if value is not provided the default is undefined

src/shared-components/dropdown/index.tsx
```diff
@@ -42,7 +42,7 @@ type DropdownProps = {
 export const Dropdown = ({
   borderRadius = '4px',
   onChange,
-  options,
+  options = [{ value: undefined, label: '' }],
```
daigof, line 45: Even though options is required, this was the default value inside Desktop and Mobile Components so to DRY I placed it up here and we will always pass at least this to those components

src/shared-components/dropdown/index.tsx
```diff
@@ -112,17 +114,17 @@ export const Dropdown = ({
...
       value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
       label: PropTypes.string,
       disabled: PropTypes.bool,
     }),
   ).isRequired,
-  textAlign: PropTypes.oneOf(['left', 'center']),
-  onChange: PropTypes.func.isRequired,
   optionsContainerMaxHeight: PropTypes.string,
+  textAlign: PropTypes.oneOf(['left', 'center']),
+  value: allowNullPropType(
```
daigof, line 127: I dont get this, value was never set as default null so the only reason we should leave this here is if we want consumers to pass null to value explicitly

src/shared-components/dropdown/test.tsx
```diff
@@ -81,6 +83,8 @@ describe('<DesktopDropdown />', () => {
         onSelectClick={() => null}
         closeDropdown={() => null}
         onOptionClick={() => null}
+        textAlign="left"
```
daigof, line 86: Mobile and Desktop Dropdowns are never meant to be used directly those values are now required because I know we have those available in the index. These are just fillers to make the TS pass and has no impact on test or snapshots. 
We never use Mobile/Desktop dropdowns directly

src/shared-components/loadingSpinner/__snapshots__/test.tsx.snap
```diff
@@ -101,17 +101,17 @@ exports[`<LoadingSpinner /> UI snapshots renders the correct css with custom pro
...
     <span
       className="emotion-0 emotion-1"
-      color="#FFFFFF"
+      color="#ffffff"
       size="20px"
     />
     <span
       className="emotion-0 emotion-1"
-      color="#FFFFFF"
+      color="#ffffff"
```
daigof, line 114: replaced hardoded with {COLORS.white}. below

src/shared-components/dropdown/index.tsx
daigof: Actually if we allow passing null explicitly this comment was sort of right so I will revert this change

src/shared-components/chip/style.ts
```diff
@@ -14,7 +14,7 @@ export const ChipText = styled.span`
 `;
 
 const defaultStyle = css`
-  background-color: ${COLORS.purple10};
+  background-color: ${COLORS.disabled};
```
michaeljaltamirano, line 17:
```suggestion
  background-color: ${COLORS.defaultBackground};
```

They're the same value, but semantically `chip` component can't be disabled, so I think we should prefer `defaultBackground`.

src/shared-components/dropdown/index.tsx
daigof: Well looks like we allow null and number as well and that was not reflected in the previous version. I pushed a fix for that, still dont know who would pass null as value

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -94,7 +94,7 @@ export const DesktopDropdown = ({
           isOpen={isOpen}
           optionsContainerMaxHeight={optionsContainerMaxHeight}
           role="menu"
-          aria-activedescendant={value}
+          aria-activedescendant={value as string}
```
michaeljaltamirano, line 97: Can we remove these two `as string` casts now that we've widened the `value` type? Especially for `aria-activedescendant`, which doesn't look like it needs casting:

```tsx
aria-activedescendant: string | number | null | undefined
```

michaeljaltamirano (changes requested):
37 of 40 files look good to go. The `Request changes` is for `DesktopDropdown`

Our TypeScript types and `prop-types` have been inconsistent in this component (and sub-components), which is understandable. Our TypeScript types only allowed `string`, not `number`, and it appears that is because the `DesktopDropdown` component *only works when values are provided as strings*. You can test this out yourself with the Storybook locally by changing the values in the options array to numbers instead of strings. 

The HTML elements we use *do* accept other types (`value?: string | number | readonly string[] | undefined`), which is why our `MobileDropdown` works, since it relies on those native elements. Our `DesktopDropdown` component, however, does not, and as such any `number` we pass is in casted to a `string` in our event handlers. Because of this, when we update the state in the component, we don't see the result we're expecting. 

So the changes requested here would be to make the `DesktopDropdown` compatible with `value: number` types. Feel free to @ me if you have any questions about this, and re-tag me when it's ready for re-review!

src/shared-components/dropdown/index.tsx
```diff
@@ -32,7 +32,7 @@ type DropdownProps = {
   /**
    * The currently selected option. Can mount as `null`
    */
-  value?: string;
+  value?: string | number | null;
```
michaeljaltamirano, line 35: I think we should try to match the allowed types for `<option>` and `<select>`: `value?: string | number | readonly string[] | undefined` (Feel free to skip `readonly string[]` since it doesn't fit our use case.)

src/shared-components/dropdown/index.tsx
michaeljaltamirano: If we tighten up the types / defaults, I'm comfortable removing this utility from `radiance-ui`.

daigof:
@michaeljaltamirano thanks for the feedback, I didnt wanted to dig much into Dropdown, it was sort of a confusing and just wanted to cleanup without impact. But screw that I overhauled the component, I took the requirement to allow `string | number | undefined` for values (as initial value and optionType). And I reflected that in the story changes so we can test all scenarios.

I suppose there will be some Chromatic changes because of this

src/shared-components/dropdown/desktopDropdown.tsx
```diff
@@ -93,22 +94,22 @@ export const DesktopDropdown = ({
           isOpen={isOpen}
           optionsContainerMaxHeight={optionsContainerMaxHeight}
           role="menu"
-          aria-activedescendant={value}
+          aria-activedescendant={`${value}`}
```
michaeljaltamirano, line 97: I don't think this is quite what we want--if `value` is `undefined` or `null` we will then pass in `"undefined"` or `"null"`. 
```tsx
value ? `${value}` : undefined
```

 perhaps? (It looks like the DOM type is `string | undefined`).

michaeljaltamirano (changes requested):
> @michaeljaltamirano thanks for the feedback, I didnt wanted to dig much into Dropdown, it was sort of a confusing and just wanted to cleanup without impact. But screw that I overhauled the component, I took the requirement to allow `string | number | undefined` for values (as initial value and optionType). And I reflected that in the story changes so we can test all scenarios.
> 
> I suppose there will be some Chromatic changes because of this
> 
> All changes are contained in this last Commit [2f9c0cd](https://github.com/curology/radiance-ui/commit/2f9c0cdc3f64b88a3cde4b92cf6fd391c47c0f9d)

Yep, the chromatic tests are breaking, which I don't think should be happen for this kind of re-factor.

1. It looks like we can avoid it for the `Mobile` story by using the same label: let's keep it pre-populating with `First Option` instead of `Number Value`. If you want to add an explicit number value option, let's not make it the first one to keep the visual tests passing. 

Screen Shot 2020-10-27 at 5 08 57 PM

1. It looks like the changes have also changed the default behavior of the DesktopDropdown, which now auto-populates:

Screen Shot 2020-10-27 at 5 08 48 PM

Before:

Screen Shot 2020-10-27 at 5 31 48 PM

After: undefined option has `aria-checked="true"` before interaction

Screen Shot 2020-10-27 at 5 32 05 PM

It does look like we're passing the actual type in the Desktop version now, which is nice, but let's button this up a bit before merging it in.

daigof:
@michaeljaltamirano changes are in this commit https://github.com/curology/radiance-ui/pull/464/commits/5179ca78ca34e215e39e086cebdf068ed369305b

I just left strings and numbers as values examples and restored the original labels for the story.

src/shared-components/dropdown/index.tsx
```diff
@@ -30,9 +29,9 @@ type DropdownProps = {
   optionsContainerMaxHeight?: string;
   textAlign?: 'left' | 'center';
   /**
-   * The currently selected option. Can mount as `null`
+   * The currently selected option. Can be initially `undefined`
```
michaeljaltamirano, line 32: Previous versions showed that anything `undefined` would be pre-selected, so I think that we might want to re-word this. I'm just not sure what `initially` undefined means, and what the use case would be of something going from initially undefined to some truthy value.

michaeljaltamirano (approved):
I appreciate the comments in `onDesktopSelectChange`. This LGTM. I left one other comment but it is non-blocking (though I do think it would be helpful to be clarified, lest there be confusion about it in the future re: `undefined` use cases.)

Since there were more refactors for Dropdowns than other components let's just be sure to especially double-check those when updating consumer apps.

PR #466, opened 2020-10-28
[Epic] Add Theming Functionality

michaeljaltamirano:
This PR includes all of the changes necessary for a new major version (`v13`) of `radiance-ui`. 

1. Adds theming around all component usage.
    - Exports `primaryTheme` and work-in-progress `secondaryTheme` from `radiance-ui/lib/constants`, too. 
1. Overhauls storybook and test suite for new theme paradigm. 
1. Updates types declarations (once #504 is merged in).


package.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "radiance-ui",
-  "version": "12.4.0",
+  "version": "13.0.0-alpha.3",
```
michaeljaltamirano, line 3: This will get changed to 13 once we release in `master` anyway. Don't feel like another commit build for it.

daigof (approved):
Great Job. there is a change in Chromatic but I tested and it works the same as PROD there is a slight animation of the cards at the beginning. Maybe there is an option in chromatic to take a snapshot after XXXms. I will checkout on monday

michaeljaltamirano:
Merging this in--it looks like we might need one follow-up PR for types, it doesn't look like `ThemeType` is persisting as I'd expect in the consumer apps, but I'd like to get this in `master`

PR #467, opened 2020-10-28
[Theming] Convert Component Usage (Part I)

michaeljaltamirano:
### What & Why

This PR starts moving over component styling usage from direct `COLORS` imports to pulling from the `theme` object.

It handles a fair number of the trivial cases:

1. Direct `COLORS` property access within component logic.
1. Direct `COLORS` property access within styled components.

Since it passed the Chromatic visual testing, I think it is safe to merge in, and so am tagging for review. 

### Next Steps

1. Deprecate direct `COLORS` property access within stand-alone `css` definitions.
    - We need access to a theme object--we'll either in-line `css` SerializedStyles that we're currently nesting within `styled` components, or we'll set up functions to grant access to the theme object. 
    - **This will likely be a much larger diff, and so I wanted to get an early PR in before tackling that work.**
1. Figure out a plan for handling very loose Colors usage
    - Example: our buttons accept any `buttonColor` prop. We need to tighten it to only be aliases present in a theme, or something else once we work through it.


yarn.lock
```diff
@@ -2110,6 +2110,17 @@
     react-syntax-highlighter "^12.2.1"
     regenerator-runtime "^0.13.3"
 
+"@storybook/addon-toolbars@^6.0.27":
```
michaeljaltamirano, line 2113: I added this in another PR but accidentally removed it in the #454 bump. This adds it back.

src/shared-components/accordion/style.ts
```diff
@@ -16,11 +15,11 @@ export const ExpansionWrapper = styled.div<{ contentHeight: string }>`
   transition: max-height ${ANIMATION.defaultTiming} ease-in-out;
 `;
 
-const getBorderStyle = (theme: ThemeType, isOpen: boolean) => css`
+const getBorderStyle = (theme: ThemeType, isOpen: boolean) => `
   border: 1px solid ${theme.COLORS.border};
 
   ${ExpansionWrapper} {
-    ${isOpen && `border-top: 1px solid ${theme.COLORS.border}`};
+    ${isOpen ? `border-top: 1px solid ${theme.COLORS.border};` : ''}
```
michaeljaltamirano, line 22: Since I removed the serialized style, I inadvertently broke some sanitizing of the return object `css` was providing us. Now we explicitly add this semi-colon.

src/shared-components/accordion/test.tsx
```diff
@@ -13,40 +12,44 @@ const testAccordionProps = {
...
     const spy = jest.fn();
-    const titleIndex = 0;
 
-    const component = shallow(<AccordionWithTheme onClick={spy} />);
-    const title = component.childAt(titleIndex);
+    const component = mount(
+      <DecoratedAccordion {...testAccordionProps} onClick={spy} />,
+    );
+
+    const title = component.find('div[role="button"]');
```
michaeljaltamirano, line 52: The way we were accessing the title was quite brittle, and adding a higher order component blew up the test. I relied on `mount` in conjunction with `find` to make it more robust.

.size-snapshot.json
```diff
@@ -430,16 +430,16 @@
     }
   },
   "shared-components/alert/style.js": {
-    "bundled": 53374,
-    "minified": 52524,
-    "gzipped": 5533,
+    "bundled": 35497,
+    "minified": 34532,
+    "gzipped": 5262,
```
michaeljaltamirano, line 435: Interestingly, removing the `css` SerializedStyles in `styled` components appears to bloat the size a bit. Removing some of those instances accounts for the change in this PR.

babel.config.js
```diff
@@ -6,4 +6,9 @@ module.exports = {
     ['@babel/plugin-proposal-class-properties', { loose: true }],
     ['@babel/plugin-transform-parameters', { loose: true }],
   ],
+  env: {
+    test: {
+      presets: ['@babel/preset-typescript'],
```
michaeljaltamirano, line 11: This was required to get `tests/utils/decorateWithThemeProvider.tsx` working. It's not really taking advantage of things right now but if we add more typed utilities it might.

src/constants/colors/secondary.ts
```diff
@@ -6,48 +6,48 @@ const SECONDARY_COLORS = {
   // brand colors
   primary: '#4c0000',
   primaryTint1: '#b20000',
-  primaryTint2: null,
+  primaryTint2: '',
   primaryTint3: '#ff0000',
   secondary: '#ff8000',
-  tertiary: null,
+  tertiary: '',
```
michaeljaltamirano, line 12: The `svg.fill` type is `string | undefined`, so `null` was causing problems. Adding an empty string means we maintain some styles like `border: 1px solid` (which defaults to black), whereas `undefined` would add the style `border: 1px solid undefined`, which does not apply.

This is all a TODO, anyway, but this is just an FYI.

src/shared-components/avatar/style.ts
```diff
@@ -17,9 +15,11 @@ const determineSize = ({
   width: ${SIZES[avatarSize]}px;
 `;
 
-const AvatarImage = styled.img`
+const AvatarImage = styled.img<{
+  avatarSize: 'small' | 'medium' | 'large';
```
michaeljaltamirano, line 19: As an FYI as we convert things over, since I hadn't added types to this (which was never flagged since `${determineSize}` pulled in the props implicitly), I saw a `ThemeType`-related error. Once I added this typing for `avatarSize` the compiler was able to interpret the theme typing correctly.

daigof (approved):
👍

PR #475, opened 2020-10-29
[Theming] Convert Component Usage (Part II, Buttons)

michaeljaltamirano:
### What & Why

This PR updates the logic in our Button components (and sub-components) to pull `COLORS` usage from a theme object instead of relying on direct imports.

### CHANGELOG

1. Exports colors `as const`, and exports a new `COLORS_PROP_TYPES`, so that we can use strong type-checking and prop-checking in Button component usage
    - This, after we update the TypeScript declaration files, will be **a very breaking change.** We'll have to make sure all consumer component usage of the `buttonColor` prop is passing an `aliased` variable from a theme object. Any usages that fall outside of those use cases will need to be documented and triaged with Design.
1. Adds new theme aliases `secondaryTint1`, `secondaryTint2`, and `secondaryTint3`, to reflect lavender usage.
    - `secondaryTint3`/`lavender10` is not used in `radiance-ui` but is used directly in one of our consumer apps. 
1. In addition to those broader changes, it also removes a number of superfluous `css` tagged template literal usage, which is the cause of the significant decrease in bundle size.

Since there are no visual regressions--nor snapshot regressions--this appears to be safe to merge in.

### Next Steps

- Converting `TYPOGRAPHY_STYLE`. 
    - `baseButtonStyles` relies on `TYPOGRAPHY_STYLE.button`, so there's a chain of serialized style dependence here. This PR was already large enough though that I wanted to get in the discrete Button changes before also changing Typography, since that'll impact practically every component.


src/constants/themes/index.ts
```diff
@@ -1,2 +1,2 @@
-export { default as primaryTheme } from './primary';
-export { default as secondaryTheme } from './secondary';
+export { default as primaryTheme } from './primaryTheme';
```
michaeljaltamirano, line 1: Renamed to match `default export` name.

src/shared-components/button/shared-components/loader/index.tsx
```diff
@@ -1,27 +1,27 @@
...
-  disabled?: boolean;
+  disabled: boolean;
   isFullWidth?: boolean;
-  isLoading?: boolean;
-  textColor?: string;
+  isLoading: boolean;
+  textColor: string;
 };
 
 const Loader = ({
```
michaeljaltamirano, line 17: Loader is only used by Button components internally, so we can type things more strongly.

daigof (approved):
👍

PR #476, opened 2020-10-29
[Theming] Convert Components (Part III)

michaeljaltamirano:
### What & Why

This PR converts **all remaining components** besides our exported `TYPOGRAPHY`-related code. 

Since there are no visual regressions, this is ready for review and merging. 

### Next Steps

- Convert `TYPOGRAPHY_STYLE`
    - We embed a lot of these `css` styles in other styled components, so I wanted to save this for last.
- Update `stories/**` usage
    - We're still importing `COLORS` directly in there. I want to keep non-`src` changes separate. 
- Determine what other properties we should be adding to the theme object. 
    - This work can occur simultaneously to work integrating theming in consumer applications.


.eslintrc
```diff
@@ -54,6 +54,12 @@
       }
     ],
     "no-use-before-define": "off",
+    "object-curly-newline": [
```
michaeljaltamirano, line 57: We'll want to update `eslint-config-curology` eventually, and/or add `eslint-plugin-prettier` since `--fix` overwrites our `prettier` rules, which typically occurs with 4 imports being broken into multiple lines even though the character limit does not exceeded our `prettier` settings.

.size-snapshot.json
```diff
@@ -1,13 +1,13 @@
 {
   "bundle.js": {
-    "bundled": 997397,
-    "minified": 929528,
-    "gzipped": 99130
+    "bundled": 900134,
```
michaeljaltamirano, line 3: Removing a bunch of `css` usage continues to reduce the bundle size considerably. 

Since it isn't triggering regression in Chromatic, my expectation is that this'll be fine in consumer component usage, but just calling out that it's possible we might have to add more of the calls back in order to get things working, which will increase the bundle again.

.storybook/preview-head.html
```diff
@@ -0,0 +1,41 @@
+<style>
+  @font-face {
```
michaeljaltamirano, line 2: Finally moved this stuff to `preview-head.html`, per your earlier suggestion.

src/shared-components/container/style.ts
```diff
@@ -1,13 +1,11 @@
...
 import PropTypes from 'prop-types';
+import { ThemeType } from 'src/constants/themes/types';
 
-import {
- SPACER, BOX_SHADOWS, COLORS, MEDIA_QUERIES, 
-} from '../../constants';
+import { SPACER, BOX_SHADOWS, MEDIA_QUERIES } from '../../constants';
 
-const clickableStyle = css`
+const clickableStyle = `
```
michaeljaltamirano, line 8: Assuming `BOX_SHADOWS` end up in here too, we'll need to change this function signature, but didn't want to move away from the focus on `COLORS` for a v1.

daigof (approved):
👍

daigof:
wait I clicked the wrong link and I reviewed only 1 commit of 4 files. Reviewing all now

daigof (approved):
Alright looks good, I like how u are cleaning up those DOM attributes related to styles

PR #477, opened 2020-10-29
[Theming] Convert Components (Part IV)

michaeljaltamirano:
### What & Why

This is the final PR for converting component `COLORS` usage over to our new theme object(s). 

It also adds `eslint-plugin-prettier` so that our `eslint` rules do not overwrite our `prettier` rules anymore. There's a large number of files touched because I ran `eslint --fix` on `src` and `stories`. 

Since there are no visual regressions, it should be safe to merge. 

### Next Steps

- Remove direct `import { COLORS } from 'radiance-ui/lib/constants'` imports from consumer apps
    - This will prevent unintentional usage outside of the theme object, and code failures will allow us to more easily track down consumer app code we need to modify.


.eslintrc
```diff
@@ -57,7 +57,7 @@
     "object-curly-newline": [
       "warn",
       {
-        "ObjectPattern": { "multiline": false }
+        "ObjectPattern": { "multiline": true, "minProperties": 10 }
```
michaeljaltamirano, line 60: Still trying to figure this out 😅

src/shared-components/typography/index.ts
```diff
@@ -93,15 +93,33 @@ export const style = {
...
+  ${({ theme }) => errorStyle(theme)}
+`;
+const Heading = styled.h2`
+  ${({ theme }) => headingStyle(theme)}
+`;
+const Label = styled.label`
+  ${({ theme }) => labelStyle(theme)}
+`;
+const Link = styled.a`
+  ${linkStyle()}
```
michaeljaltamirano, line 115: Left this without arguments for now, but wanted the function signatures to be identical.

daigof (approved):
👍

PR #480, opened 2020-10-30
[Theming] Update Colors Export

michaeljaltamirano:
### What & Why

1. Removes some `COLORS` usage I missed in `src` components. 
1. Replaces `COLORS` usage in `stories` with theme usage. 
1. **Exports `primaryTheme` and `secondaryTheme`, making available for use in consumer apps at our typical import path: `radiance-ui/lib/constants`**
    - TODO: Add stories once theming is set.
1. Adds `react-is` as `devDependency` to fix `yarn install` warning. 

### Next Steps
- Start implementing theming in consumer apps! 🎉 
- We are still exporting the `COLORS` constant, accessible in consumer apps via `radiance-ui/lib/constants`. Once we have switched over to theming in consumer apps, we will remove this export so that all `COLORS` usage must occur via theming.


src/constants/themes/index.ts
```diff
@@ -1,2 +1,3 @@
-export { default as primaryTheme } from './primaryTheme';
-export { default as secondaryTheme } from './secondaryTheme';
+export { primaryTheme } from './primaryTheme';
+export { secondaryTheme } from './secondaryTheme';
+export { COLORS_PROP_TYPES } from './types';
```
michaeljaltamirano, line 3: This is only used internally.

src/constants/index.ts
```diff
@@ -1,12 +1,14 @@
...
-export { default as COLORS, COLORS_PROP_TYPES } from './colors';
+// TODO: Remove once all consumer usage is using theming
+export { default as COLORS } from './colors';
 export { default as FONTS } from './fonts';
 export { default as GRID } from './grid';
 export { default as LAYOUT } from './layout';
 export { default as MEDIA_QUERIES } from './mediaQueries';
 export { default as PROGRESS_BAR_STATUS } from './progressBarStatus';
 export { default as SPACER } from './spacer';
+export { primaryTheme, secondaryTheme, COLORS_PROP_TYPES } from './themes';
```
michaeljaltamirano, line 12: Consumer apps do not currently use `COLORS_PROP_TYPES` but we make it available in case we create custom components around it and want the `prop-types` checking. We can remove this later if it's unnecessary.

yarn.lock
```diff
@@ -2,6 +2,28 @@
 # yarn lockfile v1
 
 
+"@actions/core@^1.2.4":
```
michaeljaltamirano, line 5: The `yarn.lock` diff is from the Chromatic update in `master`. Not sure why the base comparison is so imprecise with these branches.

package.json
```diff
@@ -124,6 +124,7 @@
     "jest-junit": "^12.0.0",
     "lint-staged": "^10.1.3",
     "prettier": "2.1.2",
+    "react-is": "^16.13.1",
```
daigof, line 127: what u used to determine this version?

src/shared-components/button/components/roundButton/style.ts
```diff
@@ -54,14 +54,14 @@ export const RoundButtonBase = styled(ButtonBase)`
   }
 `;
 
-export const roundButtonLoader = (disabled: boolean) => css`
+export const roundButtonLoader = (disabled: boolean, theme: ThemeType) => css`
   width: 36px;
   margin: -3px -3px 0 0;
 
   ${disabled &&
```
daigof, line 61: good catch here.  Aren't  we were replacing all `&&` with ternary operator

daigof (approved):
👍 only a couple of q

src/shared-components/button/components/roundButton/style.ts
michaeljaltamirano: It's only necessary when in a `styled` tagged template literal. `css` cleans up `false` statements but `jest` tests would fail from CSS parse issues for the same in `styled` usage.

package.json
michaeljaltamirano: It's a sub-package in the `react` repo, and as such the versions almost perfectly align with `react` release. There's no `16.14`, though, and so I thought `^17.x` wasn't appropriate since we're not yet on it.

PR #498, opened 2020-11-02
[Theming] Fix tsc output

michaeljaltamirano:
The changes in `epic/add-theming-functionality` changed a fair amount of scaffolding around on `src` files. The below screenshot shows the difference in `lib` outputs between this branch, the base branch, and `master`:

Screen Shot 2020-11-02 at 1 08 39 PM

The current behavior in master is to output all of the source in `src` without being nested in `src`, while the changes I've made resulted in the type declarations we emit to be nested in a `src` file. This broke current behavior in consumer apps, e.g. TypeScript could not infer the types of files imported from `radiance-ui/lib/constants`. This PR corrects that behavior by moving the new files in `tests` and `types` that were added in this branch nested within `src`, which corrects the output regression. (Hopefully the code comments + README committed in here are sufficient--if not please leave a comment about improvements!) 

I tried another approach using [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html), but it didn't change the nesting issues. 

Any other changes I will explain in the diff.


package.json
```diff
@@ -8,6 +8,7 @@
     "dist",
     "lib",
     "lib/constants",
+    "lib/icons",
```
michaeljaltamirano, line 11: We import directly from here, so I think it should also be present here?

package.json
```diff
@@ -30,7 +31,7 @@
     "build:icons:deprecated": "svgr --expand-props none --template src/utils/svgToIconTemplate/index.js -d src/icons/ src/svgs/deprecated/",
     "build:icons:index": "create-index src/icons/",
     "build:rollup": "rollup -c",
-    "build:babel": "babel src -d lib --extensions '.js,.ts,.tsx'",
+    "build:babel": "babel src -d lib --extensions '.js,.ts,.tsx' --ignore '**/*.test.{js,ts,tsx}'",
```
michaeljaltamirano, line 34: We currently build test files, which seems unnecessary/inadvisable. 

I got side-tracked a bit about keeping actual source vs. test separate, could be a good Cooldown item to get things up to Open Source Best Practices.

tests/setupTests.ts
```diff
@@ -1,15 +1,19 @@
...
 
-// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
 configure({ adapter: new Adapter() });
 
 window.matchMedia =
   window.matchMedia ||
   function () {
     return {
       matches: false,
+      dispatchEvent: () => false,
```
michaeljaltamirano, line 11: Fleshed out these properties to remove the TypeScript error, and I guess related changes made the `eslint` stuff not matter anymore.

tsconfig.json
```diff
@@ -3,16 +3,13 @@
   "compilerOptions": {
     "baseUrl": "./",
     "declaration": true,
+    "declarationMap": true,
     "emitDeclarationOnly": true,
     "outDir": "lib",
-    "rootDirs": ["src"]
+    // The folder structure in outDir is based on "The longest common path of all non-declaration input files."
```
michaeljaltamirano, line 9: Comments are valid `json` in `tsconfig` files 🙂

package.json
daigof: yeah, it seems you have to manually specify the nested paths here so I don't know how it was working. I remember I had to specify in daylight exactly ```  "files": [
    "dist",
    "dist/static",
    "dist/static/media"
  ],``` otherwise it would not output the media folder . But whatever this seems consistent

daigof (approved):
👍

PR #503, opened 2020-11-04
[TypeScript] Update Declarations

michaeljaltamirano:
### What & Why

After working through type errors in a consumer application, this PR includes the changes in that repo to get the types working properly. In almost all instances we're widening our types, or fixing `null/undefined` issues.

### DIFF
--- a/src/shared-components/button/deprecatedPropsHandler.ts
+++ b/src/shared-components/button/deprecatedPropsHandler.ts
@@ -6,10 +6,12 @@ export const isLoadingPropFunction = (
   props: Record<string, unknown>,
   propName: string,
   componentName: string,
-): Error | void => {
+): Error | null => {
   if (props[propName] !== undefined) {
     return new Error(
       `'loading' prop will be deprecated in a future major release. Please rename 'loading' to 'isLoading' in ${componentName}`,
     );
   }
+
+  return null;
 };
--- a/src/shared-components/button/index.tsx
+++ b/src/shared-components/button/index.tsx
@@ -47,7 +47,7 @@ type ButtonProps = {
    */
   isLoading?: boolean;
   loading?: boolean;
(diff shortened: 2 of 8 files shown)


src/shared-components/button/index.tsx
```diff
@@ -130,7 +133,7 @@ Button.propTypes = {
   isFullWidth: PropTypes.bool,
   isLoading: PropTypes.bool,
   loading: isLoadingPropFunction,
-  onClick: PropTypes.func.isRequired,
+  onClick: PropTypes.func,
```
michaeljaltamirano, line 136: Some of our codes uses buttons with `type="submit"`, and then attaches an `onSubmit` handler to the form component. As such, I was incorrect to suggest we should make this required. Sorry!

src/shared-components/button/deprecatedPropsHandler.ts
```diff
@@ -6,10 +6,12 @@ export const isLoadingPropFunction = (
   props: Record<string, unknown>,
   propName: string,
   componentName: string,
-): Error | void => {
+): Error | null => {
```
michaeljaltamirano, line 9: Fixes error in `styled(Button)` usage.

src/shared-components/immersiveModal/index.tsx
```diff
@@ -43,7 +43,7 @@ type ImmersiveModalProps = {
   /**
    * The title of the modal
    */
-  title?: string;
+  title?: string | React.ReactNode;
```
michaeljaltamirano, line 46: We have one instance to cover that is:

```
<React.Fragment>
   Some text here
   <span>Some emoji here</span>
<React.Fragment>
```

michaeljaltamirano:
Closing in favor of #504

PR #504, opened 2020-11-04
[Theming] Update Declarations 

michaeljaltamirano:
### What & Why

After working through type errors in a consumer application, this PR includes the changes in that repo to get the types working properly. In almost all instances we're widening our types, or fixing `null/undefined` issues. 

See also: #503 for initial PR (merged from db467ab).


src/shared-components/button/deprecatedPropsHandler.ts
```diff
@@ -6,10 +6,12 @@ export const isLoadingPropFunction = (
   props: Record<string, unknown>,
   propName: string,
   componentName: string,
-): Error | void => {
+): Error | null => {
```
michaeljaltamirano, line 9: Fixes error in styled(Button) usage.

src/shared-components/button/index.tsx
```diff
@@ -45,7 +45,9 @@ type ButtonProps = {
    */
   isLoading?: boolean;
   loading?: boolean;
-  onClick: () => void;
+  onClick?: (
```
michaeljaltamirano, line 48: Some of our codes uses buttons with `type="submit"`, and then attaches an `onSubmit` handler to the form component. As such, I was incorrect to suggest we should make this required. Sorry!

src/shared-components/immersiveModal/index.tsx
```diff
@@ -43,7 +43,7 @@ type ImmersiveModalProps = {
   /**
    * The title of the modal
    */
-  title?: string;
+  title?: string | React.ReactNode;
```
michaeljaltamirano, line 46: We have one instance to cover that is:

```
<React.Fragment>
   Some text here
   <span>Some emoji here</span>
<React.Fragment>
```

src/shared-components/dropdown/index.tsx
```diff
@@ -31,21 +37,21 @@ type DropdownProps = {
...
-  value?: string | number;
+  value?: OptionValue;
 };
 
 /**
  * `<Dropdown />` is a controlled component and should be wrapped by a parent to control the dropdown's state.
  * This ships with a mobile implementation that will handle mobile devices automatically.
  */
-export const Dropdown = ({
+export const Dropdown = <T extends OptionType>({
```
michaeljaltamirano, line 47: The majority of changes are to support using Generics to infer the type in `onChange`, based on the `options` provided. Here is a screenshot that shows the behavior:

Screen Shot 2020-11-04 at 2 27 43 PM

And the simpler case that applies to most of our usage:

Screen Shot 2020-11-04 at 2 56 04 PM

This gets around the problem we have with `semantic-ui-react` usage where the type is flexible and so we often have to cast things `as string` and such in order to comply with TypeScript. No longer! It will be determined automatically by the options we pass in 🙂

src/shared-components/dropdown/index.tsx
```diff
@@ -59,10 +65,10 @@ export const Dropdown = ({
     event: React.ChangeEvent<HTMLSelectElement>,
   ) => {
     const { target } = event;
-    const { value: nextValue, selectedOptions } = target;
+    const { selectedIndex, selectedOptions } = target;
+    const selectedOption = options[selectedIndex];
     if (selectedOptions && selectedOptions.length) {
-      const { label } = selectedOptions[0];
-      onChange({ value: nextValue, label });
+      onChange(selectedOption);
```
michaeljaltamirano, line 71: I also changed this: it did not make sense to me why we were returning the `label` with the change, but in any case now we're returning the whole option object, including `value` and `label`, so that usage can do whatever it wants with this. 

I have an example of simplifying some code in a consumer app in another PR I will be tagging you in, where we were mapping and transforming an object and removing properties, which led to convoluted types: an initial state value was the whole object, but with `onChange` it turned into a `Partial` property. The new code is simpler.

src/shared-components/dropdown/mobileDropdown.tsx
```diff
@@ -4,28 +4,28 @@ import { useTheme } from 'emotion-theming';
...
+  value?: OptionValue;
 };
 
 /**
  * The mobile dropdown is built into the `<Dropdown />` component and cannot be used individually.
  *
  * `<Dropdown /> ` will determine if the user is on a mobile device and render a true `select` tag with `option`(s).
  */
-export const MobileDropdown = ({
+export const MobileDropdown = <T extends OptionType>({
```
daigof, line 22: I've never seen this, sort of dont understand it but I trust you. My only question is do we need this whatever it's named `<T extends OptionType>` here having declaring the same  above `type MobileDropdownProps<T extends OptionType>`
same question would apply to Desktop

src/shared-components/immersiveModal/index.tsx
daigof: `type React.ReactNode` includes string do you think we can remove string at the beginning

yarn.lock
```diff
@@ -5231,10 +5231,10 @@ commander@^5.0.0:
   resolved "https://registry.yarnpkg.com/commander/-/commander-5.1.0.tgz#46abbd1652f8e059bddaef99bbdcb2ad9cf179ae"
   integrity sha512-P0CysNDQ7rtVw4QIQtm+MRxV66vKFSvlsQvGYXZWR3qFU0jlMKHZZZgw8e+8DSah4UDKMqnknRDQz+xuQXQ/Zg==
 
-commander@^6.0.0:
-  version "6.1.0"
-  resolved "https://registry.yarnpkg.com/commander/-/commander-6.1.0.tgz#f8d722b78103141006b66f4c7ba1e97315ba75bc"
-  integrity sha512-wl7PNrYWd2y5mp1OK/LhTlv8Ff4kQJQRXXAvF+uU/TPNiVJUxZLRYGj/B0y/lPGAVcSbJqH2Za/cvHmrPMC8mA==
+commander@^6.2.0:
```
daigof, line 5234: why do we have changes in yarn lock? I didnt see any change in dependencies

michaeljaltamirano: It's in #503 and the merge PR: `Bump lint-staged from 10.5.0 to 10.5.1 (#497)`

src/shared-components/immersiveModal/index.tsx
michaeljaltamirano: Went with `JSX.Element` instead to narrow the type more from `ReactNode`. Updated in PocketDerm as well without errors.

src/shared-components/dropdown/mobileDropdown.tsx
michaeljaltamirano: It's required in the component usage, but not the type-definition. Updated in f07da61.

daigof (approved):
👍

PR #505, opened 2020-11-04
[Theming] Update Test Suite

michaeljaltamirano:
+262/-363 diff tells the tale: the `decorateWithThemeProvider` wasn't a particularly good (or DRY) approach. Wrapping helper methods around the test utils is a much better approach, especially if we choose to update our testing API later.


daigof (approved):
👍 nice

PR #509, opened 2020-11-09
[Themes] - Add fonts to themes

daigof:
- To be consistent with Colors for now I left the Default Index Font export so we dont have to refactor in apps yet. The baseBodyStyles that does have a theme param will work out of the box


src/constants/fonts/test.ts
```diff
@@ -0,0 +1,10 @@
+import PRIMARY_FONTS from './primary';
+import SECONDARY_FONTS from './secondary';
+
+describe('theme fonts', () => {
+  it('primary and secondary fonts have the same number of properties', () => {
+    expect(Object.keys(PRIMARY_FONTS).length).toEqual(
```
michaeljaltamirano, line 6: While we're here and adding more of these, do you think we should create a helper util to just compare the number of keys of two provided objects? It's a little cute--it's not a big deal to have to write this out every time--but it makes sense to couple some of our test scaffolding around this.

daigof: yep added here https://github.com/curology/radiance-ui/pull/509/commits/837c9c58c0fe600926d923c415dc2643f04ea235

src/utils/compareObjectsKeysLength/index.ts
```diff
@@ -0,0 +1,9 @@
+const compareObjectsKeysLength = (object1: {}, object2: {}) => {
```
michaeljaltamirano, line 1: [Nit] Prefer `Record<string, unknown>`, `{}` will type it as a literal empty object.

Screen Shot 2020-11-09 at 12 36 35 PM

src/utils/compareObjectsKeysLength/index.ts
```diff
@@ -0,0 +1,9 @@
...
+  if (typeof object1 !== 'object' || typeof object2 !== 'object') {
```
michaeljaltamirano, line 2: I'd be okay with removing this. Our tests should be written in TS as well from now on, so if we violate this we should be warned via the compiler as well.

michaeljaltamirano (approved):
LGTM. Commit history looks messed up because I pushed a new `13.x.x` branch we can work off of, based off `master`. That way we can release `FONTS` (and other changes, depending on our workflow this week) under `13.1.x` etc.

PR #511, opened 2020-11-09
[Theming] - Add FONTS support and expand COLORS object

daigof:
- Merged Fonts
- Support for `color: transparent`  in themes


michaeljaltamirano (approved):
Releasing `13.1.0` with support for FONTS via theming and expanded `ThemeColor` type definition.

PR #517, opened 2020-11-11
[Bugfix] Immersive Modal Focus Handling

michaeljaltamirano:
### What & Why

1. This PR removes some errant `focusManager` handling I wrote by hand that was resulting in focus being trapped, inadvertently, on CTA buttons, preventing full scroll behavior on Desktop when interacting with the ImmersiveModal. It was only covering a rare focus case anyway: tabbing to the close icon, then using `arrow down` to scroll through content. It's not a *requirement* to have that focus maintained, but it seemed nice to provide. Only it broke behavior for Desktop:

dashboard

1. This PR also removes `autoFocus` from the `FocusScope` because it resulted in weird click behavior for desktop: since the focus was automatically set, when scrolling through content and clicking on the body of the content, focus would jump to the top of the immersive modal since a top-most element still possessed focus. That element focus is also not visible because `focus-visible` only displays it for keyboard users. 
    - It is a minor inconvenience--tabbing to the first tabbable element--for keyboard users. 

immersiveClick


daigof (approved):
👍

PR #520, opened 2020-11-11
[Theming] Add TYPOGRAPHY_CONSTANTS

michaeljaltamirano:
### What & Why

- Adds TYPOGRAPHY_CONSTANTS support to theming (with some naming changes). 
- Updates our `compareObjectsKeysLength` helper to include support for nested objects, along with error messages for why the test failed:
    **- Objects of different lengths:**Screen Shot 2020-11-11 at 12 04 20 PM
    **- Objects of same length with non-matching keys:** Screen Shot 2020-11-11 at 12 23 34 PM
- Adds test for `primaryTheme` and `secondaryTheme` having the same keys
    - Probably overkill, but why not.


src/constants/themes/primaryTheme.ts
```diff
@@ -1,9 +1,11 @@
 import PRIMARY_COLORS from '../colors/primary';
 import PRIMARY_FONTS from '../fonts/primary';
+import { PRIMARY_TYPOGRAPHY } from '../typography/primaryTypography';
 
 export const primaryTheme = {
   __type: 'primary',
   BOX_SHADOW: {},
   COLORS: PRIMARY_COLORS,
   FONTS: PRIMARY_FONTS,
+  TYPOGRAPHY: PRIMARY_TYPOGRAPHY,
```
michaeljaltamirano, line 10: Went with just `TYPOGRAPHY` instead of `TYPOGRAPHY_CONSTANTS` since we're accessing it off the `theme` now, so it is more obvious they are constants: there's no need to distinguish between `TYPOGRAPHY` and `TYPOGRAPHY_CONSTANTS` since they're not being imported from the same place like we currently do.

src/constants/themes/primaryTheme.ts
```diff
@@ -1,9 +1,11 @@
 import PRIMARY_COLORS from '../colors/primary';
 import PRIMARY_FONTS from '../fonts/primary';
+import { PRIMARY_TYPOGRAPHY } from '../typography/primaryTypography';
```
michaeljaltamirano, line 3: In general I think named exports are preferable

src/constants/typography/secondaryTypography.ts
```diff
@@ -0,0 +1,19 @@
+// TODO: Update when values are finalized
```
michaeljaltamirano, line 1: Keeping the values the same for now, rather than do empty strings like we are for Colors.

src/constants/typography/primaryTypography.ts
```diff
@@ -0,0 +1,18 @@
+const fontSize = {
```
michaeljaltamirano, line 1: These values are duplicated from `src/constants/typography/index` because I don't think we should be wrapping them in `throwOnUndefinedProperty`. The value-add is less certain the more we have converted over to TS.

daigof (approved):
👍

PR #524, opened 2020-11-12
[Theming] Update SECONDARY_COLORS and SECONDARY_FONTS

michaeljaltamirano:
- Updates `SECONDARY_COLORS` (and adds corresponding aliases to `PRIMARY_COLORS` that have not yet been added). 
- Updates `SECONDARY_FONTS` (and Storybook usage).


src/constants/colors/index.ts
```diff
@@ -76,6 +76,7 @@ export const colorAliases = {
   primaryTint1: brandColors.purple85,
   primaryTint2: brandColors.purple70,
   primaryTint3: brandColors.purple30,
+  primaryTint4: brandColors.purple15,
```
michaeljaltamirano, line 79: I suppose this will be removed anyway, so it does not really matter. However the PocketDerm review app currently depends on it.

daigof (approved):
👍

PR #529, opened 2020-11-17
[Theming] Overhaul Icon Generation

michaeljaltamirano:
**This PR does not require urgent review.** The diff is large primarily because we're committing more build artifacts in order to better track changes and modifications.

### Background

`radiance-ui` currently leverages the CLI provided by SVGR to generate all of our icons (as React components) at build time, placed in `radiance-ui/lib/icons`, using raw svg files and a custom transform template, with the output looking like this:

```
import React from 'react';
import { propTypes, defaultProps, iconStyles } from '../utils/icons';

const SvgAcneBodySprayGlyph = (props) => {
  return (
    <svg
      viewBox="0 0 48 48"
      fill="none"
      css={css`
        ${iconStyles(props)};
      `}
      {...props}
    >
      <path
        d="M23.51 16.32h.57c.72 0 1.31-.59 1.31-1.31v-.52c0-.72-.59-1.31-1.31-1.31h-.57v-.58c0-1.16-1.61-1.69-3.11-1.69-1.5 0-3.11.53-3.11 1.69l.04 3.6c-1.35.43-2.33 1.17-2.33 2.22l.09 18.96c.08 1.36 2.34 2.38 5.25 2.38s5.17-1.02 5.25-2.38v-.02l.08-18.94c0-.95-.96-1.66-2.16-2.1zm0-1.86h.64v.61h-.64v-.61zm-3.1-2.3c1.01 0 1.64.28 1.82.44-.19.16-.82.44-1.82.44s-1.64-.28-1.83-.44c.19-.16.82-.44 1.83-.44zm1.81 5.02v.04c-.17.16-.84.43-1.82.43-.98 0-1.64-.27-1.82-.43v-3.23c.54.2 1.19.3 1.82.3.63 0 1.28-.1 1.82-.3v3.19zm2.14 19.71l-.03.42c-.02.28-1.3 1.2-4 1.2s-3.98-.92-4-1.23l-.02-.29V20.22c1.03.56 2.53.85 4.03.85 1.47 0 2.99-.3 4.03-.87v16.69h-.01zm-4.03-17.07c-2.45 0-4.08-.84-4.08-1.39 0-.28.41-.62 1.12-.9.01.02.01.05.02.07.23.85 1.62 1.3 3.01 1.3 1.39 0 2.79-.45 3.02-1.31v-.01c.62.27.99.58.99.84.01.56-1.62 1.4-4.08 1.4zM28.68 17.96l-.45.43 1.93 2.03.45-.43.46-.43-1.94-2.03-.45.43zM28.97 11.84L31 9.91 30.14 9l-2.03 1.93.43.46.43.45zM29.527 14.02l-.031 1.25 4.468.111.032-1.25-4.47-.111z"
        fill="currentColor"
      />
    </svg>
  );
};

SvgAcneBodySprayGlyph.propTypes = propTypes;
SvgAcneBodySprayGlyph.defaultProps = defaultProps;
export default SvgAcneBodySprayGlyph;
```

This setup has worked great for us--adding new icons has been as simple as dropping new SVGs in a folder and letting our build setup handle generating the React component and exporting it from the appropriate directory.

However, this setup does not work with our new theming requirements, where the "themeable" icons may be wholesale swapped out for a distinctly different SVGs. By way of visual example, the `Arrow` and `ChevronIcon` examples in the screenshots below show totally different icons, made possible by the configuration in this PR. 

In order to support this behavior (with the bonus of adding type-safety) we're removing the scaffolding around the SVGR CLI in favor of using the built-in defaults. The CLI requires one file at a time for its transform, so we cannot pass it two files and conditionally return the SVG we want based on props. The setup in this PR prefers straightforwardness, even if it is a bit more manual.

### Alternatives

Alternatives to the work in this PR included: 

1. Extending our `babel` reliance.
    - We build our `lib` folder using `babel` directly, but there doesn't currently exist an `@svgr` plugin for this. [Airbnb has a plugin](https://github.com/airbnb/babel-plugin-inline-react-svg) that transforms `svg` imports to directly in-lined code, but it felt incorrect to use a separate `babel` plugin when we rely on `@svgr` for `rollup` and `webpack` compatibility.
1. Advanced SVG usage
    - I looked into how to conditionally render SVGs. I found material on [SVG stacks](https://simurai.com/blog/2012/04/02/svg-stacks) and [nested SVGs](https://stackoverflow.com/a/14630703), which, in conjunction with a complex `@svgr/cli` transform template, means we could slip in `useTheme` logic. However, that would seem to increase the maintenance burden for all parts touched (svg markup, `@svgr` usage, etc.) to support separate themes within a single SVG file.

**This PR does all the manual work of re-configuring the setup with the aim of keeping our setup of adding new SVGs relatively simple, and making sure it is not error-prone**. 

### Adding a new SVG

There are two new considerations for when adding an SVG.

**1. Is the SVG new and without a themed counterpart?** If so, add the `.svg` file in the appropriate directory in `src/svgs`, generate the component icon (via `yarn run build:icons:all`), and then create a complementary TypeScript file that imports it directly and passes it into `useIcon`. 

```tsx
// src/icons/icons/arrowLeftIcon.ts
import { ArrowLeftIcon } from './svgs';
import { useIcon, IconProps } from '../../shared-components/icon';

export default (props: IconProps) =>
  // We pass in ArrowLeftIcon to the Secondary arg because there is no Secondary complement (yet). 
  useIcon(ArrowLeftIcon, ArrowLeftIcon, props);
```

**2. Is the SVG a themed counterpart to an existing SVG?** If so, add the`.svg` as a sibling file in `src/svgs`, and append the icon filenames with `-primary` and `-secondary`, respectively. Run the same build step and update the exported icon file (with theming logic) to import both the Primary and Secondary icons:

```tsx
// This naming is handled automatically by `@svgr`
import { LogoPrimary, LogoSecondary } from './svgs';
import { IconProps, useIcon } from '../../shared-components/icon';

export default (props: IconProps) => useIcon(LogoPrimary, LogoSecondary, props);
```

```zsh
yarn run v1.22.5
$ jest src/icons/test.ts
 PASS  src/icons/test.ts
  icons
    exported icons
      logos
        ✓ should have the same number of svg-generated logos pairings, with complementary naming
    generated icons
      logos
        ✓ should have the same number of logos svgs as svg-generated components, with complementary naming (1 ms)
```

Ideally, committing more of these built files gives us a better understanding of how this repo works. I think it would be nice to include more automatic handling, but as it is I've invested a bit too much time into this, and I don't know how important optimizing our icon build workflow actually is, since they do not change very often. 

### CHANGELOG

1. `src/` and `stories/` icon imports updated to use React components with conditional rendering logic, so that when we switch the theme context the icons reflect the changes. (This was not possible with direct `.svg` imports.)
1. Removes `src/utils/icons` and `src/utils/svgToIconTemplate` in favor of `src/shared-components/icon` and SVGR config options. 
1. Removes `src/svgs/deprecated` (none were in use).
1. Updates `SVGR` packages
    - The blocker to our updating was due to our heavy reliance on the CLI, which is no longer the case with thees changes. 

### TODOs

- [x] Update `create-index`: PR is ready to be merged in following approval of this PR: https://github.com/curology/create-index/pull/1
- [x] Update storybook
- [ ] Update documentation (Will do after PR approval, before merging in--until approval I'm not sure what will need to be documented and what will change)
- [x] Double-check the snapshot diff is O.K. by digging more into Icon/useIcon behavior. 
    - We're actually capturing more detail now!


src/icons/emojis/anxiousEmoji.ts
```diff
@@ -0,0 +1,5 @@
+import AnxiousEmojiPrimary from './svgs/anxious-emoji.svg';
+import { useIcon, IconProps } from '../../shared-components/icon';
+
+export default (props: IconProps) =>
```
michaeljaltamirano, line 4: The default behavior of `create-index` is to generate an `index.ts` file comprising `default` exports renamed as their "safe" named export value. It automatically picks up that this should be `export { default as AnxiousEmoji }` from the filename. 

There's good React reasons to not use anonymous default exports--the component shows up as `Anonymous` in the Component tree in dev tools--but since the underlying components in `useIcon` will be named, this seemed a fine trade-off to make for enhanced compatibility with `create-index`. Moreover, It reduces auto-import name-spacing noise a bit. (Currently if you add `AnxiousEmoji` in a file, the auto-import will suggests both `../../src/icons` and `../../src/icons/emojis` as relative imports.)

stories/icons/mockIcon.stories.tsx
```diff
@@ -1,28 +0,0 @@
-import React from 'react';
-import { IconsProps, defaultProps, propTypes } from 'src/utils/icons';
-
-/**
- * We construct a mock Icon components to use Storybook's ArgsTable functionality.
```
michaeljaltamirano, line 5: With our new `Icon` component we no longer need to do this, so I moved all the Icon logic into `index.stories.tsx` instead of having the separate files + MDX.

stories/icons/index.stories.tsx
```diff
@@ -1,3 +1,154 @@
...
+    </Style.Icons>
+  </React.Fragment>
+);
+
+IconLibrary.parameters = {
+  chromatic: { viewports: [BREAKPOINTS.xs] },
+};
+
+const addingSvgText = `
+In order to add an SVG to Radiance, you first need to determine what type of icon you're adding:
```
michaeljaltamirano, line 106: This section has been updated to reflect the changes in this PR.

src/shared-components/icon/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+import PropTypes from 'prop-types';
+import { useTheme } from 'emotion-theming';
+
+import Style from './style';
+
+export interface IconProps extends React.SVGProps<SVGSVGElement> {
+  className?: string;
+  fill?: string;
+  height?: string | number;
+  displayInline?: boolean;
```
michaeljaltamirano, line 11: Renamed `inline` to `displayInline` to prevent an invalid DOM type error. There's only a few usages where we violate this, so it's a quick fix, but this is also technically a breaking change that I will make in the PRs that update `radiance-ui` versions.

src/utils/icons/style.ts
```diff
@@ -1,20 +0,0 @@
-import styled from '@emotion/styled';
-
-export const Icons = styled.div`
```
michaeljaltamirano, line 3: These were only ever used for Storybook anyway, and so were already not used since I did not catch that in my earlier Storybook conversion work.

.storybook/webpack.config.js
```diff
@@ -41,8 +38,7 @@ module.exports = (webpackSettings) => {
       {
         loader: '@svgr/webpack',
         options: {
-          template: transformTemplateForUtilLocation(UTIL_LOCATION),
-          expandProps: false,
+          expandProps: 'end',
```
michaeljaltamirano, line 41: With `expandProps: 'end'`, we can capture most of what we were achieving with our custom setup:

```diff
- import React from 'react';
- import { css } from '@emotion/core';
- import { propTypes, defaultProps, iconStyles } from '../utils/icons';
- 
- const SvgAcneBodySprayGlyph = (props) => {
-   return (
-     <svg
-       viewBox="0 0 48 48"
-       fill="none"
-       css={css`
-         ${iconStyles(props)};
-       `}
-       {...props}
-     >
-       <path
-         d="M23.51 16.32h.57c.72 0 1.31-.59 1.31-1.31v-.52c0-.72-.59-1.31-1.31-1.31h-.57v-.58c0-1.16-1.61-1.69-3.11-1.69-1.5 0-3.11.53-3.11 1.69l.04 3.6c-1.35.43-2.33 1.17-2.33 2.22l.09 18.96c.08 1.36 2.34 2.38 5.25 2.38s5.17-1.02 5.25-2.38v-.02l.08-18.94c0-.95-.96-1.66-2.16-2.1zm0-1.86h.64v.61h-.64v-.61zm-3.1-2.3c1.01 0 1.64.28 1.82.44-.19.16-.82.44-1.82.44s-1.64-.28-1.83-.44c.19-.16.82-.44 1.83-.44zm1.81 5.02v.04c-.17.16-.84.43-1.82.43-.98 0-1.64-.27-1.82-.43v-3.23c.54.2 1.19.3 1.82.3.63 0 1.28-.1 1.82-.3v3.19zm2.14 19.71l-.03.42c-.02.28-1.3 1.2-4 1.2s-3.98-.92-4-1.23l-.02-.29V20.22c1.03.56 2.53.85 4.03.85 1.47 0 2.99-.3 4.03-.87v16.69h-.01zm-4.03-17.07c-2.45 0-4.08-.84-4.08-1.39 0-.28.41-.62 1.12-.9.01.02.01.05.02.07.23.85 1.62 1.3 3.01 1.3 1.39 0 2.79-.45 3.02-1.31v-.01c.62.27.99.58.99.84.01.56-1.62 1.4-4.08 1.4zM28.68 17.96l-.45.43 1.93 2.03.45-.43.46-.43-1.94-2.03-.45.43zM28.97 11.84L31 9.91 30.14 9l-2.03 1.93.43.46.43.45zM29.527 14.02l-.031 1.25 4.468.111.032-1.25-4.47-.111z"
-         fill="currentColor"
-       />
-     </svg>
-   );
- };
- 
- SvgAcneBodySprayGlyph.propTypes = propTypes;
- SvgAcneBodySprayGlyph.defaultProps = defaultProps;
- export default SvgAcneBodySprayGlyph;
+ import * as React from 'react';
+ 
+ function SvgAcneBodySprayGlyph(props) {
+   return (
+     <svg
+       viewBox="0 0 48 48"
+       fill="none"
+       xmlns="http://www.w3.org/2000/svg"
+       {...props}
+     >
+       <path
+         d="M23.51 16.32h.57c.72 0 1.31-.59 1.31-1.31v-.52c0-.72-.59-1.31-1.31-1.31h-.57v-.58c0-1.16-1.61-1.69-3.11-1.69-1.5 0-3.11.53-3.11 1.69l.04 3.6c-1.35.43-2.33 1.17-2.33 2.22l.09 18.96c.08 1.36 2.34 2.38 5.25 2.38s5.17-1.02 5.25-2.38v-.02l.08-18.94c0-.95-.96-1.66-2.16-2.1zm0-1.86h.64v.61h-.64v-.61zm-3.1-2.3c1.01 0 1.64.28 1.82.44-.19.16-.82.44-1.82.44s-1.64-.28-1.83-.44c.19-.16.82-.44 1.83-.44zm1.81 5.02v.04c-.17.16-.84.43-1.82.43-.98 0-1.64-.27-1.82-.43v-3.23c.54.2 1.19.3 1.82.3.63 0 1.28-.1 1.82-.3v3.19zm2.14 19.71l-.03.42c-.02.28-1.3 1.2-4 1.2s-3.98-.92-4-1.23l-.02-.29V20.22c1.03.56 2.53.85 4.03.85 1.47 0 2.99-.3 4.03-.87v16.69h-.01zm-4.03-17.07c-2.45 0-4.08-.84-4.08-1.39 0-.28.41-.62 1.12-.9.01.02.01.05.02.07.23.85 1.62 1.3 3.01 1.3 1.39 0 2.79-.45 3.02-1.31v-.01c.62.27.99.58.99.84.01.56-1.62 1.4-4.08 1.4zM28.68 17.96l-.45.43 1.93 2.03.45-.43.46-.43-1.94-2.03-.45.43zM28.97 11.84L31 9.91 30.14 9l-2.03 1.93.43.46.43.45zM29.527 14.02l-.031 1.25 4.468.111.032-1.25-4.47-.111z"
+         fill="currentColor"
+       />
+     </svg>
+   );
+ }
+ 
+ export default SvgAcneBodySprayGlyph;
```

Prop and type information, as well as the `css` prop application, is handled by `shared-components/icon`. This way, our SVG components are now only concerned with SVG properties, and not component logic around it.

src/icons/index.ts
```diff
@@ -0,0 +1,8 @@
+// Note: this file has to be manually updated if we add more types of icons.
+// We do not generate it via create-index
```
michaeljaltamirano, line 2: The build process previously was:

1. Target all the SVGs in their respective directory (e.g. `src/svgs/icons`), build them as React components, and output the result to `src/icons`, via `@svgr/cli`
1. Build one large `// @create-index` file in `src/icons` from those files.

This process is not theme-able, though, without stepping in and creating an intermediary file between steps 1 and 2. That also raises the possibility of some messy workflows. So the steps are now driven by `@svgr/webpack` (in step 1) instead of `@svgr/cli`: 

1. Import the SVGs from their respective directory (e.g, `src/icons/icons/svgs`) in TypeScript files that export a themed icon (e.g. `src/icons/icons/arrow.ts`)
1. Build a `// @create-index` file in that respective directory (`src/icons/icons`).
1. Import all exports from that directory (and the others) here, in `src/icons`.

src/icons/test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const ICONS_MAP = {
+  emojis: 'emojis',
+  glyphs: 'glyphs',
+  icons: 'icons',
+  logos: 'logos',
+  navIcons: 'navIcons',
+};
+
+describe('icons', () => {
```
michaeljaltamirano, line 49: Since there is more manual effort involved in adding/maintaining existing icons, this test at least ensures that typos do not inadvertently break expected functionality.

.eslintrc
```diff
@@ -123,21 +124,15 @@
     {
       "files": ["*.tsx"],
       "rules": {
-        "react/prop-types": "off"
+        "react/prop-types": "off",
+        "react/jsx-props-no-spreading": "off"
```
michaeljaltamirano, line 128: We now have full type coverage in `radiance-ui/src` 🥳 So we can turn this off for TS files.

package.json
```diff
@@ -13,28 +13,25 @@
   ],
   "types": "./lib/index.d.ts",
   "engines": {
-    "node": ">=8.12.0"
+    "node": ">=10.0"
```
michaeljaltamirano, line 16: This is a change we discussed last month. `@svgr/cli@^5` dropped node 8 support. I didn't notice any issues with this change, and our Heroku pulls in node 15 anyway 😳 . We should probably enforce versions more strictly during cooldown so everyone is on the same version.

src/shared-components/callout/index.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { useTheme } from 'emotion-theming';
 
-import { CalloutContainer, Text, Icon, ParentContainer } from './style';
+import Style from './style';
```
michaeljaltamirano, line 5: Change is to prevent auto-import namespace collision with `Icon`.

package.json
```diff
@@ -13,28 +13,25 @@
...
   },
   "scripts": {
     "test": "jest",
-    "storybook": "run-s build:svgs:index storybook:dev",
+    "storybook": "run-s build:icons:index storybook:dev",
     "storybook:dev": "start-storybook --port 9001 --config-dir .storybook --static-dir ./public",
     "build-storybook": "build-storybook --config-dir .storybook --output-dir .out --static-dir ./public",
-    "clean": "rimraf dist lib src/icons/",
```
michaeljaltamirano, line 23: CHANGELOG:

1. No longer clean `src/icons`--we're committing it now.
1. No longer build icons then use `create-index`--we're committing the icon setup, so we only build the index now.
1. Run `tsc` after `create-index` to ensure type compliance. 
1. Update `@svgr` packages.

.eslintrc
```diff
@@ -123,20 +124,20 @@
...
       }
     },
     {
-      "files": ["stories/**/*.{js,ts,tsx}"],
+      "files": ["src/icons/**/svgs/index.tsx"],
       "rules": {
-        "no-alert": "off"
+        "prettier/prettier": "off"
```
michaeljaltamirano, line 134: The `index.tsx` file generated by `@svgr/cli` lacks a newline. Without this, the respective files will always appear to have been modified, which is annoying. 🙃

rollup.config.js
```diff
@@ -60,6 +52,7 @@ export default [
           '@emotion/styled': 'styled',
           '@emotion/styled-base': '_styled',
           '@react-aria/focus': '@react-aria/focus',
+          'emotion-theming': 'emotionTheming',
```
michaeljaltamirano, line 55: Fixes `build` warning.

src/shared-components/tooltip/__snapshots__/test.tsx.snap
```diff
@@ -59,9 +59,20 @@ exports[`<Tooltip /> UI snapshot renders content and children 1`] = `
   margin-left: -10px;
 }
 
+.emotion-4 {
```
michaeljaltamirano, line 62: The application of `emotion-4` on the `svg` in the snapshot indicates there's no regression with the behavior we were achieving with the custom `@svgr` template, that put the `css` prop directly on the `<svg>` element.

.size-snapshot.json
```diff
@@ -1,123 +1,81 @@
 {
   "bundle.js": {
-    "bundled": 823371,
-    "minified": 754553,
-    "gzipped": 91244
+    "bundled": 820258,
```
michaeljaltamirano, line 3: This shakes up the `dist` build artifact since the icons included in Component logic now reside in `dist/bundle-es/icons/icons` instead of `dist/bundle-es/svgs`, so the file ordering got jumbled around and it's hard to track changes. It's always nice to see a smaller bundle, though 🥳

src/icons/emojis/anxiousEmoji.ts
```diff
@@ -0,0 +1,4 @@
+import { AnxiousEmoji } from './svgs';
+import { useIcon, IconProps } from '../../shared-components/icon';
+
+export default (props: IconProps) => useIcon(AnxiousEmoji, AnxiousEmoji, props);
```
michaeljaltamirano, line 4: The default behavior of `create-index` is to generate an `index.ts` file comprising `default` exports renamed as their "safe" named export value. It automatically picks up that this should be `export { default as AnxiousEmoji }` from the filename. 

There's good React reasons to not use anonymous default exports--the component shows up as `Anonymous` in the Component tree in dev tools--but since the underlying components in `useIcon` will be named (by `@svgr/webpack`), this seemed a fine trade-off to make for maintaining compatibility with `create-index`. 
Screen Shot 2020-11-18 at 10 21 49 AM

Moreover, anonymous default exports reduces auto-import name-spacing noise a bit. (Currently if you add `AnxiousEmoji` in a file, the auto-import will suggests both `../../src/icons` and `../../src/icons/emojis` as relative imports, and TypeScript definitions work best with relative imports.)

.storybook/webpack.config.js
daigof: Since we are importing the icons in the stories now as React components we dont need this anymore, I tested and works. we can also get rid of `@svgr/webpack` I guess

daigof (approved):
Looks good, I like this solution, I like how you got rid of all those confusing ast templates.

Left 1 comment regarding `@svgr/webpack` and I guess I have more cleanups comments for future PRs: 
- we have 3 babel configs: I guess we can remove the `babel.config` at root since it was used by `svgr/webpack` the other 2 are for build and storybook, wonder if we can combine those somehow
- With this PR I think we reached 100% TS coversion. -> we can remove rules extensions that have `js` in many configs I also saw a `md`  somewhere

michaeljaltamirano:
> Looks good, I like this solution, I like how you got rid of all those confusing ast templates.
> 
> Left 1 comment regarding `@svgr/webpack` and I guess I have more cleanups comments for future PRs:
> 
> * we have 3 babel configs: I guess we can remove the `babel.config` at root since it was used by `svgr/webpack` the other 2 are for build and storybook, wonder if we can combine those somehow
> * With this PR I think we reached 100% TS coversion. -> we can remove rules extensions that have `js` in many configs I also saw a `md`  somewhere

Simplified the storybook config and removed the `@svgr/webpack` dependency. I have a follow-up PR planned for streamlining the babel conifg.

PR #550, opened 2020-11-24
[Dependencies] Upgrade Emotion 10 to Emotion 11 

michaeljaltamirano:
Opening a branch mostly for spec and visibility purposes.

https://emotion.sh/docs/emotion-11

### Blockers

1. Unit tests are broken due to underlying dependency (`stylis`)
    - The fix is in place [(see Github issue)](https://github.com/emotion-js/emotion/issues/2103#issuecomment-728877555) but hasn't been released yet, and we're in no rush, anyway.
    - **EDIT**: This turns out to have been caused by our `jest.config` setup. [(See here)](https://github.com/curology/radiance-ui/pull/550#issuecomment-753018630)
1. There might be some `@typescript-eslint` incompatibilities with using the `css` prop (with `theme` as a function argument) in environments not properly configured for it (e.g. `stories/**`). Will look into this later but preferring `useTheme` works fine because the underlying code in `css` calls `React.useContext(ThemeProvider)` anyway so there's no runtime difference. 

### Considerations
- We would want to change how we're handling `css` prop handling if we want to take advantage of the automatic JSX transform: https://emotion.sh/docs/css-prop##babel-preset
    - I'm not bothered by requiring to add `import React from 'react';` when using JSX, but it promises to also provide a reduction in bundle size in some scenarios, which we should investigate more thoroughly.


michaeljaltamirano:
Screen Shot 2020-12-21 at 7 24 38 PM

Still blocked that Stylis aware that is supposed to have been fixed in 4.0.4.

michaeljaltamirano:
> Screen Shot 2020-12-21 at 7 24 38 PM
> 
> Still blocked that Stylis aware that is supposed to have been fixed in 4.0.4.

When removing all Storybook packages, this error goes away, which suggests the problem is conflicting package versions (Storybook relies on many v10 packages). 

However, a new issue appears where the `require('react')` call within emotion itself appears to pull up nothing, such that React is an empty object.

Screen Shot 2020-12-31 at 11 00 40 AM

michaeljaltamirano:
> > Screen Shot 2020-12-21 at 7 24 38 PM
> > Still blocked that Stylis aware that is supposed to have been fixed in 4.0.4.
> 
> When removing all Storybook packages, this error goes away, which suggests the problem is conflicting package versions (Storybook relies on many v10 packages).
> 
> However, a new issue appears where the `require('react')` call within emotion itself appears to pull up nothing, such that React is an empty object.
> 
> Screen Shot 2020-12-31 at 11 00 40 AM

The offending cause of this was our `jest.config` setup, which has been updated to more accurately reflect what we're trying to do (map `src/**` imports to `<rootDir>`, not treat `src` like `node_modules`): https://github.com/curology/radiance-ui/pull/550/commits/69082969768186c74b1194410cee67cebc69e440#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1L17

rollup.config.js
```diff
@@ -30,7 +30,13 @@ const defaultConfig = {
     }),
     sizeSnapshot(),
   ],
-  external: [...Object.keys(pkg.dependencies), '@emotion/styled-base'],
+  /**
+   * @emotion/styled/base needs to be adding explicitly because certain @emotion/styled usage,
```
michaeljaltamirano, line 34: https://github.com/emotion-js/emotion/blob/master/packages/styled/CHANGELOG.md#major-changes

michaeljaltamirano:
Not sure why there's some brittle snapshots in CI. I can't repro the diff locally.

michaeljaltamirano:
Closing in favor of #910, CI build step seemingly breaks once there are too many commits in a very long-standing branch.

PR #551, opened 2020-11-24
[Babel] Streamline multiple config into babel.config.js

michaeljaltamirano:
For historic reasons, our repo has ended up with three separate babel configs:

1. `.storybook/.babelrc`, to support TypeScript-specific handling for `yarn run storybook`
1. `src/.babelrc`, to support src-specific handling (`rollup`, mostly, presumably), added in 2018. 
1. `<rootDir>babel.config.js`, also added in 2018, for general repo config. 

I'm not super interested in illuminating the finer details of the Why and How for each file, but suffice it to say this is a confusing setup. This PR merges all necessary config in `<rootDir>babel.config.js`. There's no regressions in the `dist` and `lib` build.


michaeljaltamirano:
Build appears to be failing not for visual regression reasons but because the markup is different, e.g. `<div class="css-1j5vtd3 e1oss35m6">` to `<div class="css-1133tjo-Container e1oss35m6">`, looking into it now.

babel.config.js
```diff
@@ -1,5 +1,22 @@
...
+    },
+  ],
+  '@babel/react',
+  '@babel/preset-typescript',
+  '@emotion/babel-preset-css-prop',
+];
+
 module.exports = {
-  presets: ['@babel/env', '@babel/react', '@emotion/babel-preset-css-prop'],
+  presets: presets(false),
```
daigof: note storybook was using modules:'auto' and here is picking up as false.
Also in Chromatic apart from the css class name I'm also seeing `css="[object Object]"` in svg components this is likely related to `babel-preset-css-prop` but I dont see how this is different than old config

michaeljaltamirano: @daigof The issue was that storybook doesn't automatically pick up `babel.config.js`, only `.babelrc`. So when we removed the `.storybook/.babelrc` it wasn't picking up the other babel config. I've directly imported it in 
af030c4 and now the build is passing.

.storybook/main.ts
```diff
@@ -27,4 +31,25 @@ module.exports = {
     check: true,
     checkOptions: {},
   },
+  babel: babelConfig,
```
michaeljaltamirano, line 34: See https://storybook.js.org/docs/react/configure/babel#custom-config-file for why I broke the build, and just below that for why this fixes things.

.storybook/main.ts
```diff
@@ -27,4 +31,25 @@ module.exports = {
...
+  webpackFinal: (config, { configType }) => {
```
michaeljaltamirano, line 35: It looks like this is the preferred way to add custom `webpack` logic now: https://storybook.js.org/docs/react/configure/webpack.

Given that we removed most of the logic in #529, I made this change as well to conform to Storybook standards.

daigof (approved):
Looks good now, thanks for simplifying this repo of configs

PR #579, opened 2020-12-01
Fix tooltip carrot svg

daigof:
- This icon is not part of the Figma Radiance: https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance-UI?node-id=2300%3A9919 . It was created when the tooltip component was overhauled so it doesn't follow the same structure as other formal Radiance Icons and had the hardcoded primary hex color

- This PR cleans up the svg a bit fixes the color bug and tries to replicate the same format as other svgs. 

Can be merged to secondary theme PR or later to master

PR #580, opened 2020-12-01
Fix Arrow svg to support theme

daigof:
context #579
I branched that of secondary-theme PR so it was not clean this merges to master directly, besides I removed one unnecessary class in the svg path

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "bundle.js": {
-    "bundled": 820389,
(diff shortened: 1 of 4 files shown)


michaeljaltamirano (approved):
Thanks for cleaning up the unused classname too. Must've been a Figma copy-paste artifact of some sort. 

Longer-term I think we should document this arrow in Figma, so we should let Ben know that this was moved outside of Tooltip to be its own icon, so that Design can be aware of this usage too.

PR #583, opened 2020-12-02
[TypeScript] Enforce Interfaces over Types

michaeljaltamirano:
`@typescript-eslint/consistent-type-definitions` allows us to prefer one kind of type definition, `Type` or `Interface`. The `interface` rule is flexible enough such that it still permits using `type` for non-object types (e.g. `string`, `number`). However there are performance reasons for wanting to prefer Interface, so much so that it's the first recommendation the TypeScript team makes on their [performance page](https://github.com/microsoft/TypeScript/wiki/Performance#writing-easy-to-compile-code). Since we bundle `radiance-ui` as a dependency it's important for us to keep our declaration files as small as possible. 

**Reviewing this PR is not urgent.**


src/shared-components/button/style.ts
```diff
@@ -162,15 +162,15 @@ function parseTheme(
   }
 }
 
-type BaseButtonStylesTypes = {
+export interface BaseButtonStylesTypes {
```
michaeljaltamirano, line 165: It looks like relying on interfaces instead of types requires us to export the interface, even when they aren't explicitly being used. (Key word is _explicitly_, more on that soon.) If we don't, we get the following error:

```shell
error TS4023: Exported variable 'RoundButtonBase' has or is using name 'BaseButtonStylesTypes' from external module "/Users/michaelaltamirano/radiance-ui/src/shared-components/button/style" but cannot be named.

export const RoundButtonBase = styled(ButtonBase)`
```

Looking at the type output, when we use `type` we are re-defining the type in-line:

Screen Shot 2021-02-10 at 5 59 51 PM

(This is one of the keys for why preferring `interface` results in a much smaller declaration filesize.)

When we use interface, and export it, the type is this simple instead: `Pick<Pick<import("../../style").BaseButtonStylesTypes...` 🙂

.eslintrc
```diff
@@ -25,7 +25,6 @@
     }
   },
   "rules": {
-    "@typescript-eslint/camelcase": "off",
```
michaeljaltamirano, line 28: `CHANGELOG` here:

1. Alphabetize rules
1. Add new rule for PR, `@typescript-eslint/consistent-type-definitions": ["error", "interface"]`. 
    - `eslint --fix` automatically converts `type` definitions to `interface`. Most of this PR was automated 🙂 . 
1. Removes `.js` rules for files that no longer exist.

daigof (approved):
👍 cool. I've been using interface in curology ever since I heard you mention this

src/shared-components/callout/index.tsx
```diff
@@ -5,7 +5,7 @@ import { useTheme } from 'emotion-theming';
 import Style from './style';
 import { COLORS_PROP_TYPES, ThemeColors } from '../../constants';
 
-type CalloutProps = {
+interface CalloutProps {
```
timmartin19, line 8: Not important for this, but I've found that it's really helpful in other libraries when they always export interfaces like all of these.  That way if I want to create a wrapper or something like that I can just do `React.FC<{myProp1: number, subProps: CalloutProps}>` (that's probably a bad example).

timmartin19 (approved):
🙌

PR #584, opened 2020-12-02
Colors organization

daigof:
- Based on: https://www.notion.so/curology/Theme-COLORS-open-questions-af029ea8820f4e2ab302351fc9166079

@benkolde I'd like you to review both themes files
src/constants/colors/primary.ts
src/constants/colors/secondary.ts

Those 2 files will contain all the colors available in Radiance


src/constants/themes/types.ts
```diff
@@ -15,7 +15,7 @@ type Typography =
   | typeof primaryTheme['TYPOGRAPHY']
   | typeof secondaryTheme['TYPOGRAPHY'];
 
-export type ThemeColors = valueof<Colors>;
+export type ThemeColors = valueof<Colors> | '';
```
daigof, line 18: Before this PR there were some colors with empty string value, now that we have the final structure it started to throw error in buttons components where `textColor = ''`

src/constants/themes/types.ts
```diff
@@ -16,6 +16,7 @@ type Typography =
   | typeof secondaryTheme['TYPOGRAPHY'];
 
 export type ThemeColors = valueof<Colors>;
+export type ThemeColorOrEmptyString = ThemeColors | '';
```
daigof, line 19: Before this PR there were some colors with empty string value in the themes, now that we have the final structure with full values, it started to throw errors in buttons components where `textColor` uses empty string as  default value

michaeljaltamirano: It looks like we can safely remove those default arguments and handle them as undefined. I don't think we should be handling empty strings--the types (and usage) should be strict.

michaeljaltamirano (changes requested):
I set up the `secondary-theme` branch separately not to be a long-standing branch like our `epic/add-theming-support` branch. I think this should be a discrete commit since it is a very breaking change (e.g. no more `COLORS` imports available in consumer apps). So we should have in our history two separate PRs merged:

1. Add `secondaryTheme` `COLORS` values. 
1. Remove `COLORS` file and misc. cleanup. 

Feel free to take over and modify https://github.com/curology/radiance-ui/pull/524/files as needed, merge into `master`, and then base this branch on `master`. We will not be doing another `radiance-ui` release until all of this is ready, anyway, so it can all go together in one large release version (`v14`)

src/constants/themes/types.ts
daigof: I initially did that, I found one conditional where it checks for `textColor === ''` 
I didnt wanted to touch the component logic, but I agree with you it will be cleaner to handle as undefined. Ill do this tommorrow when I am fresh.

daigof:
Yeah I will merge secondary-theme separately. I am also fixing the Colors story in sotrybook which is currently broken

src/constants/colors/secondary.ts
```diff
@@ -1,60 +1,52 @@
...
+  successLight: '#ECEFEB',
+  successBorder: '#ECEFEB',
+  error: '#A33F29',
+  errorLight: '#F6ECEA',
+  errorBorder: '#F6ECEA',
+  info: '#2C2A25',
+  infoLight: '#FDF9D3',
+  infoBorder: '#FDF9D3',
+  black: '#2D2926',
+  white: '#FFFEF2',
```
benkolde, line 35: #fffdf2

src/constants/themes/types.ts
daigof: refactored this to use undefined

daigof:
@benkolde for a final review and a more visual one you can use the Storybook here
https://curology-radiance-pr-584.herokuapp.com/?path=/story/constants-constants--colors

You can change the theme via the 2 overlapped squares icon. at the top bar.

I noticed for example that `divider` and `border` have the same value (in primary and secondary theme), so I want to ask you if you want to eliminate `divider` and just use `border`.

stories/constants/colors.stories.tsx
```diff
@@ -1,54 +1,26 @@
 import React from 'react';
 import styled from '@emotion/styled';
-import { SPACER } from 'src/constants';
-import { Typography } from 'src/shared-components';
-import * as ALL_COLORS from 'src/constants/colors';
+import { useTheme } from 'emotion-theming';
 
 import Color from './color';
```
michaeljaltamirano, line 5: I think this story is no longer appropriately nested in `/constants/`, since we're removing the export. We'll also need to update `constants/index.stories.tsx` to update the usage example: `import { BOX_SHADOWS, COLORS, SPACER } from 'radiance-ui/lib/constants';` 

I don't want this to be a blocker, especially since we're relying on Ben to double-check values on that route, so let's make a follow-up item (to be done in a future PR) about re-arranging our stories to reflect usage for v14.

stories/constants/colors.stories.tsx
```diff
@@ -1,54 +1,26 @@
...
+  const theme = useTheme();
+  const colorKeys = Object.keys(theme.COLORS);
 
   return (
     <MainContainer>
-      {colorKeys.map((category) => {
-        const categoryColors = VALID_COLORS[category];
-
+      {colorKeys.map((colorKey) => {
+        const colorHex = theme.COLORS[colorKey] as string;
```
michaeljaltamirano, line 21: This file had particular typing setup to get inference up to the point of needing to cast with `as string`. Since we're only mapping over once now, we no longer need to do that, and should restore the inference we had like so:

Screen Shot 2020-12-03 at 9 31 11 AM

src/shared-components/button/components/roundButton/style.ts
```diff
@@ -107,10 +107,10 @@ const determineAlternateTextColor = (
  */
 const buttonTextColor = (
   buttonColor: ThemeColors,
-  textColor: string,
   theme: ThemeType,
+  textColor?: string,
```
michaeljaltamirano, line 111: I think we should include constraining this type as part of this PR: 
```suggestion
  textColor?: ThemeType['COLORS'],
```

We would also want to update the JSDoc comment above, since `string` is not a valid `textColor` (and it looks like I didn't update it when I converted `color` to `buttonColor`, as well.)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "radiance-ui",
-  "version": "13.2.0",
+  "version": "14.0.0-beta.1",
```
michaeljaltamirano, line 3:
```suggestion
  "version": "13.2.0",
```

.size-snapshot.json
```diff
@@ -1,137 +1,109 @@
 {
   "bundle.js": {
-    "bundled": 821205,
-    "minified": 752699,
-    "gzipped": 90131
+    "bundled": 815565,
```
michaeljaltamirano, line 3: It looks like we'll want to update this again--let's wait until right before merging in so it's not being updated throughout PR fixes.

src/constants/boxShadows/index.ts
```diff
@@ -3,17 +3,18 @@
  */
 
 import { transparentize } from '../../utils';
-import COLORS from '../colors';
+// TODO: Add theme support for box shadow
```
michaeljaltamirano, line 6: 👍

src/constants/colors/index.ts
```diff
@@ -1,186 +0,0 @@
...
-// Dont change these keys -- will affect treatment page guide colors
-export const guideColors = {
-  sunscreenGuide: '#fcf4d5',
-  bigZitsGuide: '#f3d7be',
-  birthControlGuide: '#d0eae5',
-  oilySkinGuide: '#bdc5cb',
-  dietGuide: '#efd5ca',
-  drySkinGuide: '#f3f1a4',
-  makeupGuide: '#f7a9b7',
-  azelaicAcidGuide: '#9ddcc5',
```
michaeljaltamirano, line 146: Is the plan to refer back to this diff when we encounter `LOCAL_COLORS` in usage? I'd also keep an eye out for `resources/assets/jsx/dashboard/components/guideCard/style.ts` which is dynamic and relies on all of these `guideColors`.

benkolde:
> @benkolde for a final review and a more visual one you can use the Storybook here
> https://curology-radiance-pr-584.herokuapp.com/?path=/story/constants-constants--colors
> 
> You can change the theme via the 2 overlapped squares icon. at the top bar.
> 
> I noticed for example that `divider` and `border` have the same value (in primary and secondary theme), so I want to ask you if you want to eliminate `divider` and just use `border`.

Actually yeah, that'd be great

benkolde (approved):

src/constants/colors/primary.ts
```diff
@@ -1,66 +1,52 @@
...
+  errorLight: '#F7EAE8',
+  errorBorder: '#F3DEDC',
+  info: '#554CB3',
+  infoLight: '#EEEDF7',
+  infoBorder: '#E5E3F3',
+  black: '#000000',
+  white: '#FFFFFF',
+  transparent: 'transparent',
 
   // ui colors
```
michaeljaltamirano, line 38: I don't think we're beholden to these code comments anymore and it looks like the UI Toolkit only mentions `Primary`, `Status`, and `Brand` colors. 

If you want to keep some of them that's fine, but I don't think distinguishing between `general` and `ui` is helpful anymore (if it ever was).

michaeljaltamirano (approved):
🥇 This will be much easier to work with in the future!

stories/constants/colors.stories.tsx
daigof: created a task to fix the stories of Colors, box shadows and fonts once they are forced into theme

stories/constants/colors.stories.tsx
daigof: thanks fixed

src/shared-components/button/components/roundButton/style.ts
daigof: refactored it

src/constants/colors/index.ts
daigof: yes, I am aware of the guides portion. I will work with v14 tommorrow or monday and cover all cases using to this PR

src/constants/colors/primary.ts
daigof: yes thanks cleaned up a bit more it looks super sick now

PR #585, opened 2020-12-02
[Theming] Adjust Secondary Theme Button Styling

michaeljaltamirano:
- Secondary Theme button should have larger font (`14px` vs. `12px`).
- Secondary Theme button should have no `font-weight` application (default is `bold`).

### DIFF
--- a/src/shared-components/typography/index.ts
+++ b/src/shared-components/typography/index.ts
@@ -58,10 +58,16 @@ const labelStyle = (theme: ThemeType) => `
 `;
 
 const buttonStyle = (theme: ThemeType) => `
-  ${labelStyle(theme)}
-  font-weight: ${TYPOGRAPHY_CONSTANTS.fontWeight.bold};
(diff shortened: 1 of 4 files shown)


src/shared-components/typography/index.ts
```diff
@@ -58,8 +58,14 @@ const labelStyle = (theme: ThemeType) => `
 `;
 
 const buttonStyle = (theme: ThemeType) => `
-  ${labelStyle(theme)}
-  font-weight: ${TYPOGRAPHY_CONSTANTS.fontWeight.bold};
+  color: ${theme.COLORS.primaryTint1};
+  font-size: ${TYPOGRAPHY_CONSTANTS.fontSize.button};
+  line-height: ${round(20 / 12, 2)};
```
michaeljaltamirano, line 63: First three properties pulled from `labelStyle`, with the updated button font-size. 

The conditional below only adds `bold` to the primary theme.

PR #588, opened 2020-12-03
[Theming] Replace primaryTint3 with textDisabled in Tab

michaeljaltamirano:
Updates the inactive usage. Same value for Primary theme but achieves parity for Secondary theme:

Screen Shot 2020-12-03 at 12 12 42 PM


daigof (approved):
Oh yes go go go. Im thinking on releasing a version 13.3.0 with all these refactors and fixes before releasing the breaking colors v14

PR #589, opened 2020-12-03
[Theming] Apply distinct secondary Display, Heading, and Title font

michaeljaltamirano:
We use `larssiet` for all the `styles` in `src/shared-components/typography/index.ts`. However, we want to use a different font for `Display`, `Heading`, and `Title` for our secondary theme:

image

These changes set the `font-family` on the class, giving it higher precedence than `body`. 

This also removes the direct `TYPOGRAPHY_CONSTANTS` import in `injectGlobalStyles`, which has no impact on output.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "bundle.js": {
-    "bundled": 822671,
-    "minified": 754102,
-    "gzipped": 90261
+    "bundled": 827878,
+    "minified": 759096,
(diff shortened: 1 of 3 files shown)


daigof (approved):
cool, You can release 13.X with this change if you want

michaeljaltamirano:
> cool, You can release 13.X with this change if you want

It's not an urgent item so it can go out with `v14`! 🙂

PR #578, opened 2020-12-01
[POC] Testing Library

michaeljaltamirano:
WIP

PR #609, opened 2020-12-07
[POC] React Testing Library

michaeljaltamirano:
This PR replaces our `react-test-renderer` and `enzyme` test usage with `@testing-library/rfc` as outlined in https://github.com/curology/rfc/pull/77.

Supersedes #578.


src/shared-components/alert/test.tsx
```diff
@@ -14,97 +15,81 @@ const CustomContentComponent = () => (
   </div>
 );
 
-const createNodeMock = (element: ReactElement) => {
```
michaeljaltamirano, line 17: As evidenced by the minimal snapshot changes, this workaround is no longer required.

src/shared-components/alert/test.tsx
```diff
@@ -14,97 +15,81 @@ const CustomContentComponent = () => (
...
         content={<CustomContentComponent />}
         onExit={spy}
         duration="sticky"
       />,
-      { createNodeMock },
     );
 
-    component.root.findByType('button').props.onClick();
-    jest.runAllTimers();
+    assert(container.firstElementChild);
```
michaeljaltamirano, line 64: The DOM type for `firstElementChild` is `Element | null`, so type consistency for `userEvent.click` requires it be non-null. `assert` is a helpful utility added in this PR that automatically handles the type-narrowing + throws an error if our actual code does not our typing of it.

src/shared-components/button/test.tsx
```diff
@@ -10,64 +9,52 @@ import { Button } from './index';
...
-          onClick={() => undefined}
-          id="red-button"
-        >
+      const { container } = render(
+        <Button disabled onClick={() => undefined} icon={<CameraIcon />}>
           Button Text
         </Button>,
       );
 
-      expect(button.children().prop('buttonColor')).toEqual(
```
michaeljaltamirano, line 35: Removed this test because it is testing an implementation detail.

src/shared-components/dialogModal/test.tsx
```diff
@@ -10,13 +8,13 @@ const modalBody = 'Dialog Modal Children Content';
...
-    const wrapper = mount(
+    const { getAllByText } = render(
       <DialogModal title={modalTitle}>
         <div>{modalBody}</div>
       </DialogModal>,
     );
 
-    expect(wrapper.find(ModalTitle).text()).toBe(modalTitle);
-    expect(wrapper.text().includes(modalBody)).toBeTruthy();
+    getAllByText(modalTitle);
```
michaeljaltamirano, line 17: `getBy*` helper methods throw if 0 matches are found, so we don't need to check for truthiness.

src/shared-components/selectorButton/index.tsx
```diff
@@ -76,10 +76,10 @@ export const SelectorButton = ({
         </SelectorIcon>
         <Selector
           type={type}
-          checked={checked}
+          selectorChecked={checked}
```
michaeljaltamirano, line 79: One downside is that our `@testing-library/react` snapshots will remove any invalid HTML attributes we're applying to DOM nodes. In this case I noticed these two props being passed to a div that `react-test-renderer` serialized. We're only using these for styling, and we've been cleaning these up as we encounter them. It might be the case that we clean up fewer of these using `@testing-library/react`, but they're mostly a [nit] anyway so I do not think it should be weighed as a reason to keep `react-test-renderer`.

src/shared-components/banner/test.tsx
```diff
@@ -1,43 +1,38 @@
...
-    const component = renderer.create(<Banner content="Default banner" />);
+    const { container } = render(<Banner content="Default banner" />);
 
-    const tree = component.toJSON();
-    expect(tree).toMatchSnapshot();
+    expect(container.firstElementChild).toMatchSnapshot();
   });
 
   test('Banner with click handler', () => {
-    jest.useFakeTimers();
```
michaeljaltamirano, line 33: This looks like it was copy-pasted from the `Alert` test. There's no `setTimeout` logic involved in this component.

src/shared-components/selectorButton/test.tsx
```diff
@@ -82,11 +74,11 @@ describe('<SelectorButton />', () => {
           </SelectorButton>,
         );
 
-        expect(tree).toMatchSnapshot();
+        expect(container.querySelectorAll('svg').length).toBe(0);
```
michaeljaltamirano, line 77: This test, I think, suffers from the problem of unhelpful snapshots. The `it` block is focusing on a very specific part of the test, but it is serializing everything as a snapshot. That's not super clear. This just takes advantage of `container` being a DOM node to use DOM API to check if there's an SVG present in the markup.

src/shared-components/dropdown/test.tsx
```diff
@@ -17,52 +16,48 @@ describe('<Dropdown />', () => {
...
     it('renders <MobileDropdown />', () => {
       window.document.documentElement.ontouchstart = () => undefined;
-      const wrapper = mount(
+      const { getByRole, getAllByRole, queryAllByRole } = render(
         <Dropdown value="test1" options={options} onChange={() => undefined} />,
       );
       delete window.document.documentElement.ontouchstart;
 
-      expect(wrapper.children().children().first().name()).toEqual(
-        'MobileDropdown',
```
michaeljaltamirano, line 26: Both of these tests are, to an extent, testing implementation details. The `@testing-library/react` assertions compare the differences in the markup between Mobile and Desktop, though. The enzyme version is testing the React component name, but not what it actually looks like.

src/shared-components/selectorButton/test.tsx
michaeljaltamirano: Currently, we strip the `<title>` from our `SVGs`. If we were to change our [`svgr` settings](https://react-svgr.com/docs/options/#title) to keep the title, we could use the built-in `getByTitle` helper method and check for the presence of the icon name we are passing in.

daigof (approved):
I am not familiar with the new library, but this looks way better than before. 👍

PR #614, opened 2020-12-09
[Theme] Fonts and Typography cleanup

daigof:
- Removed the index files
- removed typography test 
- To be released as v15 along with boxShadows work
- removed fonts and typography fro the constants story (a future PR will deal with maybe a new theme based story?)


michaeljaltamirano (approved):
Can you updated the size snapshot, too?

PR #616, opened 2020-12-09
[Theming] Add Secondary Nav Icons

michaeljaltamirano:
### What & Why

This PR adds the Nav Icons for our secondary theme. 

It sets `theming/secondary-icons` as the base branch, which will serve as a long-standing branch for all of the icons changes in progress, after which we can merge into `master`, so that we do not hold up any releases that need to go out. I'd prefer to release the icons as one large body of work rather than have small changes throughout, in large part because I intend to make breaking changes, but also to be able to review the icon release as a whole instead of each respective change. 

### Next Up
- [x] Secondary Nav Icon
- [ ] Secondary Icons
- [ ] Secondary Glyphs
- [ ] Secondary Emoji
- [ ] Secondary Wordmark


src/icons/test.ts
```diff
@@ -35,9 +35,9 @@ const testIconDirectory = (directory: string) => {
       path.resolve(__dirname, `${directory}/${SVG_DIRECTORY_NAME}`),
     );
 
-    const associatedSvgIconFilenames = Array.from(new Set(svgFilenames))
-      .map(transformIconFileNames)
-      .sort();
+    const associatedSvgIconFilenames = Array.from(
+      new Set(svgFilenames.map(transformIconFileNames).sort()),
```
michaeljaltamirano, line 39: The uniqueness application (`Array.from(new Set([]))`) should've been happening after the transform, but the test didn't value when initially added because there wasn't a difference yet.

daigof (approved):
👍

PR #623, opened 2020-12-10
[Theming] Add Secondary Glyphs

michaeljaltamirano:
### What & Why

This PR adds the Glyphs for our secondary themes.

Unlike the Nav Icons in #616, there are some new icons here that are present **only** for the secondary theme. I've added new handling to our Icon `useIcon` logic to return null for the primary theme when that is the case. The Primary Icons story is unchanged, because I added some additional logic around that rendering to prevent adding secondary theme glyphs to the list we map. 

**There is a breaking change** in the naming of our `CalendarAlways` through `CalendarNever` icons. I've renamed these to just be `Always`, `Never`, etc. There is still the `TimeAlways` through `TimeNever` icons, because both themes use clocks in the image, so the naming makes sense. Since the `Calendar`-version of the Secondary themes is a sand dial, the name `Calendar` is inappropriate. I've left a comment on Figma asking about renaming as follows, and will update the PR depending on the outcome of the question: `CalendarAlways` --> `TimeAlways` (encompassing both Calendar and Sand Dial glyphs), and `TimeAlways` --> `ClockAlways` (since both glyphs use a clock of sorts.)

### Next Up
- [x] Secondary Nav Icon
- [ ] Secondary Icons
- [x] Secondary Glyphs
    - [ ] **TODO IN CONSUMER USAGE**: Rename `Calendar`-prefixed Glyphs in consumer apps with `Time` (e.g. `TimeAlwaysGlyph`) and `Time`-prefixed Glyphs with `Clock` (e.g. `ClockAlwaysGlyph`). 
- [ ] Secondary Emoji
- [ ] Secondary Wordmark


src/icons/glyphs/antiAgingAndAcneGlyph.ts
```diff
@@ -0,0 +1,5 @@
+import { AntiAgingAndAcneGlyphSecondary } from './svgs';
+import { useIcon, IconProps } from '../../shared-components/icon';
+
+export default (props: IconProps) =>
+  useIcon(null, AntiAgingAndAcneGlyphSecondary, props);
```
michaeljaltamirano, line 5: Since there is no corresponding primary icon, we pass in `null` here, and add a `null` check to the body of useIcon to return null if there is no icon.

daigof (approved):
I dont like the idea of returning null for icons based on the theme type. I dont know what was discussed with Ben, but I imagine a page in curology app where we have this new icon used and the primary theme will displace all the layout. Why not  just return the same icon for both themes exactly like we are doing if there is no secondary icon present?

Maybe this is just temporary and Design will eventually provide the primary icon missing so we have no null returns in prod?

michaeljaltamirano:
> I dont like the idea of returning null for icons based on the theme type. I dont know what was discussed with Ben, but I imagine a page in curology app where we have this new icon used and the primary theme will displace all the layout. Why not just return the same icon for both themes exactly like we are doing if there is no secondary icon present?
> 
> Maybe this is just temporary and Design will eventually provide the primary icon missing so we have no null returns in prod?

I think the idea is that we do not want to leak secondary theme icons into primary theme usages. These should (I assume) only be used on sub-domains exclusive to the secondary theme, so this scenario you mention should not apply. (There are lots of shoulds here, of course.) In any case we can re-arrange the logic later on if this seems sub-optimal.

PR #624, opened 2020-12-10
[Theming] Add Secondary Emoji

michaeljaltamirano:
### What & Why

This PR adds the Emoji for our secondary theme.

There's a 1:1 mapping of emoji so there are no null returns.

**There is a breaking change** in the naming of our `AcneEmoji`. It looks like we map these to `Irritation` in consumer apps, so I've gone ahead and renamed them as `Irritation` and will have to make the corresponding change in consumer apps, too.

### Next Up
- [x] Secondary Nav Icon
- [ ] Secondary Icons
- [x] Secondary Glyphs
    - [ ] **TODO IN CONSUMER USAGE**: Rename `Calendar`-prefixed Glyphs in consumer apps with `Time` (e.g. `TimeAlwaysGlyph`) and `Time`-prefixed Glyphs with `Clock` (e.g. `ClockAlwaysGlyph`). 
- [x] Secondary Emoji
    - [ ] **TODO IN CONSUMER USAGE**: Rename `Acne` emoji in consumer apps with `Irritation` (e.g. `MildAcneEmoji` --> `MildIrritationEmoji`). 
- [ ] Secondary Wordmark
    - This change should also reflect that when we pass in custom height and width, we should update the `viewBox` accordingly, which we do not currently do.


daigof (approved):
👍

PR #625, opened 2020-12-11
Theme Story

daigof:
it may not be perfect but it's something

Review App: http://curology-radiance-pr-625.herokuapp.com/?path=/story/constants-theme--box-shadows

this is the not SSL HTTP version since the https gets cross domain errors


stories/theme/index.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+  parameters: {
+    chromatic: { viewports: [BREAKPOINTS.xs] },
+    docs: {
+      page: () => (
+        <React.Fragment>
+          <Title />
+          <Description />
+          <Heading>Usage:</Heading>
+          <Description>
+            The `theme` object is accessible (in theme aware apps using
```
michaeljaltamirano, line 36: I think we want to make this description as comprehensive as possible (without running the risk of going out-of-date), so that someone with no context of emotion can look to this page and see all the ways we use it in our apps. 

For usage, we should cover the following scenarios:

1. `styled` as prop (currently covered)
1. `css` as function parameter (currently covered)
1. `withTheme/useTheme`
    - The three above usages can probably just point to the docs: https://emotion.sh/docs/theming#usage
1. `<Global styles={(theme) => {}} />`, the docs note "It accepts a styles prop which accepts the same values as the css prop except it inserts styles globally": https://emotion.sh/docs/globals

I would also considering pointing to the docs to show what a "theme-aware app" is (code wrapped in a `ThemeProvider` that has provided a `theme`: https://emotion.sh/docs/theming#themeprovider-reactcomponenttype)

stories/theme/index.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+export * from './fonts.stories';
+export * from './typography.stories';
+
+const styledComponentSource =
+  // eslint-disable-next-line no-template-curly-in-string
+  'export const StyledComponent = styled.div`\n  color: ${({ theme }) => theme.COLORS.primary};\n`';
+
+const cssPropSource =
+  // eslint-disable-next-line no-template-curly-in-string
+  '<Component css={(theme: ThemeType) => css`color: ${theme.COLORS.primary}`}>';
```
michaeljaltamirano, line 23: I think we can remove the in-line typing to reduce confusion. I don't think our other code usage examples include types that aren't present in the source.

stories/theme/index.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export default {
+  title: 'Constants/Theme',
+  parameters: {
+    chromatic: { viewports: [BREAKPOINTS.xs] },
```
michaeljaltamirano, line 28:
```suggestion
    chromatic: { disable: true },
```

Since we'll effectively be evaluating the theme in every component story, I think we can skip snapshot testing on the actual theme stories themselves. 

I kept the story for the `Available Constants` in Chromatic since it groups everything on one big page, but since we're (wisely) breaking up the stories into smaller chunks for the theme, it'll push us closer to the limits of our Open Source snapshot allotment every month, so it's probably best to remove redundant visual tests.

michaeljaltamirano (approved):
Thank you for getting to this! A nice Friday documentation treat 🙂

stories/theme/index.stories.tsx
daigof: our code does not include the ThemeType => that's true. I discovered I could type it after I did the Colors refactor. I will remove it from here yes I agree. I guess I added so people know they can use the typing otherwise the theme will have `any` type and they can make typos (I know I will use it in some cases)

stories/theme/index.stories.tsx
daigof: yep revamped that section a bit, could be updated further in next prs

PR #628, opened 2020-12-11
[Theming] Add Secondary Icons

michaeljaltamirano:
### What & Why

Long-standing feature branch that adds new icons (icons, emoji, glyphs, and nav icons) from PRs #616, #623, #624, and #627. 

Since there are breaking changes documented in those PRs, I will cut a new major version with the icons changes for our next release. I am working on another PR that standardizes icon sizing before doing that (#629), though. 

There's no visual regression beyond what's already been approved: shifting icons due to name changes + new icons: https://www.chromatic.com/pullrequest?activeElementId=comparison-test-key-5f903d228f8e91dd33ae03db-330&appId=5f6e46f1e7081000221523cf&number=628&view=changes

I will probably do a beta release in the event that I am tagging this PR for review too late before the end of the week.

### What & Why

This PR adds the Icons (specifically, the "Icon" subset of our Icons) for our secondary theme.

There are **four** new icons: `AppleIcon`, the company logo, and three Secondary theme-exclusive icons: `GraphIcon`, `LightbulbIcon`, and `MoonIcon`. 

**There is a breaking change** in the naming of our `Arrow` icon. It was added as just `Arrow` and so did not conform to our other Icon naming. It is now named `ArrowIcon`. 

### Next Up
- [x] Secondary Nav Icon
    - [x] TODO: Clarify naming of `Camera` Nav Icon since it is listed as `Progress` in the Style Guide UI but named with `Camera` as the actual Figma filename. 
        - Leaving this as is for now, not as important as the other name changes. 
- [x] Secondary Icons
    - [ ] **TODO IN CONSUMER USAGE**: Rename `Arrow` icon usage in Consumer apps to `ArrowIcon`. 
- [x] Secondary Glyphs
    - [ ] **TODO IN CONSUMER USAGE**: Rename `Calendar`-prefixed Glyphs in consumer apps with `Time` (e.g. `TimeAlwaysGlyph`) and `Time`-prefixed Glyphs with `Clock` (e.g. `ClockAlwaysGlyph`). 
- [x] Secondary Emoji
    - [ ] **TODO IN CONSUMER USAGE**: Rename `Acne` emoji in consumer apps with `Irritation` (e.g. `MildAcneEmoji` --> `MildIrritationEmoji`). 
- [x] Secondary Logos
    - Doing this at a later date. 
- [x] **BREAKING CHANGE** Remove icon height and width default prop of `16`
[...]

PR #629, opened 2020-12-11
[Theming] Icons Height/Width Standardization

michaeljaltamirano:
### What & Why

Our style guide is prescriptive about icon sizing: regular icons should be 16x16, navigation icons 20x20, emoji and glyphs 48x48. We currently rely on `viewBox` for those dimensions, which does not always translate perfectly into usage. As a result, we are hard-coding `height` and `width` at the default sizes when we should not need to. 

I've accepted the visual changes as the new baseline, which is just icons shifting due to the Logo and Wordmark not receiving custom dimensions in the icons story. 

### CHANGELOG

1. Pass in custom `svgr` `svgoConfig` option to maintain `viewBox` if `height` and `width` are present on `svg`: https://github.com/curology/radiance-ui/pull/629/files#r541256383
1. Pass in `height` and `width` according to the specifications outlined above (16, 20, and 48 px height/width). 
1. Remove instances in `src`/`stories` where we were passing in the same values are the defaults. 
    - Removes the default values of `16` for height and width in `src/shared-components/icon/index.tsx`, which was probably why folks had to specify so many default properties in the first place, since 16 only applies to icon icons and none of the other icon types. 

This branch is currently based against `theming/secondary-icons` but can be updated to `master` after #628 is merged in, after which we can release `v16`.


stories/icons/index.stories.tsx
```diff
@@ -43,33 +41,46 @@ WithControls.parameters = {
...
+          </Style.IconContainer>
+        );
+      })}
     </Style.Icons>
 
     <Typography.Heading>Emojis</Typography.Heading>
     <Style.Icons>
       {Object.entries(emojiComponents).map(([name, Icon]) => (
         <Style.IconContainer key={name}>
-          <Icon width={48} height={48} />
```
michaeljaltamirano, line 72: Since we are defining this within individual SVGs, we no longer need to explicitly pass in these values in usage (when they match the default).

package.json
```diff
@@ -24,11 +24,11 @@
     "build": "run-s clean && run-s build:icons:all && run-s tsc && run-p build:rollup build:babel",
     "build:icons:all": "run-s clean && run-s build:icons build:icons:index",
     "build:icons": "run-p build:icons:emojis build:icons:glyphs build:icons:icons build:icons:logos build:icons:navicons",
-    "build:icons:emojis": "svgr --typescript src/svgs/emojis --out-dir src/icons/emojis/svgs",
-    "build:icons:glyphs": "svgr --typescript src/svgs/glyphs --out-dir src/icons/glyphs/svgs",
-    "build:icons:icons": "svgr --typescript src/svgs/icons --out-dir src/icons/icons/svgs",
-    "build:icons:logos": "svgr --typescript src/svgs/logos --out-dir src/icons/logos/svgs",
-    "build:icons:navicons": "svgr --typescript src/svgs/navIcons --out-dir src/icons/navIcons/svgs",
+    "build:icons:emojis": "svgr --typescript --svgo-config='{\"plugins\": {\"removeViewBox\": false}}' src/svgs/emojis --out-dir src/icons/emojis/svgs",
```
michaeljaltamirano, line 27: `svgr`, courtesy of its underlying use of `svgo`, automatically removes `viewBox` if `height` and `width` are provided. We want to keep both, so that when a custom `height` and `width` is provided the image scales appropriately. 

https://react-svgr.com/docs/options/#svgo-config

src/icons/logos/svgs/WordmarkPrimary.tsx
```diff
@@ -3,7 +3,9 @@ import * as React from 'react';
 function SvgWordmarkPrimary(props: React.SVGProps<SVGSVGElement>) {
   return (
     <svg
-      viewBox="0 0 112 26"
+      width={112}
+      height={25}
+      viewBox="0 0 112 25"
```
michaeljaltamirano, line 8: Matches the spec in Figma

src/shared-components/optionButton/__snapshots__/test.tsx.snap
```diff
@@ -264,9 +264,9 @@ exports[`<OptionButton /> UI snapshots radio unselected, with icon node prop 1`]
       <svg
         className="emotion-0"
         fill="none"
-        height={16}
+        height={48}
         viewBox="0 0 48 48"
-        width={16}
+        width={48}
```
michaeljaltamirano, line 269: This is not a regression: we were passing in `AcneGlyph` without specifying `height`/`width` so it took on the default values of `16`.

src/icons/icons/svgs/AppleIcon.tsx
```diff
@@ -3,8 +3,9 @@ import * as React from 'react';
 function SvgAppleIcon(props: React.SVGProps<SVGSVGElement>) {
   return (
     <svg
-      width={14}
+      width={16}
       height={16}
+      viewBox="0 0 14 16"
```
daigof, line 8: this looks inconsistent with all the other usages, we should try to get from design a 16x16 svg

src/icons/icons/svgs/ArrowIcon.tsx
```diff
@@ -3,6 +3,8 @@ import * as React from 'react';
 function SvgArrowIcon(props: React.SVGProps<SVGSVGElement>) {
   return (
     <svg
+      width={16}
+      height={16}
       viewBox="0 0 46.29 32.92"
```
daigof, line 8: I guess this can be the only exception, I still think we should move this out of our Icons library. I'll take a look after v16 is released

src/icons/icons/svgs/PlanIcon.tsx
```diff
@@ -3,6 +3,8 @@ import * as React from 'react';
 function SvgPlanIcon(props: React.SVGProps<SVGSVGElement>) {
   return (
     <svg
+      width={16}
+      height={16}
       viewBox="0 0 14 16"
```
daigof, line 8: another case?

src/icons/icons/svgs/PlayIconSecondary.tsx
```diff
@@ -3,6 +3,8 @@ import * as React from 'react';
 function SvgPlayIconSecondary(props: React.SVGProps<SVGSVGElement>) {
   return (
     <svg
+      width={16}
+      height={16}
       viewBox="0 0 17 16"
```
daigof, line 8: marking so I can list them all

daigof (approved):
Thanks I love the standarization and simpler  usage. Found 3 icons that are not perfect 16x16 I think we should ask Design for perfect 16x16 cases.

src/icons/icons/svgs/AppleIcon.tsx
michaeljaltamirano: `viewBox="0 0 14 16"` helps keep the icon centered. Here's an example of the PlanIcon which has the same dimensions. 

Screen Shot 2020-12-14 at 9 24 25 AM

I think the `viewBox="0 0 16 16"` deviations--since they're pulled directly from Figma--are fine to keep unless QA in usage notes that the positioning looks weird.

src/icons/icons/svgs/ArrowIcon.tsx
michaeljaltamirano: I left a comment in Figma about this for Ben to resolve last week. If you'd like to bring it up with him again you're welcome to but I consider the matter low-priority.

daigof: Yeah saw that, I actually was always thinking on removing from public Radiance. I added this icon and Ida created it for me after trying JPG and PNGs. I don't think design intention was to put it here ever, it was my bad to be able to use the SVGR generation. Ill take a look in Cooldown,

daigof: I know there is directly usage in Curology Ill figure something out

PR #642, opened 2020-12-15
[Theme] - Border Radius

daigof:
- Review APP: http://curology-radiance-pr-642.herokuapp.com (some components to look for: Modals, Tooltips, Dropdown, Field, Buttons, Alerts, option button, Indicator, Chip)

- Refactor all full circles radius to 50% for consistency:  https://www.codecademy.com/forum_questions/559fe347e39efe4cf40005a9

- Change Container message radius from 16px to theme medium (8px)

- Named exports for some constants (e.g. SPACER) for better intellisense and consistency

- added Border Radius story 

- Changed  Toggle focus border radius to theme large: was hardcoded 12px (beyond 12px the rounded visual is the same so no impact in primary theme). This affects secondary theme, which focus as square
image

it is more aligned to the theme but let me know if this needs to be changed back and hardcoded as rounded focus

### DIFF
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useState, useRef } from 'react';
 import { useTheme } from 'emotion-theming';
 
+import { ThemeType } from '../../constants';
 import { ChevronIcon } from '../../icons';
 import Thumbnails from './thumbnails';
 import {
@@ -14,9 +15,15 @@ import {
   Truncate,
 } from './style';
 
+export type BorderRadiusValues =
+  | valueof<ThemeType['BORDER_RADIUS']>
+  | '0.25rem'
+  | '0.5rem'
+  | '2rem';
+
 type AccordionProps = {
   /** Sets the border-radius of Accordion.Container, AccordionBox, and TitleWrapper */
-  borderRadius?: string;
+  borderRadius?: BorderRadiusValues;
   /** node(s) that will render only when expanded */
   children: React.ReactNode;
   /** when true, the accordion will be greyed out and the onClick prop will be disabled */
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -4,6 +4,7 @@ import { useTheme } from 'emotion-theming';
 
 import { ChevronIcon } from '../../icons';
 import Thumbnails from './thumbnails';
+import { BORDER_RADIUS } from '../../constants';
 import {
   AccordionBox,
   ArrowWrapper,
@@ -41,7 +42,7 @@ type AccordionProps = {
  * The accordion component expands to reveal hidden information. They should be used when you need to fit a large amount of content but don't want to visually overwhelm the user.
  */
 export const Accordion = ({
-  borderRadius = '4px',
+  borderRadius = BORDER_RADIUS.small,
   children,
   disabled = false,
   isOpen,
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -55,8 +55,6 @@ export const Accordion = ({
 
   const contentRef = useRef<HTMLDivElement>(null);
 
-  const borderRadiusValue = borderRadius || theme.BORDER_RADIUS.small;
-
   useEffect(() => {
     const nextHeight =
       isOpen && contentRef.current
@@ -82,7 +80,7 @@ export const Accordion = ({
             onClick(event);
           }
         }}
-        borderRadius={borderRadiusValue}
+        borderRadius={borderRadius}
         onKeyDown={handleKeyDown}
         disabled={!!disabled}
         role="button"
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -4,7 +4,6 @@ import { useTheme } from 'emotion-theming';
 
 import { ChevronIcon } from '../../icons';
 import Thumbnails from './thumbnails';
-import { BORDER_RADIUS } from '../../constants';
 import {
   AccordionBox,
   ArrowWrapper,
@@ -42,7 +41,7 @@ type AccordionProps = {
  * The accordion component expands to reveal hidden information. They should be used when you need to fit a large amount of content but don't want to visually overwhelm the user.
  */
 export const Accordion = ({
-  borderRadius = BORDER_RADIUS.small,
+  borderRadius,
   children,
   disabled = false,
   isOpen,
@@ -55,6 +54,8 @@ export const Accordion = ({
   const [contentHeight, setContentHeight] = useState('0px');
 
   const contentRef = useRef<HTMLDivElement>(null);
+
+  const borderRadiusValue = borderRadius || theme.BORDER_RADIUS.small;
 
   useEffect(() => {
     const nextHeight =
@@ -81,7 +82,7 @@ export const Accordion = ({
             onClick(event);
           }
         }}
-        borderRadius={borderRadius}
+        borderRadius={borderRadiusValue}
         onKeyDown={handleKeyDown}
         disabled={!!disabled}
         role="button"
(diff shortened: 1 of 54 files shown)


src/shared-components/accordion/style.ts
```diff
@@ -98,30 +98,34 @@ export const Container = styled.div<{
   background-color: ${({ theme }) => theme.COLORS.white};
   max-width: ${BREAKPOINTS.md}px;
 
-  ${({ borderRadius = '4px' }) => `
+  ${({ borderRadius, theme }) => {
+    const borderRadiusValue = borderRadius || theme.BORDER_RADIUS.small;
```
michaeljaltamirano, line 102: Since this PR is concerned with implementing the style guide prescription for how `border-radius` should operate, I think we should define an enum for what the acceptable `border-radius` values are, in order to actually enforce it. 

It could get thorny because most user-land usage is probably in `rem` but we're using mostly `pixels` in this PR, but something like this might work:

`type BorderRadiusValues = valueof<ThemeType['BORDER_RADIUS']> | '0.25rem' | '0.5rem' | '2rem'`

src/shared-components/accordion/index.tsx
```diff
@@ -80,7 +82,7 @@ export const Accordion = ({
             onClick(event);
           }
         }}
-        borderRadius={borderRadius}
+        borderRadius={borderRadiusValue}
```
michaeljaltamirano, line 85: I think, wherever possible, it might be preferable to kick this `borderRadius` to `borderRadiusValue` logic into the `styled` component to hide it from the component logic markup. I know i've been guilty of this in the past with default colors, and we're already pulling in `theme` for the icon fill, but only after seeing more and more examples have I realized a preference against it 😅. I think we should save it for when we need to use it, like the Icon cases.

src/shared-components/button/style.ts
```diff
@@ -171,7 +171,7 @@ export const baseButtonStyles = ({
 }: BaseButtonStylesTypes) => `
   ${TYPOGRAPHY_STYLE.button(theme)}
   appearance: none;
-  border-radius: ${SPACER.xsmall};
+  border-radius:  ${theme.BORDER_RADIUS.small};
```
michaeljaltamirano, line 174: [Nit] The spacing looks over here--curious that `prettier` didn't fix it

stories/theme/borderRadius.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import styled from '@emotion/styled';
+import { useTheme } from 'emotion-theming';
+import { SPACER } from 'src/constants';
+
+const MainContainer = styled.div`
+  padding: ${SPACER.large} ${SPACER.large};
+`;
+
+export const BorderRadius = () => {
```
michaeljaltamirano, line 10: It would be nice to make this story more like `COLORS`, `BOX_SHADOWS`, etc., that shows the differences in action, rather than just shows the text values.

src/shared-components/container/style.ts
```diff
@@ -15,7 +15,7 @@ const clickableStyle = (theme: ThemeType) => `
 `;
 
 const messageStyle = (theme: ThemeType) => `
-  border-radius: 16px;
+  border-radius: ${theme.BORDER_RADIUS.medium};
```
michaeljaltamirano, line 18: I am okay with this change but it is probably worth double-checking with Ben that this does not conflict with what is present throughout Design prototypes in Figma: https://www.chromatic.com/test?appId=5f6e46f1e7081000221523cf&id=5fd8d768e3a0560021add39c

src/shared-components/dropdown/index.tsx
```diff
@@ -45,15 +46,17 @@ type DropdownProps<T> = {
...
   onChange,
   options,
   optionsContainerMaxHeight = '250px',
   textAlign = 'left',
   value,
 }: DropdownProps<T>) => {
+  const theme = useTheme();
   const [isOpen, setIsOpen] = useState(false);
   const touchSupported = 'ontouchstart' in document.documentElement;
+  const borderRadiusValue = borderRadius || theme.BORDER_RADIUS.small;
```
michaeljaltamirano, line 59: My previous comments about preferring to keep the logic in `styled` components doesn't apply here, since we wouldn't want to duplicate ourselves 🙂

src/shared-components/optionButton/index.tsx
```diff
@@ -40,41 +41,46 @@ export const OptionButton = ({
...
-      <TextContainer>
-        <Text>{text}</Text>
-        {subtext && <SubText>{subtext}</SubText>}
-      </TextContainer>
-    </FlexContainer>
-  </ClickableContainer>
-);
+}: OptionButtonProps) => {
+  const theme = useTheme();
+  const borderRadiusValue = borderRadius || theme.BORDER_RADIUS.small;
```
michaeljaltamirano, line 46: Let's kick this to the styled component logic--it'll keep the diff smaller, too.

benkolde (commented):
@daigof I see two ways forward with the toggle

Either square off the toggle, or round the focus. I think I'd like to round the focus for now, and then revisit the design later

michaeljaltamirano:
> @daigof I see two ways forward with the toggle
> 
> Either square off the toggle, or round the focus. I think I'd like to round the focus for now, and then revisit the design later

@daigof The toggle is a dependency (`react-toggle-button`) that is probably our biggest tech debt in this repo. Let's keep the focus by hard-coding for now. We can add the appropriate `// TODO` or JSDoc comments to memorialize the hard-coding as appropriate.

michaeljaltamirano (approved):
Approving without requesting changes but would like to see the following before merging this in:

1. Keep the border-radius for `Toggle` and add `TODO`s as needed
1. Move more `border-radius` theming logic into the `styled` component logic instead of the main component body.
1. Updating the `BORDER_RADIUS` story to show divs (containers, etc.) that show the `border-radius` in effect. 
1. Enforcing a strict set of valid `border-radius` values so that we can guide users in the user-land to style guide-approved happy paths. (e.g. `type BorderRadiusValues = valueof<ThemeType['BORDER_RADIUS']> | '0.25rem' | '0.5rem' | '2rem'`).

I've accepted the Container visual regressions (for `messageType`) as the new baseline so future commits should pass that part of the `build` step. 

After those changes, we should be good to go with releasing a minor version. Nothing is breaking though the stricter type might requires some refactors in user-land.

src/shared-components/button/style.ts
daigof: done

stories/theme/borderRadius.stories.tsx
daigof: done, also refactored other stories for consistency (i.e. same initial padding in page)

src/shared-components/container/style.ts
daigof: this was approved by Ben, in my opinion it looks even better with 8px

src/shared-components/optionButton/index.tsx
daigof: done

daigof:
@michaeljaltamirano I did 1-3 items, but I disagree with number 4). Historically you can just pass whatever value here it was never mentioned to be strict to 4/8/32. So you could theoretically pass down a 16px or whatever you like. I don't understand why start enforcing strict values now

michaeljaltamirano:
> @michaeljaltamirano I did 1-3 items, but I disagree with number 4). Historically you can just pass whatever value here it was never mentioned to be strict to 4/8/32. So you could theoretically pass down a 16px or whatever you like. I don't understand why start enforcing strict values now

We do the same thing for `COLORS`, etc. The component library should inform (and, courtesy of TypeScript, enforce) how our common components are used. If something is outside of the use case, then it should be explicitly documented in both the library and design style guides. 

I did not request changes because I don't think it is mandatory, but I think it is something we need to be much stricter about--especially as the team grows--so that we do not end up with inconsistent code patterns in user-land.

daigof:
ok I enforced the new type

benkolde (approved):

PR #651, opened 2020-12-16
[Theme] - Change Primary Buttons colors for Secondary Theme

daigof:
- Taken from https://www.notion.so/curology/Audit-of-Staging-Environment-0394febd9b934132887ecf6449fedfc8
- and Figma https://www.figma.com/file/FjOFrQaYVSdlaMIhpgR192/Ruth-UI?node-id=2300%3A18669&viewport=480%2C520%2C0.8739789724349976

- I tested other Custom colors for `buttonColor` prop and none look good for secondary theme. I dont know if design team is aware of this custom prop, I think we should explore the posibility to remove that Button prop and treat the special case in the page itself and not in Radiance the only valid usage for it I found was in dashboard modular cancellation flow, can be easily reworked. The other usages were in Admin pages which could be refactored without much design impact

- Review App: https://curology-radiance-pr-651.herokuapp.com/?path=/story/components-button-button--default

# Disabled Button Review

@benkolde if we change disabled button to tertiaryTint3 background there is a visual change in Primary Theme

### PRIMARY THEME with `tertiaryTint3 background  #F8F8FA`
image

### PRIMARY THEME with `defaultLight background  #EDEDF0` current in PROD
image

## Options:
1- Accept the change for Primary Theme
2- Add a conditional and use different background colors for themes
3- Accept to use the current defaultLight background color and accept the change to secondary theme (barely noticeable)

### SECONDARY THEME with `tertiaryTint3 background  #F2F0EF`
image

### SECONDARY THEME with `defaultLight background  #F0EEEC`
image


benkolde (commented):
🙌 can we remove the border from the disabled buttons? only thing I noticed. These look great!

daigof:
Oh I forgot to do/review the disabled comment, yeah I will take a look at it

daigof:
@benkolde can you review the Disabled button section and choose 1/2/3 option please. (easiest less impactful option would be 3 for me)

benkolde:
> @benkolde can you review the Disabled button section and choose 1/2/3 option please. (easiest less impactful option would be 3 for me)

@daigof I'm fine with 3, is it not easy to remove the border between both themes?

daigof:
@benkolde the border was never an issue / concern.  You probably didn't noticed but in the review app secondary theme has a background of #F0EEEC (defaultLight) instead of what we have in Figma #F2F0EF (teritaryTint3) the diference is not noticeable but if we change to tint3 we are creating a visual regression for primary theme. so It is better to just use defaultLight for both

benkolde:
@daigof ahh i see. Okay yeah, number 3 is good

src/shared-components/button/style.ts
```diff
@@ -113,7 +129,7 @@ const loadingStyles = `
 
 const disabledStyles = (theme: ThemeType) => `
   background-color: ${theme.COLORS.defaultLight};
-  border-color: ${theme.COLORS.border};
+  border-color: ${theme.COLORS.defaultLight};
```
daigof, line 132: no visual change for primary theme since have same hex code

benkolde (approved):

src/shared-components/button/shared-components/loader/style.ts
```diff
@@ -13,7 +13,9 @@ const statefulLoader = keyframes`
 `;
 
 const primaryLoadingStyles = (theme: ThemeType) => `
-  background-color: ${theme.COLORS.white};
+  background-color: ${
+    theme.__type === 'primary' ? theme.COLORS.white : theme.COLORS.primary
```
michaeljaltamirano, line 17: If we end up having more exceptions like this, we might want to create a separate file that contains helper methods that contains all the exceptions (e.g. `primaryButtonLoadingBackgroundColor = (theme) => {}` and then embed it). It might be worth clarifying with Ben if we expect more, or adding a TODO for tracking purposes, but this works for now!

src/shared-components/button/style.ts
```diff
@@ -7,21 +7,37 @@ import { textColorsAssociatedWithColors } from './constants';
 
 import { ButtonTypeWithAction } from '.';
 
-const primaryStyles = (buttonColor: ThemeColors, theme: ThemeType) => `
-  background-color: ${buttonColor};
-  border-color: ${buttonColor};
-  color: ${theme.COLORS.white};
-  fill: ${theme.COLORS.white};
+const primaryStyles = (buttonColor: ThemeColors, theme: ThemeType) => {
+  let backgroundColor = buttonColor;
```
michaeljaltamirano, line 11: Made my [previous comment](https://github.com/curology/radiance-ui/pull/651/files#r544659164) before seeing more bespoke logic. I think we should move logic like this (that breaks the normal 1:1 theme mapping of styling) into a separate file, or if we use a special convention for how to name these exceptions so that they're easily findable across the application. (I think the separate file is preferable, because it's more easily identifiable, ideally with a big JSDoc comment explaining the intention of the file). 

I imagine that there will be more exceptions as we catch them in review so we might as well start with this PR.

michaeljaltamirano (changes requested):
No concerns about the actual styling but would like to see your thoughts on [how to better organize 1:1 theme mapping exceptions](https://github.com/curology/radiance-ui/pull/651/files#r544699030) in this repo.

src/shared-components/button/shared-components/loader/style.ts
daigof: Yeah that's a good idea, we have very cases in Radiance but we can do as preparation for PocketDerm considering C8 will introduce a bunch of those in Curology

src/utils/themeStyles/index.ts
```diff
@@ -0,0 +1,31 @@
+import { ThemeColors, ThemeType } from '../../constants';
+
+export const primaryButtonFontColor = (theme: ThemeType) =>
```
michaeljaltamirano, line 3: This is great! I wonder if we should add a comment at the top of the file that explains that any ternary that is checking `theme.__type` should probably be moved into a util function in this file? (With the one exception being the Icon component, as you left it).

This isn't not a blocking comment--thank you for this cleanup 🙂

michaeljaltamirano (approved):
🥳

PR #666, opened 2020-12-30
[Styling] Remove Superfluous css Serialization to Reduce Size

michaeljaltamirano:
### What & Why

We have some instances of in-lining `css` serialization within other `css` serializations and `styled` serializations, which are unnecessary and contribute to bundle size bloat, particularly in our Tooltip component. 

Removing them in favor of plain string in-lining results in an overall ~5.1% reduction in gzip size and ~12.1% reduction in minified size in the `radiance-ui` package.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "bundle.js": {
-    "bundled": 838566,
-    "minified": 768893,
-    "gzipped": 90370
+    "bundled": 745130,
+    "minified": 676071,
+    "gzipped": 85753
   },
   "bundle.umd.js": {
-    "bundled": 848390,
-    "minified": 744983,
-    "gzipped": 86334
+    "bundled": 754776,
+    "minified": 652557,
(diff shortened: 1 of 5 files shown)


daigof (approved):
👍 so this only happens when we nest serializations? The size difference is very noticeable and since this is a library it makes more sense but deeply I think plain strings for handling styles is error prone.

michaeljaltamirano:
> 👍 so this only happens when we nest serializations? 

The bloating does, at least. I did not compare what would happen if we converted `styled` components over to `css`, which I think is worth recording. 

> The size difference is very noticeable and since this is a library it makes more sense but deeply I think plain strings for handling styles is error prone.

I don't think the use-case for nested strings is very common, and I don't think the Intellisense benefit outweighs the identifiable performance cost our decreasing `.size-snapshot.json` has been shedding. In a world where most everything is nested in `css` calls anyway, we get the Intellisense benefit for the vast majority of style usage (>95% of css lines?)

PR #667, opened 2020-12-30
[POC] Add Bundlewatch to Build Step

michaeljaltamirano:
### What & Why

I added back the `.size-snapshot.json` file to source control in #352 so we could track size changes and identify unintended regressions and/or bundle bloat. The process of committing the changes was overly manual, though: developers would either have to run `yarn run build` on their own to regenerate the file in their respective feature branches, or they would have to commit it when cutting releases (which includes running `yarn run build`). 

This PR adds [Bundlewatch](https://github.com/bundlewatch/bundlewatch) to our Build step in CI in order to automate it. I've added a secret token to our CI environment variables to report the status as a status check, along with some fairly arbitrary size limits as thresholds for violation (and I did not set it as a `required` step, either, so it is currently non-blocking). 

The status check generates a URL (example: https://ja2r7.app.goo.gl/9QPqJfAZAZSyutAG6) that is clickable under `details`:

Screen Shot 2020-12-30 at 5 37 40 PM

It also removes `.size-snapshot.json` and adds the file back to `.gitignore` since we no longer need to manually update it. It does not remove the `rollup` plugin that generates the `.size-snapshot`, even though we're looking at the build results in Bundlewatch, because it may be useful to look at the logs on CI in the build step if we notice regressions in Bundlewatch.

### Alternatives

I was most familiar with [Bundlesize](https://github.com/siddharthkp/bundlesize), which was the inspiration for Bundlewatch, in part because `emotion` uses `BundleSize`. However, the API is pretty much the same, and there appears to be a [Bundlesize2](https://github.com/siddharthkp/bundlesize2) in development by the same author, so we can swap in the future if it seems appropriate. The main advantage of `Bundlewatch` is the UI linked in the status check.


daigof (approved):
👍 looks good

PR #681, opened 2021-01-08
[Icons] Fix truck-icon-secondary

michaeljaltamirano:
Accidentally copy-pasted `treatment-plan-secondary` in its place.

Before:

Screen Shot 2021-01-08 at 3 29 08 PM

After:

Screen Shot 2021-01-08 at 3 30 32 PM


dbushy727 (approved):
🎉

PR #704, opened 2021-01-21
[Theming] Modify Secondary Theme line-height and font-weight

michaeljaltamirano:
### What & Why

1. Sets `line-height: 1.4` across all applicable `line-height` stylings.
    - Exceptions are of the type like the Dropdown `line-height`, which is set to a very large value to center the text. 
1. Removes `bold` font-weight from secondary header components (display/h1, heading/h2, title/h3).

### Next steps

Release as patch version in consumer apps, but also we're at the point where we should add visual regression testing for our secondary theme, too. So I will open up that PR as well.

### DIFF
--- a/src/shared-components/button/style.ts
+++ b/src/shared-components/button/style.ts
@@ -7,6 +7,7 @@ import { textColorsAssociatedWithColors } from './constants';
 import {
   primaryButtonFontColor,
   primaryButtonBackgroundColor,
+  setThemeLineHeight,
 } from '../../utils/themeStyles';
 
 import { ButtonTypeWithAction } from '.';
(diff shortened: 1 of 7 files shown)


daigof (approved):
Code looks ok, but Im not sure about the Typography line-height changes:
Currently we have:
- display: 1.33
- heading: 1.67
- title: 1.6
- body: 1.75
- caption> 1.71
- label/button: 1.67

Do they really want to fix everything to 1.4 for secondary?

By the way we could just remove `lodash.round` dependency and pass the above values for primary.

michaeljaltamirano:
> Code looks ok, but Im not sure about the Typography line-height changes:
> Currently we have:
> 
> * display: 1.33
> * heading: 1.67
> * title: 1.6
> * body: 1.75
> * caption> 1.71
> * label/button: 1.67
> 
> Do they really want to fix everything to 1.4 for secondary?

Yep. You can look through the Figma file of Ruth UI: everything is 1.4. It was one of Ben's QA requests last cycle, too, that we didn't get to implementing. 

> By the way we could just remove `lodash.round` dependency and pass the above values for primary.

I'm going to keep this for now. It's < 1kb gzipped. If we want to remove it in a future PR (for cooldown) we can do that.

PR #718, opened 2021-01-26
[Components] Add background color to Indicator component

jenniferhm:
This PR updates `<Indicator />` to include a `backgroundColor` prop to allow the component to be reused for purposes outside of messages and notifications. 

Screen Shot 2021-01-26 at 3 47 53 PM

Other: the border-radius has been hardcoded to 50% in order to show the component consistently between the primary and secondary themes.

### DIFF
--- a/src/shared-components/indicator/index.tsx
+++ b/src/shared-components/indicator/index.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { COLORS_PROP_TYPES, ThemeColors } from 'src/constants';
+import { ThemeColors } from 'src/constants';
+import { useTheme } from 'emotion-theming';
 
 import { IndicatorContainer } from './style';
 
@@ -13,13 +14,19 @@ type IndicatorProps = {
  * Indicators are used in navigation to help with wayfinding for messages and notifications.
  * It can also be used for non-navigational purposes for information-intensive pages.
  */
-export const Indicator = ({ text, backgroundColor }: IndicatorProps) => (
-  <IndicatorContainer backgroundColor={backgroundColor}>
-    <div>{text}</div>
-  </IndicatorContainer>
-);
+export const Indicator = ({ text, backgroundColor }: IndicatorProps) => {
+  const theme = useTheme();
+
+  const bgColorWithTheme = backgroundColor || theme.COLORS.error;
+
+  return (
+    <IndicatorContainer backgroundColor={bgColorWithTheme}>
+      <div>{text}</div>
+    </IndicatorContainer>
+  );
+};
 
 Indicator.propTypes = {
   text: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
-  backgroundColor: COLORS_PROP_TYPES,
+  backgroundColor: PropTypes.string,
 };
--- a/src/shared-components/indicator/index.tsx
+++ b/src/shared-components/indicator/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { ThemeColors } from 'src/constants';
+import { COLORS_PROP_TYPES, ThemeColors } from 'src/constants';
 import { useTheme } from 'emotion-theming';
 
 import { IndicatorContainer } from './style';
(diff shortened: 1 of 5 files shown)


src/shared-components/indicator/index.tsx
```diff
@@ -1,21 +1,28 @@
...
 
 type IndicatorProps = {
   text: string | number;
+  backgroundColor?: ThemeColors;
 };
 
 /**
  * Indicators are used in navigation to help with wayfinding for messages and notifications.
+ *
+ * By adding an optional `backgroundColor` prop, the component can be reused for other purposes
```
michaeljaltamirano, line 15: Since this is an open source library we tend to keep comments more generic than this. Perhaps something added to the previous sentence like `It can also be used for non-navigation purposes for information-intensive pages.` Truly spitballing here 😅

jenniferhm: oh right my bad, I forgot, will update it to be more generic! 😅

michaeljaltamirano (approved):
This LGTM. As a courtesy I've approved the visual regression changes on Chromatic and so the build is passing now. 

I also apologize because when you had originally asked the question of how to add styling to the `Indicator`, I was under the impression this would be a user-facing feature, so I hope you do not feel like you went through too many hoops to add this (and hopefully feel more comfortable making changes in here in the future!)

src/shared-components/indicator/style.ts
```diff
@@ -1,10 +1,11 @@
 import styled from '@emotion/styled';
 
 import { style as TYPOGRAPHY_STYLE } from '../typography';
-import { SPACER } from '../../constants';
+import { SPACER, ThemeColors } from '../../constants';
 
-export const IndicatorContainer = styled.div`
-  background-color: ${({ theme }) => theme.COLORS.error};
+export const IndicatorContainer = styled.div<{ backgroundColor?: ThemeColors }>`
```
daigof, line 6: An informal convention we had in Cycle 7 with Michael was to use ES6 default values for optional parameters but  in case where we need to use the theme, we moved the logic to the component instead of style file. COuld you please refactor similar to LoadingSpinner component to keep consinstency please.

daigof (changes requested):

src/shared-components/indicator/style.ts
jenniferhm: Thanks for letting me know! I refactored and moved the theme to the component. See [e24278e](https://github.com/curology/radiance-ui/pull/718/commits/e24278ecb23f6c686131efaef2863d60525771dc)

src/shared-components/indicator/style.ts
```diff
@@ -3,8 +3,8 @@ import styled from '@emotion/styled';
 import { style as TYPOGRAPHY_STYLE } from '../typography';
 import { SPACER } from '../../constants';
 
-export const IndicatorContainer = styled.div`
-  background-color: ${({ theme }) => theme.COLORS.error};
+export const IndicatorContainer = styled.div<{ backgroundColor?: string }>`
```
daigof, line 6: you could remove the `?` optional now since we'll always have a value.

If you want this to looks right in your tests. You could fix the border radius issue now in line 19 below please.
mentioned here https://app.asana.com/0/1156667858696586/1199679191263525

just set to border-radius to 50%

daigof (approved):
Thanks for the changes

src/shared-components/indicator/style.ts
jenniferhm: I can remove `?`.

The border-radius draws from the theme, do you want me to hardcode to 50% in `/indicator/style.ts`?

```
  border-radius: ${({ theme }) => theme.BORDER_RADIUS.medium};
  ```

Or do you want me to update `SECONDARY_BORDER_RADIUS.medium` to be 50%?

jenniferhm:
> We should still enforce that only Theme-defined colors are acceptable props (via prop-types and TypeScript). If the other component uses string, that was an oversight.

Okay, so keep `{ backgroundColor: ThemeColors }` in `style.ts` and `backgroundColor: COLOR_PROP_TYPES` in `Indicator.propTypes`

src/shared-components/indicator/style.ts
michaeljaltamirano: @jenniferhm We should hard-code it for the secondary theme. The secondary theme should not have any `border-radius` but there may be component-level exceptions to this rule: [see Figma link.](https://www.figma.com/file/FjOFrQaYVSdlaMIhpgR192/Ruth-UI?node-id=2315%3A6585&viewport=482%2C328%2C1.202054738998413)

jenniferhm: Okay will do!

PR #720, opened 2021-01-28
Vertical Alignment util function

daigof:
- Create an util to be used here and the consumer repos
- Applied it to Indicator and Chip
- Update Readme publish steps with updated info


yarn.lock
```diff
@@ -3049,14 +3049,6 @@
     "@typescript-eslint/typescript-estree" "4.14.1"
     debug "^4.1.1"
```
daigof, line 3051: I dont why these appeared, I have latest from master, I deleted the `node_modules` folder and hit `yarn`

src/shared-components/chip/style.ts
```diff
@@ -10,6 +11,7 @@ export const ChipText = styled.span`
   font-weight: ${({ theme }) => theme.TYPOGRAPHY.fontWeight.bold};
   position: relative;
   top: 1px;
```
michaeljaltamirano, line 13: Something I would like us to consider is the extent to which our existing primary theme `larssiet ` font is atypical, and that the source of any transforms should only apply to the `primary` theme rather than the secondary theme.

We currently apply `position: relative; top: 1px` to this element because it appears `larssiet` elements are not automatically centered. Below is a video toggling the `font-family` on and off, and you can see that the browser defaults are more accurately centered:

https://user-images.githubusercontent.com/13544620/106490874-e2936e80-647b-11eb-9703-97742ec4bd90.mov

If we remove `position: relative; top: 1px`, the secondary theme version looks perfectly fine! It's because we're accommodating the primary theme that we are adding `applyVerticalAlignmentOffset`. 

So if, for the sake of time, we need to use this `applyVerticalAlignmentOffset` as a v0, I think we can look at this again. However, I don't think this is an approach that we should propagate widely. I did not know that we're accommodating `larssiet` to this extent until now, and I think we should seek to limit its shortcomings to the primary style applications.

daigof: `Chip` maybe a confusing example. Actually this is the only place where we accommodate for larssiet, so the top:1px is the normal code and larssiet is the normal font. The atypical font family here are agency that needs the fix here and in Curology app. We are not doing top:1px anywhere else

michaeljaltamirano: We’re also doing it in Indicator, under the `> div` selector.

daigof: Yeah I meant `indicator` (and chip too). What I remember from these components is that the the top:1px was required because of the `font-size: 12px`.
In the wild all the examples we dont have to apply the top:1px for larssiet, it is perfectly aligned with any technique used So the weird case here is the Agency fonts, I tested via changing to sans-serif and the culprit always is agency font. Actually it varies greatly depends on the fonts installed, I just tested all the other browser fonts. The point here is that we use larssiet without any trick except for these 2 cases. See this magic scenario for example
http://curology.localhost/magic/signup/address?as=patient&scenario=QuizCompleted&patientBrand=1
the numbers in the checkout section.

Now I wonder if the nested div is causing the required top:1px, Ill see if I can get rid of them in Radiance if I have time

michaeljaltamirano: The `Chip` and `Indicator` stylings are not being used because of `font-size`, you can modify the styling in the storybook and play around with the elements to confirm this. These are `font-size: 1.75rem` (up from `0.75rem`) with `font-family` for `primary`, browser default, and then `secondary`:

<div style="display: flex;">
Screen Shot 2021-02-01 at 12 04 13 PM
Screen Shot 2021-02-01 at 12 04 19 PM
Screen Shot 2021-02-01 at 12 04 05 PM
</div>

The bulleted numbers in the checkout section vary by theme because we hard-coded styling that only works for the `primary` `font-family`, which is why it appears the secondary theme is the culprit. The real problem is that hard-coded stylings, in general, will not be compatible. If you remove the `font-family` from the two different themes they are visually identical:

<div style="display: flex;">
Screen Shot 2021-02-01 at 12 13 08 PM
Screen Shot 2021-02-01 at 12 12 54 PM
</div>

In fact, it looks like all that is required to make the `secondary` theme Bullet compatible is to add `display: flex; justify-content: center; align-items: center;` to its container. That's it. _However_, the same does not work for the `primary` theme, because of the aforementioned `font-family` incompatibilities, where we have a hard-coded `line-height` to accommodate its distinctive layout. The previous flex styling works if we also add `position: relative; top: 2px;` to the child (which I wrapped in a span). 

In summary, I don't think applying an additional transform to the secondary theme in `radiance-ui` is the solution. As it concerns `radiance-ui`, I think we should be applying a transform to _only_ the primary theme. And, in userland, I think we should define a helper method there that _also_ targets the primary theme styling. I think this should be the approach because I think it should be understood that are accommodating the `primary` `font-family`. If we end up targeting the `secondary` theme in userland for expediency, I think we should understand that as short-term tech debt we're accruing.

daigof: I cannot focus on this now, all I know is I think I remebered applying flex technique to agency (along with other techniques) and it did not work. Ill revisit this later next week. I dont undestand what you mean by hardcoding setting line-height same as height is a valid vertical centering technique and should accomplish same effect as using flex cetnter

daigof: @michaeljaltamirano You were right, looks like the `line-height` technique is not suppose to work the way we were applying it, it did look good because the primary font is the bugged one. I am all in for refactoring to `display:flex` in the cases. 

Having said that, we still need a common util function to be used in Curology and posibly other repos like Gatsby, so how about leaving this util function here? Ill just have to refactor this to target primary theme instead of secondary and remove the Chip top:1px etc

michaeljaltamirano: I'm comfortable adding a util that targets the primary theme! We just need to also share it with folks with the understanding that it also requires removing and/or updating any hard-coded styling that is working around the primary theme `font-family` peculiarity. So it's more work but I think a step in the right direction.

daigof:
I now repurposed this PR to target the primary theme. In curology for the work (where needed) would be to use the flex technique instead of line-height and apply the fix, I can cover all the known cases so far, is not that bad.

This change should not have any impact in Curology (if we don't touch the code that currently in use)

src/utils/themeStyles/index.ts
```diff
@@ -42,3 +42,6 @@ export const setThemeFontWeight = (theme: ThemeType) =>
   theme.__type === 'secondary'
     ? theme.TYPOGRAPHY.fontWeight.normal
     : theme.TYPOGRAPHY.fontWeight.bold;
+
+export const applyVerticalAlignmentOffset = (theme: ThemeType, offset = '1') =>
```
michaeljaltamirano, line 46: I would consider re-naming this so the intent is more obvious to downstream users who aren't familiar with [all the details of the root cause](https://github.com/curology/radiance-ui/pull/720#discussion_r568047415). 

Perhaps `applyPrimaryThemeVerticalOffset`. I think it's good for the name to refer to the theme explicitly (so more like `setSecondaryHeadingFont` and less like `setButtonStyleFontWeight`). I mentioned `PrimaryTheme` explicitly in it since `setSecondaryHeadingFont` is only ever used within `radiance-ui` so the intent is more obvious, but since this will be used in consumer apps I think the more verbose the better. I removed `Alignment` in case `applyPrimaryThemeVerticalAlignmentOffset` was *too* verbose, though.

michaeljaltamirano (approved):
🙏

src/utils/themeStyles/index.ts
daigof: yep no problem, renamed and publishing

PR #734, opened 2021-02-03
Update theme of background-color on Avatar component to secondaryTint1

malorie16:
This updates the `background-color` of the Avatar component to use `secondaryTint1` rather than `secondary`.


michaeljaltamirano (approved):
LGTM. I left the Design follow-up comments in Figma so this is good to go!

PR #767, opened 2021-02-11
Add ability to change DialogModal background color

joeyquarters:
Exposes a prop, `modalColor`, for the `DialogModal` component. This prop allows us to specify the background color of the modal. Currently, we are limited it to the theme's `background` color.

This change is necessitated by a [new design](https://www.figma.com/file/6QekjeGkdl6PUKNcdQA9mc/CS-and-provider-messages-workload-tweaks?node-id=54%3A0) for the "New Message" prompt in the patient dashboard.

I wasn't able to verify within the app via `yarn link` due to some [puzzling JS errors](https://curology.slack.com/archives/CU02B06GY/p1613010943108300), but Storybook looks 👍 to me.

To review the new stories in Chromatic, looks like I need to log in. Is this an IT helpdesk request? I don't see any Chromatic credentials in 1Password's Engineering vault.

### DIFF
--- a/src/shared-components/dialogModal/index.tsx
+++ b/src/shared-components/dialogModal/index.tsx
@@ -12,18 +12,18 @@ import {
   ModalTitle,
   CrossIconContainer,
 } from './style';
-import { Colors, primaryTheme, ThemeType } from '../../constants';
+import { Colors, primaryTheme, secondaryTheme } from '../../constants';
 
 export interface DialogModalProps {
+  /**
+   * DialogModal background color. Defaults to the current theme's `white` if not specified.
+   */
+  backgroundColor?: Colors['background'];
   /**
    * Dialog Modal content.
    * Must contain at least 1 button and is responsible for closing the modal.
    */
   children: React.ReactNode;
-  /**
-   * DialogModal background color. Defaults to the current theme's `white` if not specified.
-   */
-  modalColor?: Colors['background'];
   /**
    * If provided, DialogModal displays a Close Icon positioned top-right.
    * This function must contain the logic for closing the modal.
@@ -47,14 +47,14 @@ const getDomNode = () =>
  * Dialog Modals should always contain at least 1 button and the logic should close the modal at some point.
  */
 export const DialogModal = ({
+  backgroundColor,
   children,
-  modalColor,
   onCloseIconClick,
   title = '',
   ...rest
 }: DialogModalProps) => {
-  const theme: ThemeType = useTheme();
-  const modalColorWithTheme = modalColor || theme.COLORS.white;
+  const theme = useTheme();
+  const backgroundColorWithTheme = backgroundColor || theme.COLORS.white;
   const [isClosing, setIsClosing] = useState(false);
 
   const domNode = useRef<HTMLElement>(getDomNode());
@@ -96,7 +96,7 @@ export const DialogModal = ({
         <Overlay className={transitionState} {...rest}>
           <FocusScope contain restoreFocus autoFocus>
             <ModalContainer
-              backgroundColor={modalColorWithTheme}
+              backgroundColor={backgroundColorWithTheme}
               className={transitionState}
               onKeyDown={handleKeyDown}
             >
@@ -122,8 +122,11 @@ export const DialogModal = ({
 };
 
 DialogModal.propTypes = {
+  backgroundColor: PropTypes.oneOf([
+    primaryTheme.COLORS.background,
+    secondaryTheme.COLORS.background,
+  ]),
   children: PropTypes.node.isRequired,
-  modalColor: PropTypes.oneOf([primaryTheme.COLORS.background]),
   onCloseIconClick: PropTypes.func,
   title: PropTypes.string,
 };
--- a/src/shared-components/dialogModal/index.tsx
+++ b/src/shared-components/dialogModal/index.tsx
@@ -102,6 +102,7 @@ export const DialogModal = ({
             >
               {onCloseIconClick && (
                 <CrossIconContainer
+                  backgroundColor={backgroundColorWithTheme}
                   onClick={handleCloseIntent}
                   role="button"
                   tabIndex={0}
--- a/src/shared-components/dialogModal/index.tsx
+++ b/src/shared-components/dialogModal/index.tsx
@@ -3,6 +3,7 @@ import React, { useRef, useState, useEffect } from 'react';
 import ReactDOM from 'react-dom';
 import { Transition } from 'react-transition-group';
 import { FocusScope } from '@react-aria/focus';
+import { useTheme } from 'emotion-theming';
 
 import { CrossIcon } from '../../icons';
 import {
@@ -11,6 +12,7 @@ import {
   ModalTitle,
   CrossIconContainer,
 } from './style';
+import { Colors, primaryTheme, ThemeType } from '../../constants';
 
 export interface DialogModalProps {
   /**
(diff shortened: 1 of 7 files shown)


src/shared-components/dialogModal/index.tsx
```diff
@@ -42,10 +48,13 @@ const getDomNode = () =>
  */
 export const DialogModal = ({
   children,
+  modalColor,
   onCloseIconClick,
   title = '',
   ...rest
 }: DialogModalProps) => {
+  const theme: ThemeType = useTheme();
```
michaeljaltamirano, line 56: We've got built-in typing with `useTheme` so we can remove the annotation:

Screen Shot 2021-02-11 at 12 14 51 PM

src/shared-components/dialogModal/index.tsx
```diff
@@ -11,13 +12,18 @@ import {
...
 export interface DialogModalProps {
   /**
    * Dialog Modal content.
    * Must contain at least 1 button and is responsible for closing the modal.
    */
   children: React.ReactNode;
+  /**
+   * DialogModal background color. Defaults to the current theme's `white` if not specified.
+   */
+  modalColor?: Colors['background'];
```
michaeljaltamirano, line 26: Since we're passing the prop to a modal component--and eventually pass it as a prop named `backgroundColor `--should we call this `backgroundColor`? I think that is our convention in most other components, too.

src/shared-components/dialogModal/index.tsx
```diff
@@ -113,6 +123,7 @@ export const DialogModal = ({
 
 DialogModal.propTypes = {
   children: PropTypes.node.isRequired,
+  modalColor: PropTypes.oneOf([primaryTheme.COLORS.background]),
```
michaeljaltamirano, line 126: We'll also want to add the `secondaryTheme` version:

`modalColor: PropTypes.oneOf([primaryTheme.COLORS.background, secondaryTheme.COLORS.background]),`

src/shared-components/dialogModal/style.ts
```diff
@@ -31,7 +31,9 @@ export const Overlay = styled.div`
   }
 `;
 
-export const ModalContainer = styled.div`
+export const ModalContainer = styled.div<{
+  backgroundColor: string;
```
michaeljaltamirano, line 35: [Nit] Since this is always coming from a parent there's no requirement for specificity but I think it makes sense to be consistent and make this `Colors['background']` as well.

stories/dialogModal/index.stories.tsx
```diff
@@ -83,6 +83,82 @@ export const DefaultOpened = () => {
...
+            </Button>
+          </Button.Container>
+        </DialogModal>
+      )}
+    </React.Fragment>
+  );
+};
+
+WithColor.id = `${DIALOG_MODAL_STORY_ID_PREFIX}with-color`;
+WithColor.paramters = {
```
michaeljaltamirano, line 121:
```suggestion
WithColor.parameters = {
```

stories/dialogModal/index.stories.tsx
```diff
@@ -83,6 +83,82 @@ export const DefaultOpened = () => {
...
+  const [openModal, setOpenModal] = useState(false);
+
+  return (
+    <React.Fragment>
+      <Button onClick={() => setOpenModal(true)}>open dialog modal</Button>
+
+      {openModal && (
+        <DialogModal
+          title="Heads up!"
+          modalColor={primaryTheme.COLORS.background}
```
michaeljaltamirano, line 96: Our stories are "theme-aware" as well, so I would pull the `theme` off of a `useTheme` invocation above the return for compatibility with the `secondaryTheme` in Storybook, otherwise it has the `primary` background:

Screen Shot 2021-02-11 at 12 21 54 PM

src/shared-components/dialogModal/test.tsx
```diff
@@ -17,4 +18,15 @@ describe('<DialogModal />', () => {
...
   });
+
+  it('renders dialog modal with custom color', () => {
+    const { container } = render(
+      <DialogModal modalColor={primaryTheme.COLORS.background}>
+        <div>{modalBody}</div>
+      </DialogModal>,
+    );
+
+    // DOM Traversal necessary to find the React portal element
```
michaeljaltamirano, line 29: Good catch, and hopefully you didn't have to struggle too much to get this working. This actually ties into a WIP feature in the `emotion 10 --> 11` upgrade branch. If you want to simplify it in this PR, feel free to steal the code:

```tsx
// src/tests/testingLibraryHelpers.tsx
import React from 'react';
import { ThemeProvider } from 'emotion-theming';
import * as ReactTestingLibrary from '@testing-library/react';

import { primaryTheme, ThemeType } from '../constants';

interface RenderOptions extends ReactTestingLibrary.RenderOptions {
  theme?: ThemeType;
  withPortalContainer?: boolean;
}

/**
 * Many of our pages rely on components that make use of portals.
 *
 * Unit tests for components that do that **do not include such a portal
 * container in their unit test** will have test failures on CI due to
 * serialization order being off.
 */
const usePortalContainer = () => {
  const portalContainer = document.createElement('div');
  portalContainer.setAttribute('id', 'reactPortalSection');
  document.body.appendChild(portalContainer);

  return portalContainer;
};

// We customize @testing-library methods to bake-in theming and keep unit tests DRY.
// We do not use ReactTestingLibrary.render(Component, { wrapper }) option because
// `@testing-library/react` is (somehow) overwriting React with its own API when used.
// This issue is specific to this repo and is not present in other implementations.
// <ThemeProvider> leaves no DOM trace, anyway, so there's no functional difference.
const customRender = (
  Component: React.ReactElement,
  options: RenderOptions = {},
): ReactTestingLibrary.RenderResult => {
  const {
    theme = primaryTheme,
    container,
    withPortalContainer,
    ...rest
  } = options;

  return ReactTestingLibrary.render(
    <ThemeProvider theme={theme}>{Component}</ThemeProvider>,
    {
      container: withPortalContainer ? usePortalContainer() : container,
      ...rest,
    },
  );
};

export * from '@testing-library/react';
export { default as userEvent } from '@testing-library/user-event';

export { customRender as render };
```

After adding then, you can pass a `withPortalContainer` boolean prop, and then rely on `container.firstElementChild` as most of our tests do:

```tsx
// src/shared-components/dialogModal/test.tsx
    const { container } = render(
      <DialogModal modalColor={primaryTheme.COLORS.background}>
        <div>{modalBody}</div>
      </DialogModal>,
      { withPortalContainer: true },
    );

    expect(container.firstElementChild).toMatchSnapshot();

michaeljaltamirano (approved):
The functionality looks good! I'm approving so that my review is not a blocker. 

I "approved" the already-open modal story, and "denied" the one that ran because of the `parameters` typo, so once that commit goes in and removes the test, Chromatic should no longer block the build step. 

The only "blocker" is making sure these changes are compatible with the `secondaryTheme` in Storybook (https://github.com/curology/radiance-ui/pull/767#discussion_r574727589), otherwise we're good to merge this in and release as v16.7.0.

joeyquarters:
Thanks for the great feedback, @michaeljaltamirano. Pushed up changes to address the comments. Also realized I needed to make the background color of the close button utilize the `backgroundColor` prop too, so made that change.

michaeljaltamirano:
> Thanks for the great feedback, @michaeljaltamirano. Pushed up changes to address the comments. Also realized I needed to make the background color of the close button utilize the `backgroundColor` prop too, so made that change.

Great! Feel free to ping me if the instructions on releasing are confusing.

PR #777, opened 2021-02-12
Add glyphs finalized and published in Figma

daigof:
Need these for my project, taken from official Figma files and communicated that they are finished. 

> Except for the Moon Glyph that is use the same Secondary version for both until a curology version is finalized.

EDIT: Primary theme Moon icon was updated to it's true value, both are theme aware now

Primary theme
https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance-UI?node-id=2300%3A18242

Secondary
https://www.figma.com/file/FjOFrQaYVSdlaMIhpgR192/Ruth-UI?node-id=2300%3A18242&viewport=480%2C520%2C0.8739789724349976


michaeljaltamirano:
> Except for the Moon Glyph that is use the same Secondary version for both until a curology version is finalized.

Is that the plan? It looks like it is tagged as a unique secondary theme icon in the Figma file (yellow highlighting). For those we return `null` if they're supposed to be unique, so it might be worth clarifying with the designer if the Figma file is accurate if we want to use it for both.

michaeljaltamirano (approved):
LGTM. One comment about the Figma file being accurate. Looks like you'll just have to approve the Chromatic build and then it's good to go.

daigof:
> Is that the plan? It looks like it is tagged as a unique secondary theme icon in the Figma file (yellow highlighting). For those we return null if they're supposed to be unique, so it might be worth clarifying with the designer if the Figma file is accurate if we want to use it for both.

yes, we need the moon glyph in the Treatment Plan Recommended Routine section 3
https://www.figma.com/file/pCgObXrdupHPWtaUMJL4pV/a-treatment-plan-ew?node-id=2298%3A14679

PR #778, opened 2021-02-12
[Theme] [Components] Add AnchorLinkButton, update TYPOGRAPHY_STYLE.link

michaeljaltamirano:
### CHANGELOG

1. This PR adds a new component, `AnchorLinkButton`, to match the spec of our Figma design for buttons that look like links:

Screen Shot 2021-02-12 at 12 00 46 PM

Here is the Review App story: https://curology-radiance-pr-778.herokuapp.com/?path=/story/components-button-anchorlinkbutton--default

This component does not currently exist in `radiance-ui`. Moreover, attempts to approximate this design using our existing `TYPOGRAPHY_STYLE.link` styling **do not match this spec**. It applies `border-bottom` rather than `text-decoration: underline`, among other things. We've started [surfacing issues around text placement between the two brands](https://github.com/curology/radiance-ui/pull/720#discussion_r568047415), and it seems ideal to rely on CSS to avoid more pixel displacement as we seek out solutions for those discrepancies.

The solution is to add an `AnchorLinkButton` nested under our `Button` components, since it is an underlying `<button>`. This PR also adds TODOs related to design discrepancies that will still exist after making these changes. 

2. This PR also adds `:focus` styling to `TYPOGRAPHY_STYLE.link` so that focus states appear with an on-brand styled outline, rather than the browser/page defaults.

Before:

Screen Shot 2021-02-12 at 12 04 52 PM

Screen Shot 2021-02-12 at 12 05 31 PM

After: 

Screen Shot 2021-02-12 at 12 09 53 PM

Screen Shot 2021-02-12 at 12 09 57 PM

### Release details

This PR requires releasing a new version of radiance due to the `TYPOGRAPHY_STYLE.link` breaking change: we will have to update the API in usage to pass a theme. 

We are also removing the deprecated properties handling in `src/shared-components/typography/index.ts`. When I checked this summer it was not in use (https://github.com/curology/radiance-ui/pull/274#issue-454519414).

### DIFF
--- a/src/shared-components/button/components/anchorLinkButton/style.ts
+++ b/src/shared-components/button/components/anchorLinkButton/style.ts
@@ -1,0 +1,23 @@
+import { css } from '@emotion/core';
+
+import { style as TYPOGRAPHY_STYLE } from '../../../typography';
+import { ThemeType } from '../../../../constants';
+
+/**
+ * TODO: Consolidate, fully, this specific link style with TYPOGRAPHY_STYLE.link,
+ * or otherwise figure out how to maintain meaningful distinction.
+ */
+const anchorLinkButton = (theme: ThemeType) => css`
+  ${TYPOGRAPHY_STYLE.link(theme)};
+  background: transparent;
+  border: none;
+  padding: 0;
+  text-decoration: underline;
+  text-decoration-skip-ink: none;
+  font-size: ${theme.TYPOGRAPHY.fontSize.link};
+  border-radius: ${theme.BORDER_RADIUS.small};
+`;
+
+export default {
+  anchorLinkButton,
+};
--- a/src/shared-components/typography/index.ts
+++ b/src/shared-components/typography/index.ts
@@ -119,7 +119,7 @@ const Caption = styled.p`
 const Display = styled.h1`
   ${({ theme }) => displayStyle(theme)}
 `;
-const ErrorComponent = styled.p`
+const Error = styled.p`
   ${({ theme }) => errorStyle(theme)}
 `;
 const Heading = styled.h2`
@@ -142,16 +142,10 @@ export const Typography = {
   Button,
   Caption,
   Display,
-  Error: ErrorComponent,
+  Error,
   Heading,
   Label,
   Link,
   Success,
   Title,
-
-  // Deprecated legacy names
-  LinkTag: Link,
-  ButtonText: Button,
-  SuccessText: Success,
-  ErrorText: ErrorComponent,
 } as const;
--- a/src/shared-components/typography/index.ts
+++ b/src/shared-components/typography/index.ts
@@ -75,7 +75,10 @@ const buttonStyle = (theme: ThemeType) => `
   text-transform: uppercase;
 `;
 
-const linkStyle = () => `
+/**
+ * TODO-MA: Add theme.TYPOGRAPHY.fontSize.link to guarantee font-size compatibility
+ */
+const linkStyle = (theme: ThemeType) => `
   border-bottom: 1px solid currentColor;
   cursor: pointer;
   text-decoration: none;
@@ -87,6 +90,11 @@ const linkStyle = () => `
   &:hover {
     opacity: 0.6;
     transition: opacity 350ms;
+  }
+
+  &:focus {
+    outline: none;
+    box-shadow: ${theme.BOX_SHADOWS.focus};
   }
 `;
 
@@ -122,7 +130,7 @@ const Label = styled.label`
   ${({ theme }) => labelStyle(theme)}
 `;
 const Link = styled.a`
-  ${linkStyle()}
+  ${({ theme }) => linkStyle(theme)}
 `;
 const Success = styled.p`
   ${({ theme }) => successStyle(theme)}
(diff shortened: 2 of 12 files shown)


src/shared-components/typography/index.ts
```diff
@@ -75,7 +74,10 @@ const buttonStyle = (theme: ThemeType) => `
   text-transform: uppercase;
 `;
 
-const linkStyle = () => `
+/**
+ * TODO-MA: Add theme.TYPOGRAPHY.fontSize.link to guarantee font-size compatibility
```
michaeljaltamirano, line 78: We should enforce the Figma spec of 14px for links. Not include font-size means it'll be dependent on other styles / every dev implementation. We should provide more batteries included functionality.

src/shared-components/button/components/anchorLinkButton/style.ts
```diff
@@ -0,0 +1,23 @@
...
+ * TODO: Consolidate, fully, this specific link style with TYPOGRAPHY_STYLE.link,
+ * or otherwise figure out how to maintain meaningful distinction.
+ */
+const anchorLinkButton = (theme: ThemeType) => css`
+  ${TYPOGRAPHY_STYLE.link(theme)};
+  background: transparent;
+  border: none;
+  padding: 0;
+  text-decoration: underline;
+  text-decoration-skip-ink: none;
```
michaeljaltamirano, line 16: `text-decoration-skip-ink` was added to remove the default behavior of our fonts:

Screen Shot 2021-02-12 at 12 31 06 PM
Screen Shot 2021-02-12 at 12 30 57 PM

In our designs, they do not look like this. I do not know if that is a Figma limitation or something we want to enforce, so `text-decoration-skip-ink` gives it this presentation instead:

Screen Shot 2021-02-12 at 12 30 38 PM
Screen Shot 2021-02-12 at 12 30 43 PM

youngmanalive (approved):
code looks solid! 👌 

my only stupid nit here is the naming. I was sorta confused when i first opened this thinking we were standardizing anchors or something. our `LinkButton` returns an anchor tag. This component is sort of the opposite of that but has `Anchor` _and_ `Link` prefixed on it. for an inline external link, are we still just using a regular `<a>` tag? 

would it make sense to factor this into the `TextButton` component?
```jsx
<TextButton inline />
```

src/shared-components/typography/index.ts
```diff
@@ -148,14 +155,3 @@ export const Typography = {
   SuccessText: Success,
```
daigof, line 155: how about removing these exported members here as well?

daigof (approved):
Code/styles looks good I love the focus styling.

 I understand the rationale because I found the perfect explanation above the component code but I doesnt show in the Story. Could you please add it to the story as well

Same as Nate the `AnchorLinkButton` name threw me off, it sort of like begs for an `href` or some sort of link present if you dont read the usage comments first. I disagree it should be a modified version of TextButton because the intent is different and clearly states it should not be used along other text
https://curology-radiance-pr-778.herokuapp.com/?path=/docs/components-button-textbutton--clickable

Im ok to go with this name since I fail to suggest a better one

daigof:
just a comment fyi u may already know: toggling the disabled attribute on the button does nothing, expectable since it is not a BaseButton child. If this is used to open modals, etc then it is fine that the spec did not mention nor we need to cover this right now

src/shared-components/typography/index.ts
michaeljaltamirano: Missed that, thanks! Fixed in 0e8e2fb8714507944dd6ed026ac8eb69861d2f47

michaeljaltamirano:
> I understand the rationale because I found the perfect explanation above the component code but I doesnt show in the Story. Could you please add it to the story as well

Thanks for catching this: fixed in 3d74a64e38eb32463becdb45c70628b8ec70dea1 (then simplified in b08694dea8ff1a20f4a752bf971e85d9e43414a0). 

> Same as Nate the `AnchorLinkButton` name threw me off, it sort of like begs for an `href` or some sort of link present if you dont read the usage comments first. I disagree it should be a modified version of TextButton because the intent is different and clearly states it should not be used along other text

Yep, I'm not in love with the name but I think having an existing `LinkButton` just makes things trickier for us. We specifically want a button, not an anchor link, but we want it to look like an anchor link. I also think for this purpose it is better to have it as a separate standalone component, since it does not cohere to the same model we use for all the other buttons. (They're our own special Button stylings, not anchor link stylings.) 

The benefit of a standalone name is that we can just replace it later if we come upon a better name. Getting folks familiar with a specific prop on an already existing and large API surface seems less dev friendly.

youngmanalive:
makes sense. agree, not the best idea combining with `TextButton`... was first thing that came to mind since its just clickable text. I'm game for calling this `AnchorButton`, but not gonna split hairs here so I'm up for whatever 🤘.

thanks for adding this, will def come in handy!

ipc103 (approved):
💯  Looks great - this makes things a lot easier. I also tried to think of another name and failed to come up with anything better. Makes me think `AnchorLinkButton` might be a better name than we think 🤣

benkolde (approved):

PR #779, opened 2021-02-12
Change max-height of ModalContainer

malorie16:
The previously hardcoded `max-height` of 700px was causing issues on mobile but changing it to 100% fixes it right up 🎉

## Before

beforeModalContainerHeightBugFix

## After

afterModalContainerHeightBugFix


src/shared-components/dialogModal/style.ts
```diff
@@ -46,7 +46,7 @@ export const ModalContainer = styled.div<{
   background: ${({ backgroundColor }) => backgroundColor};
   padding: ${SPACER.x4large} ${SPACER.large} ${SPACER.xlarge};
   overflow-y: auto;
-  max-height: 700px;
+  max-height: 100%;
```
michaeljaltamirano, line 49: This fixes the `iPhone5` resolution scenario, but then creates a sub-optimal experience on Desktop, taking up the whole viewport, which is not the intent of the `DialogModal`:

Screen Shot 2021-02-12 at 3 51 20 PM

You may be able to wrap the styling in a media query in order to have it only apply to low-resolution devices. However, this seems to me like a violation of the distinction we try and maintain between `DialogModal` and `ImmersiveModal`: "Dialog modals shouldn't contain large content and should not scroll unless screen size dictates it. To display large amounts of content, use Immersive modal instead.". I think this example qualifies as large amounts of content. 

I would try and see if you can wrap `<UrgentTriageModal />` in an `ImmersiveModal` instead of a `DialogModal`. You will likely have to make a few changes if you proceed down this route (e.g. making sure the `Report a Medical Emergency` title persists in the transition as you scroll down). I would consult with stakeholders depending on the route you take--I can follow up later if you'd like more guidance but I have a meeting right now and wanted to get the comment to you 😬

malorie16: I'll take a look at the `ImmersiveModal` and may reach out on Monday with more questions!

daigof (commented):
I think you need 2 things from each type of modal:
- display scrolling content (Immersive)
- Force user to make an action (Dialog)

Immersive Modals always have a close a button and you can close clicking outside of it. So probably that was not intended for your `UrgentTriageModal`

I can think of 2 things to suggest:
- Test with immersiveModal and get an approval to go with it
- Since this is an edge case: test passing custom styles: These modals support the `{...rest}` props if you check the code it is passed to the Overlay component so you could pass a `css` prop with your own styles in your consumer app, you need to get to the ModalContainer and add the mediaquery there (in your consumer app that is)

malorie16:
> I think you need 2 things from each type of modal:
> 
> * display scrolling content (Immersive)
> * Force user to make an action (Dialog)
> 
> Immersive Modals always have a close a button and you can close clicking outside of it. So probably that was not intended for your `UrgentTriageModal`
> 
> I can think of 2 things to suggest:
> 
> * Test with immersiveModal and get an approval to go with it
> * Since this is an edge case: test passing custom styles: These modals support the `{...rest}` props if you check the code it is passed to the Overlay component so you could pass a `css` prop with your own styles in your consumer app, you need to get to the ModalContainer via a css selector and add the mediaquery there (in your consumer app that is)

Thank you for this input Diego! On mobile, the ImmersiveModal takes up the full height of the viewport (-32px) and with some tweaking in the dev tools I wasn't able to get the desired changes needed so that it took up a similar height to that of the DialogModal. This tells me that a more substantial change may be needed to the ImmersiveModal which I don't think is worth the trouble for this bug. What I'm going to try next is writing a media query like Michael mentioned and/or passing custom styles like you mentioned so that I can try and override the `max-height: 700px` on the DialogModal.

This is what I got with the ImmersiveModal out of the box:

immersiveEmergencyModelCurrent

malorie16:
@daigof and @michaeljaltamirano I didn't have any luck accessing the `ModalContainer` that lives within within the `ImmersiveContainer` from the consumer app so I added a media query here. Adding styling with a `css` prop (or even within the styled component) was only applied to the initial "How can we help?" modal but not the "Report an Emergency" modal and I noticed clicking the "Report an Emergency" button didn't trigger a re-render of the component although it's hooked into a store/mobx context (mobx magic? - still learning about this library). The media query here should just work but any feedback on a different approach would be much appreciated!

src/shared-components/dialogModal/style.ts
```diff
@@ -68,6 +68,10 @@ export const ModalContainer = styled.div<{
     }
   }
 
+  @media screen and (max-width: ${BREAKPOINTS.sm}px) {
```
michaeljaltamirano, line 71: You can use `MEDIA_QUERIES` directly as in line 75. We typically develop "mobile first" which really means the base styles we set are for all views, and then we start adding exceptions starting at wider viewport widths:

```diff
- max-height: 700px;
+ max-height: 100%

+ ${MEDIA_QUERIES.smUp} {
+  max-height: 700px;
+ }

**Comment by michaeljaltamirano** [comment]
> @daigof and @michaeljaltamirano I didn't have any luck accessing the `ModalContainer` that lives within within the `ImmersiveContainer` from the consumer app so I added a media query here. Adding styling with a `css` prop (or even within the styled component) was only applied to the initial "How can we help?" modal but not the "Report an Emergency" modal and I noticed clicking the "Report an Emergency" button didn't trigger a re-render of the component although it's hooked into a store/mobx context (mobx magic? - still learning about this library). The media query here should just work but any feedback on a different approach would be much appreciated!

I've opened a PR that (I think) shows how to switch the `UrgentTriageModal` functionality to using an ImmersiveModal. There is indeed some `mobx` magic so even if you decide to go the `radiance-ui`-only route I hope it is instructive 🙂  https://github.com/curology/PocketDerm/pull/12143

**Review by michaeljaltamirano — APPROVED** [review]
"Let's get this merged!"

As noted in my previous comment, I think there is an adequate user-land solution here, but I also think it's eminently reasonable to add this logic so that we do not overload `DialogModal` for lower-end devices! Thank you for your keen eye and penchant for setting dev tools to `iPhone5/SE` 💪

**Review by daigof — CHANGES_REQUESTED** [review]
sorry I haven't have a chance to test this thoroughly, please implement Michael change but please try with with
```
 ${MEDIA_QUERIES.xsUp}
```
Check iphones dimensions reference:
- iPhone 5: 320px
- iPhone X: 375px
- xs breakpoint: 330px

This way you will be targeting/fixing iPhone5 issue without modifying the iPhone X experience which is the device most of our users use (in terms of width dimensions). Test and let me know if that solves your bug Ill test once this is ready. Using `MEDIA_QUERIES.xsUp` is the most harmless approach, I hope it fixes your issue

michaeljaltamirano:
> sorry I haven't have a chance to test this thoroughly, please implement Michael change but please try with with
> 
> ```
>  ${MEDIA_QUERIES.xsUp}
> ```
> 
> Check iphones dimensions reference:
> 
> * iPhone 5: 320px
> * iPhone X: 375px
> * xs breakpoint: 330px
> 
> This way you will be targeting/fixing iPhone5 issue without modifying the iPhone X experience which is the device most of our users use (in terms of width dimensions). Test and let me know if that solves your bug Ill test once this is ready. Using `MEDIA_QUERIES.xsUp` is the most harmless approach, I hope it fixes your issue

I don’t think this is correct. Here is a screenshot from my iPhone 12 Pro:

9683161F-7BFE-468D-A6B1-FACD83368D5A

The real issue is that this content shouldn’t be in a dialog modal, it is too much content. Nevertheless, smUp should be used instead of xsUp. Changing max-height from 700px to 100% will not change the experience for even higher end mobile devices beyond making the close icon actually accessible without needing to scroll it into view.

michaeljaltamirano:
I forgot we can add videos directly now and that I can take a screen recording of my phone 😅

https://user-images.githubusercontent.com/13544620/108371386-509d9c80-71c3-11eb-8e5a-294a16ab3d23.MP4

daigof (approved):
That's fine I thought we were targeting iPhone5 only, I ahvent had the chance to test, but the most restrictive breakpoint that works Im ok with it.

PR #780, opened 2021-02-13
[Callout] Adds new 'success' type

michaeljaltamirano:
### What & Why

Some of our components have a `*type` paradigm that allows us to set specific styling properties, primarily our Buttons (`primary`/`secondary`/`tertiary`/`quaternary`).

This PR adds a `success` type to our Callout component--which pulls in our `success` and `successBackground` colors--to style the Callout. We do not expose styling of the component generally, and rather than overhaul the API to allow any set of colors, this PR prefers the course of adding specific presets for consistency in use. (We can also test here whether any combination is inaccessible, for which the `'success'` preset passes our built-in accessibility add-on).

Design:

Screen Shot 2021-02-12 at 6 13 38 PM

Story:

Screen Shot 2021-02-12 at 6 14 06 PM
Screen Shot 2021-02-12 at 6 14 02 PM

This PR adds a Storybook story for this version in order to test it via Chromatic, rather than add unit tests.

### DIFF
--- a/src/shared-components/callout/index.tsx
+++ b/src/shared-components/callout/index.tsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { useTheme } from 'emotion-theming';
 
 import Style from './style';
-import { COLORS_PROP_TYPES, ThemeColors } from '../../constants';
+import { COLORS_PROP_TYPES, ThemeColors, ThemeType } from '../../constants';
 
 export interface CalloutProps {
   /**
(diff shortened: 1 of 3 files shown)


jashieh (approved):

youngmanalive (approved):
🔥 lgtm!

ipc103 (approved):
💯  Great idea. You mentioned this at our D&D meeting last week, but I like this approach of extending component functionality in Radiance. Personally, I've done some UI development in Pocketderm that felt okay at the time, but now seems hacky and difficult to maintain (especially now that we're supporting two themes!) 

Stuff like this seems like a big step in the right direction - nice work!

PR #797, opened 2021-02-15
[Chromatic] Add Visual Regression Testing to Secondary Theme

michaeljaltamirano:
### What & Why

This PR adds Chromatic support for our Secondary Theme.

In Storybook we toggle between themes using globals ([docs](https://storybook.js.org/docs/react/essentials/toolbars-and-globals)), of which Chromatic is not aware. 

Screen Shot 2021-02-16 at 10 22 10 PM

I reached out to Chromatic they recommended, as a workaround, to render both versions together using [decorators](https://storybook.js.org/docs/react/writing-stories/decorators). That could work, but it seems sub-optimal: we should try to preserve a single, consistent theme across storybooks (with toggling), as well as the single component focus of the stories. (A unintended side-effect of adding multiple stories is more noise in the code/source examples in the `Docs` page.) This will prevent the UI from becoming overwhelming for non-technical stakeholders. 

After experimenting with various API (decorators, helper methods to wrap existing stories, etc.), I eventually landed on the simple solution of separate file definition. This also allowed us to also write tests to ensure that when we write new stories (for new components) the suite will catch if we neglect to add a corresponding `secondary.stories.tsx` file. This setup should make it easier to expand our theming options (like light/dark mode) if we choose to. 

I've added [new instructions in Notion](https://www.notion.so/curology/Radiance-UI-3e2736219ba14a91b14075a3d9cd922e#cdcb90d9ac6d4863a4239c0930474b76) to reflect these changes.

### Other API Considered

You can see in 7abcb5fb2a36c61ac76c60efab694088b582f3ce an attempt using (basically) higher order components and the `excludeStories` docs option to wrap existing stories to produce `*Secondary` counterparts. While it was very strongly typed, there was no way to tell if all stories were correctly duplicated, and it was much more work than just re-exporting all of the exports in `index.stories.tsx` and changing the `theme` parameter. 

### Build note

I'll review and approve the new changes tomorrow AM.


.storybook/theme.ts
```diff
@@ -24,15 +28,17 @@ const Theme = create({
   barSelectedColor: '#1EA7FD',
 
   // Typography
-  fontBase: `"Nunito Sans", -apple-system, ".SFNSText-Regular", "San Francisco", BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Helvetica, Arial, sans-serif`,
+  fontBase:
+    primaryTheme.FONTS.baseFont +
```
michaeljaltamirano, line 32: I added our fonts here so that the `Docs` page uses the respective theme font when toggling instead of the Storybook default. I explored making other changes but I couldn't really get most to show up, I think our `<Global />` resets override most of these theme configuration options.

stories/utils.tsx
```diff
@@ -10,3 +11,10 @@ import { StoryFnReactReturnType } from '@storybook/react/dist/client/preview/typ
 export const modalStoryDecoratorForChromatic = (
   storyFn: () => StoryFnReactReturnType,
 ) => <div style={{ width: '1200px', height: '1000px' }}>{storyFn()}</div>;
+
+/**
+ * We exclude stories that exists only for Chromatic visual testing from non-Chromatic environments
+ */
+export const excludeStoriesForNonChromaticEnvironments = (
+  stories: string[] = [],
+) => (isChromatic() ? stories : /.*/);
```
michaeljaltamirano, line 20: We do this to `prevent the UI from becoming overwhelming for non-technical stakeholders` as mentioned in the PR. It seems simplest to rely on the toggle for folks to switch between the different styles rather than duplicate the stories in the sidebar, which is what happens if you remove the regex:

Screen Shot 2021-02-16 at 10 47 21 PM

src/shared-components/alert/index.tsx
```diff
@@ -87,19 +87,41 @@ export const Alert = (alertProps: AlertProps) => {
   };
 
   useEffect(() => {
-    // Truncate text logic
```
michaeljaltamirano, line 90: I moved the two concerns (truncating text and alert duration) into separate effects for clarity. 

You can reproduce an issue where we get permanently stuck in the `while` loop at https://radiance-ui.curology.com/?path=/story/components-alert--with-image if you toggle between the themes once or twice. You'll notice the text is not truncated. That's because `contentElement.scrollHeight > contentElement.offsetHeight` is always true. This fixes the issue, which Chromatic actually caught in [Build 660](https://www.chromatic.com/build?appId=5f6e46f1e7081000221523cf&number=660).

daigof (approved):
cool 👍

PR #800, opened 2021-02-17
Add secondary theme icons

daigof:
New icons added for secondary theme only. primary returns null for now


src/svgs/glyphs/resveratrol-glyph-secondary.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="48" height="48" viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <title>Resveratrol Tea Glyph</title>
```
michaeljaltamirano, line 2: Is `Tea` a copy-paste carry-over from another SVG? It's not in the title.

michaeljaltamirano (approved):

jashieh (approved):

src/svgs/glyphs/resveratrol-glyph-secondary.svg
daigof: fixed, was a typo. Now I wonder if you check the generated file it wont add the title tag (no SEO benefit or anything) I wonder if we should keep adding this title tag manually? what is its worth?

michaeljaltamirano: I think updating our [SVGR settings](https://react-svgr.com/docs/options/#title) to include more information will help with our work to improve accessibility. It's something I noted when we were working on all these last cycle but did not get to.

PR #815, opened 2021-02-23
More ingredients icons

daigof:
- Added all primary: Desonide, Metronidazole, TranexamicAcid
- We have ingredients icons not present in Secondary theme so I added the option to pass null to secondary icon. Thus removing Clidamicyn and Zinc secondaries since they were put there just to comply with TS (they are not used in Secondary consumer app right now (not prescribable))

Screen Shot 2021-02-23 at 14 55 03
Screen Shot 2021-02-23 at 14 54 53


michaeljaltamirano (approved):
🚀

PR #819, opened 2021-02-24
[Linting] Expand @typescript-eslint/eslint-plugin configuration

michaeljaltamirano:
### What & Why

This PR goes through *all* of the `@typescript-eslint/eslint-plugin`-specific rules, turning most of those not included in any of the `recommended` presets on (and other less helpful ones off). I will add the rationale for all the choices individually. [You can find the list of the rules here.](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules)

I'd like for us to take advantage of `radiance-ui` as a kind of `@next` playground for tooling. The library is a good environment to try out stricter settings/standards and seeing how they play out before raising them as potential team-level conventions and rules. This would also mean that contributing to this library could improve code habits/preferences. 

The functional changes are minor, so there is a change in size reported in `bundlewatch`. I expect that is just a transpilation concern, and can do a pre-release version of these changes when merged in to double-check in PocketDerm that cypress/etc. doesn't break. Nothing was reported in Chromatic, though, and playing around in Storybook showed no regressions.

Note: This PR does not go through the [Extended Rules](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#extension-rules). These require explicitly turning off the base `eslint` rule before turning on the `@typescript-eslint`-compatible version. We already do this with many (e.g. `no-shadow`) but they're less of a priority, especially since `prettier` covers most of the purely stylistic rules. 

### Rule Changes (No code changes, rules turned "off")

1. `"@typescript-eslint/non-nullable-type-assertion-style"`: prefers `const val = maybe!` instead of `const val = maybe as string`, where `maybe: string | undefined`. We have other stricter rules that supersede this.
1. `"@typescript-eslint/no-unnecessary-qualifier"`: We don't use namespaces or enums in this repo. 
1. `"@typescript-eslint/no-unnecessary-type-arguments"`: We don't really use type arguments to generics, and in any case I think the verbosity of explicitly providing a default isn't a bad thing. 
1. `"@typescript-eslint/prefer-readonly"`: We don't really have any class models with properties, so we turn this off. 
1. `"@typescript-eslint/prefer-readonly-parameter-types"`: There's also already an existing `eslint` rule not to mutate parameters, which seems the correct concern.
1. `"@typescript-eslint/prefer-reduce-type-parameter"`: We rely on `as` casts for some more complex types in reduce, and I could not manage to get the generic type parameter working correctly. 
1. `"@typescript-eslint/sort-type-union-intersection-members"`: This was not as helpful as sorting type parameters alphabetically, e.g. `avatarSize: 'small' | 'medium' | 'large';` was alphabetized to `avatarSize: 'large' | 'medium' | 'small'`, and `0.25rem` was placed after `0.5rem`, too. There could be kind of hard-coded "sequences" that are already "sorted", just not alphabetically.
    - If reviewers don't think that's that big of a deal, we can turn it on, but I think _sometimes_ we try and include intent in union types by putting more common/prominent cases first. I'm open to that potentially being something we should automate away!

### Rule Changes (Less significant, rules set to "error")

1. `"@typescript-eslint/consistent-type-assertions"`: prefers `let x = "hello" as SomeType` over `let x = <SomeType>"hello"`, aka `as` instead of angle-brackets. 
    - (There is stricter enforcement option to reduce `as` casts but it is overly restrictive due to our reliance on `Object` properties methods that necessarily cast string literal keys to just `string`.)
1. `"@typescript-eslint/no-base-to-string"`: makes sure if we call `toString` it is meaningful and won't result in `'[object Object]'`. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/no-confusing-non-null-assertion"`: discourages using the non-null assertion (e.g. `ref.current!`), and we should not use that in this codebase, anyway. 
1. `"@typescript-eslint/no-invalid-void-type"`: Similar to the other void rule in the next **More Significant** section, I think we've been treating default `undefined` returns as "void" when they're not exactly the same thing. Apparently, void should be stricter, e.g. a function should really not be returning `void | boolean`: it should be boolean all the way through or void all the way through. We had one type definition error on this, which in retrospect looks like a misreading of some `onClick` functionality. This rule helped call that out, so I think it's good to keep. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/no-parameter-properties"`: Similar to turning off `prefer-readonly-parameter-types"`, we do not really have this convention. I also think this is confusing, and encountered in it our main repo earlier this week:
```tsx
class Foo {
  constructor(readonly name: string) {}
}
```
Transpiles to:
```tsx
class Foo {
    constructor(name) {
        this.name = name;
    }
}
```
6. `"@typescript-eslint/no-unnecessary-boolean-literal-compare"`: Because we have stricter rules on boolean expressions (addressed in the next section) this is a minor rule that just checks `if (condition)` and `if (!condition)` instead of checking `if (condition === true)` and `if (condition === false)`. **I'm not wedded to this one.** We can turn it off if we like the verbosity: I actually prefer the `=== true`-type comparisons, but I think most folks prefer the concise version. 
1. `"@typescript-eslint/no-unnecessary-condition"`: Cleans up "defensive" kind of checks when the types indicate we do not need to add them: `if (!!condition)` where `condition: boolean`. Also cleans up `document?.querySelector(...)` where `document` is always defined. We can safely enable this because of the overall type-strictness these new rules, as a whole, enforce.
1. `"@typescript-eslint/no-unnecessary-type-constraint"`: A minor rule, and the more useful flipside of `@typescript-eslint/no-unnecessary-type-arguments`, if someone defines a generic like `<T extends unknown>`, which is the default value, the linter shows "Constraining the generic type `T` to `unknown` does nothing and is unnecessary." This is a rule that will prevent examples of this behavior proliferating--potentially giving some the impression they're unnecessary--while teaching folks if they add it of their own accord, to boot. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-includes"`: `Array.includes(val)` is preferable to `Array.indexOf(val) !== -1`. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-string-starts-ends-with"`: Similar to `"@typescript-eslint/prefer-includes"`, these ES2015 methods are more readable than their `indexOf`/etc. equivalents. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/promise-function-async"`: requires any function that returns a promise be designated as async, which would require us to more thoroughly handle cases after designating it as such.
`"@typescript-eslint/switch-exhaustiveness-check"`: makes sure we handle all cases in conditionals that switch on union types. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/unified-signatures"`: ensures that if we use function overloading we can DRY up our types specified if it catches redundancies. **I think this is a good candidate for team-wide config**.

Rules that do pretty much what they say, with no code changes: `"@typescript-eslint/no-dynamic-delete"`, `"@typescript-eslint/no-extraneous-class"`, `"@typescript-eslint/no-implicit-any-catch": "error"`, `"@typescript-eslint/no-require-imports"`.

### Rule Changes (More significant, rules set to "error")

The most significant rule change is `"@typescript-eslint/strict-boolean-expressions"`. Other changes made downstream of it are in effect minor because many are only flagged when we do not follow this rule, so I am discussing this rule first. 

We rely on a lot of coercion in our casting, for truthiness, for falsiness, against basically every type (string, number, null, undefined, even objects/arrays). These are the errors that display when we violate this rule:

1. `Unexpected nullable string value in conditional. Please handle the nullish/empty cases explicitly`: Given the statement `const val = someNullishString || someFallbackValue`, `val` will be `someFallbackValue` if `someNullishString` is an empty string. Changing the `||` to `??` resolves this error. It changes the behavior to be type-consistent: if string, then truthy case; else falsy case. This same logic applies to other similarly coerced values (e.g. `number | null` but `number` is 0). 
2. `Unexpected value in conditional. A boolean expression is required.`: The following logic `const id = optionValue ? `${optionValue}` : `undefined-${index}`;`, where `optionValue: string | number | undefined` will return the falsy case if it is an empty string or 0. The following is the "fixed" version: `const id = isDefined(optionValue) ? `${optionValue}` : `undefined-${index}`;`, which has no ambiguity in outcome based on type. 
    - Another example can be found ([here](https://github.com/curology/radiance-ui/pull/819/files#diff-ad105899a99dc8bbd10e319363c8269050118df1af6637cac5ee2c511cb5d86cR60)), which removes an empty string default arg (made to be type consistent) in favor of boolean consistency without type ambiguity. 
3. It can also call out potential false positives if we are casting with `Unexpected object value in conditional. The condition is always true`, e.g. `const someVal = someNullishVal as SomeNonNullishType || someFallback`. Removing that then bounces back to rule 1, where we can write more type-consistent code with `const someVal = someNullishVal ?? someFallback`. 
    - This rule caught that we've had an existing type error in our `textColorsAssociatedWithColors` helper method. I've converted it and its usage to expect a Map instead of a POJO for type accuracy.
    
Below are the other significant rule changes: 

1. `"@typescript-eslint/member-ordering` requires that our properties in type definitions be alphabetized, with index signatures (e.g. `[key: string]: unknown`) coming last, which is our convention. Interestingly, the default appears to be to set them first, but I prefer our style. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-nullish-coalescing"`: Most instances of these are pretty much wholly intertwined with `@typescript-eslint/strict-boolean-expressions`.
1. `"@typescript-eslint/prefer-optional-chain"`: Prefers `someObj?.someNullishProperty?.deeplyNestedNullishProperty` to `someObj && someObj.someNullishProperty && someObj.someNullishProperty.deeplyNestedNullishProperty`. I think we've started adopting this in our main codebase already, and so **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-ts-expect-error"`:` Quite helpful! And also updates in `--fix` so not much work is required. The main difference with `ts-ignore` is that `ts-expect-error` will error if there is not a type error! This will be useful for us cleaning up old `ts-ignores`. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/no-confusing-void-expression"`:` I really like this rule! It is a `--fix` rule, and wraps curly braces around functions that should return `void` to make clear there is not a return value. Personally I have never done this, but updating the `void` rules here actually made me realize that our `onClick` typing for the Button component was incorrect, so anything pushing us toward type-correctness seems a good addition. 
```

### CODE REVIEW

File: src/shared-components/button/components/linkButton/index.tsx [hunk 1]
```diff
@@ -66,8 +66,10 @@ export const LinkButton = ({
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...rest}
     >
-      <ButtonContents hasIcon={false}>
-        <ButtonText>{children}</ButtonText>
+      <ButtonContents isLoading={false} isFullWidth={false} hasIcon={false}>
```
michaeljaltamirano, line 69: Rather than rely on casting undefined props in `style.ts`, we explicitly pass in what is the default behavior we've expressed. This makes it clearer from the markup outside of `style.ts` what kinds of props the component expresses.

src/shared-components/button/shared-components/container/index.ts
```diff
@@ -35,7 +35,7 @@ const ButtonContainer = styled.div<{ reverse?: boolean; flex?: boolean }>`
     margin-top: 0;
   }
 
-  ${({ reverse }) => reverse && reverseStyles};
+  ${({ reverse = false }) => reverse && reverseStyles};
```
michaeljaltamirano, line 38: Here, for instance, I used ES6 defaults because the Container is a more general component and it seemed appropriate to keep that concern in here instead of components that use it as a subcomponent.

michaeljaltamirano:
Closing in favor of #821 because we skip Chromatic builds when the feature branch is prefaced with `eslint/` 🙃

PR #821, opened 2021-02-25
[Linting] Expand @typescript-eslint/eslint-plugin configuration

michaeljaltamirano:
### What & Why

This PR goes through *all* of the `@typescript-eslint/eslint-plugin`-specific rules, turning most of those not included in any of the `recommended` presets on (and other less helpful ones off). I will add the rationale for all the choices individually. [You can find the list of the rules here.](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#supported-rules)

I'd like for us to take advantage of `radiance-ui` as a kind of `@next` playground for tooling. The library is a good environment to try out stricter settings/standards and seeing how they play out before raising them as potential team-level conventions and rules. This would also mean that contributing to this library could improve code habits/preferences. 

The functional changes are minor, so there is a change in size reported in `bundlewatch`. I expect that is just a transpilation concern, and can do a pre-release version of these changes when merged in to double-check in PocketDerm that cypress/etc. doesn't break. Nothing was reported in Chromatic, though, and playing around in Storybook showed no regressions.

Note: This PR does not go through the [Extended Rules](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin#extension-rules). These require explicitly turning off the base `eslint` rule before turning on the `@typescript-eslint`-compatible version. We already do this with many (e.g. `no-shadow`) but they're less of a priority, especially since `prettier` covers most of the purely stylistic rules. 

Also note: The failed `Build` step is because we ran out of Open Source snapshots yesterday and so the most recent changes weren't verified. They're just icon changes: https://www.chromatic.com/test?appId=5f6e46f1e7081000221523cf&id=603820542bdd1500238a6bae

### Rule Changes (No code changes, rules turned "off")

1. `"@typescript-eslint/non-nullable-type-assertion-style"`: prefers `const val = maybe!` instead of `const val = maybe as string`, where `maybe: string | undefined`. We have other stricter rules that supersede this.
1. `"@typescript-eslint/no-unnecessary-qualifier"`: We don't use namespaces or enums in this repo. 
1. `"@typescript-eslint/no-unnecessary-type-arguments"`: We don't really use type arguments to generics, and in any case I think the verbosity of explicitly providing a default isn't a bad thing. 
1. `"@typescript-eslint/prefer-readonly"`: We don't really have any class models with properties, so we turn this off. 
1. `"@typescript-eslint/prefer-readonly-parameter-types"`: There's also already an existing `eslint` rule not to mutate parameters, which seems the correct concern.
1. `"@typescript-eslint/prefer-reduce-type-parameter"`: We rely on `as` casts for some more complex types in reduce, and I could not manage to get the generic type parameter working correctly. 
1. `"@typescript-eslint/sort-type-union-intersection-members"`: This was not as helpful as sorting type parameters alphabetically, e.g. `avatarSize: 'small' | 'medium' | 'large';` was alphabetized to `avatarSize: 'large' | 'medium' | 'small'`, and `0.25rem` was placed after `0.5rem`, too. There could be kind of hard-coded "sequences" that are already "sorted", just not alphabetically.
    - If reviewers don't think that's that big of a deal, we can turn it on, but I think _sometimes_ we try and include intent in union types by putting more common/prominent cases first. I'm open to that potentially being something we should automate away!

### Rule Changes (Less significant, rules set to "error")

1. `"@typescript-eslint/consistent-type-assertions"`: prefers `let x = "hello" as SomeType` over `let x = <SomeType>"hello"`, aka `as` instead of angle-brackets. 
    - (There is stricter enforcement option to reduce `as` casts but it is overly restrictive due to our reliance on `Object` properties methods that necessarily cast string literal keys to just `string`.)
1. `"@typescript-eslint/no-base-to-string"`: makes sure if we call `toString` it is meaningful and won't result in `'[object Object]'`. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/no-confusing-non-null-assertion"`: discourages using the non-null assertion (e.g. `ref.current!`), and we should not use that in this codebase, anyway. 
1. `"@typescript-eslint/no-invalid-void-type"`: Similar to the other void rule in the next **More Significant** section, I think we've been treating default `undefined` returns as "void" when they're not exactly the same thing. Apparently, void should be stricter, e.g. a function should really not be returning `void | boolean`: it should be boolean all the way through or void all the way through. We had one type definition error on this, which in retrospect looks like a misreading of some `onClick` functionality. This rule helped call that out, so I think it's good to keep. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/no-parameter-properties"`: Similar to turning off `prefer-readonly-parameter-types"`, we do not really have this convention, [nor is it something that TC39 would be adding to JavaScript anytime soon](https://es.discourse.group/t/class-property-parameters/543):
```tsx
class Foo {
  constructor(readonly name: string) {}
}
```
Transpiles to:
```tsx
class Foo {
    constructor(name) {
        this.name = name;
    }
}
```
6. `"@typescript-eslint/no-unnecessary-boolean-literal-compare"`: Because we have stricter rules on boolean expressions (addressed in the next section) this is a minor rule that just checks `if (condition)` and `if (!condition)` instead of checking `if (condition === true)` and `if (condition === false)`. **I'm not wedded to this one.** We can turn it off if we like the verbosity: I actually prefer the `=== true`-type comparisons, but I think most folks prefer the concise version. 
1. `"@typescript-eslint/no-unnecessary-condition"`: Cleans up "defensive" kind of checks when the types indicate we do not need to add them: `if (!!condition)` where `condition: boolean`. Also cleans up `document?.querySelector(...)` where `document` is always defined. We can safely enable this because of the overall type-strictness these new rules, as a whole, enforce.
1. `"@typescript-eslint/no-unnecessary-type-constraint"`: A minor rule, and the more useful flipside of `@typescript-eslint/no-unnecessary-type-arguments`, if someone defines a generic like `<T extends unknown>`, which is the default value, the linter shows "Constraining the generic type `T` to `unknown` does nothing and is unnecessary." This is a rule that will prevent examples of this behavior proliferating--potentially giving some the impression they're unnecessary--while teaching folks if they add it of their own accord, to boot. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-includes"`: `Array.includes(val)` is preferable to `Array.indexOf(val) !== -1`. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-string-starts-ends-with"`: Similar to `"@typescript-eslint/prefer-includes"`, these ES2015 methods are more readable than their `indexOf`/etc. equivalents. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/promise-function-async"`: requires any function that returns a promise be designated as async, which would require us to more thoroughly handle cases after designating it as such.
`"@typescript-eslint/switch-exhaustiveness-check"`: makes sure we handle all cases in conditionals that switch on union types. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/unified-signatures"`: ensures that if we use function overloading we can DRY up our types specified if it catches redundancies. **I think this is a good candidate for team-wide config**.

Rules that do pretty much what they say, with no code changes: `"@typescript-eslint/no-dynamic-delete"`, `"@typescript-eslint/no-extraneous-class"`, `"@typescript-eslint/no-implicit-any-catch": "error"`, `"@typescript-eslint/no-require-imports"`.

### Rule Changes (More significant, rules set to "error")

The most significant rule change is `"@typescript-eslint/strict-boolean-expressions"`. Other changes made downstream of it are in effect minor because many are only flagged when we do not follow this rule, so I am discussing this rule first. 

We rely on a lot of coercion in our casting, for truthiness, for falsiness, against basically every type (string, number, null, undefined, even objects/arrays). These are the errors that display when we violate this rule:

1. `Unexpected nullable string value in conditional. Please handle the nullish/empty cases explicitly`: Given the statement `const val = someNullishString || someFallbackValue`, `val` will be `someFallbackValue` if `someNullishString` is an empty string. Changing the `||` to `??` resolves this error. It changes the behavior to be type-consistent: if string, then truthy case; else falsy case. This same logic applies to other similarly coerced values (e.g. `number | null` but `number` is 0). 
2. `Unexpected value in conditional. A boolean expression is required.`: The following logic `const id = optionValue ? `${optionValue}` : `undefined-${index}`;`, where `optionValue: string | number | undefined` will return the falsy case if it is an empty string or 0. The following is the "fixed" version: `const id = isDefined(optionValue) ? `${optionValue}` : `undefined-${index}`;`, which has no ambiguity in outcome based on type. 
    - Another example can be found ([here](https://github.com/curology/radiance-ui/pull/819/files#diff-ad105899a99dc8bbd10e319363c8269050118df1af6637cac5ee2c511cb5d86cR60)), which removes an empty string default arg (made to be type consistent) in favor of boolean consistency without type ambiguity. 
3. It can also call out potential false positives if we are casting with `Unexpected object value in conditional. The condition is always true`, e.g. `const someVal = someNullishVal as SomeNonNullishType || someFallback`. Removing that then bounces back to rule 1, where we can write more type-consistent code with `const someVal = someNullishVal ?? someFallback`. 
    - This rule caught that we've had an existing type error in our `textColorsAssociatedWithColors` helper method. I've converted it and its usage to expect a Map instead of a POJO for type accuracy.
    
Below are the other significant rule changes: 

1. `"@typescript-eslint/member-ordering` requires that our properties in type definitions be alphabetized, with index signatures (e.g. `[key: string]: unknown`) coming last, which is our convention. Interestingly, the default appears to be to set them first, but I prefer our style. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-nullish-coalescing"`: Most instances of these are pretty much wholly intertwined with `@typescript-eslint/strict-boolean-expressions`.
1. `"@typescript-eslint/prefer-optional-chain"`: Prefers `someObj?.someNullishProperty?.deeplyNestedNullishProperty` to `someObj && someObj.someNullishProperty && someObj.someNullishProperty.deeplyNestedNullishProperty`. I think we've started adopting this in our main codebase already, and so **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/prefer-ts-expect-error"`:` Quite helpful! And also updates in `--fix` so not much work is required. The main difference with `ts-ignore` is that `ts-expect-error` will error if there is not a type error! This will be useful for us cleaning up old `ts-ignores`. **I think this is a good candidate for team-wide config**.
1. `"@typescript-eslint/no-confusing-void-expression"`:` I really like this rule! It is a `--fix` rule, and wraps curly braces around functions that should return `void` to make clear there is not a return value. Personally I have never done this, but updating the `void` rules here actually made me realize that our `onClick` typing for the Button component was incorrect, so anything pushing us toward type-correctness seems a good addition.


src/shared-components/button/components/linkButton/index.tsx
```diff
@@ -66,8 +66,10 @@ export const LinkButton = ({
       // eslint-disable-next-line react/jsx-props-no-spreading
       {...rest}
     >
-      <ButtonContents hasIcon={false}>
-        <ButtonText>{children}</ButtonText>
+      <ButtonContents isLoading={false} isFullWidth={false} hasIcon={false}>
```
michaeljaltamirano, line 69: Rather than rely on casting undefined props in `style.ts`, we explicitly pass in what is currently the default behavior for each component. This makes it clearer from the markup outside of `style.ts` what kinds of props the component expresses. 

We could also wrap `style.ts` props in `isDefined`, or use ES6 defaults in that file. I originally started doing that and didn't think it was very elegant, instead preferring a very explicit API. **It does not change the API for the consumer**, which was more important than breaking the internal API. But I'm open to feedback here.

src/shared-components/button/shared-components/container/index.ts
```diff
@@ -35,7 +35,7 @@ const ButtonContainer = styled.div<{ reverse?: boolean; flex?: boolean }>`
     margin-top: 0;
   }
 
-  ${({ reverse }) => reverse && reverseStyles};
+  ${({ reverse = false }) => reverse && reverseStyles};
```
michaeljaltamirano, line 38: Here, for instance, I used ES6 defaults because the Container is a more general component and it seemed appropriate to keep that concern in here instead of components that use it as a subcomponent.

.eslintrc
```diff
@@ -39,13 +39,51 @@
...
+    "@typescript-eslint/no-implicit-any-catch": "error",
+    "@typescript-eslint/no-invalid-void-type": "error",
+    "@typescript-eslint/non-nullable-type-assertion-style": "off",
+    "@typescript-eslint/no-parameter-properties": "error",
+    "@typescript-eslint/no-require-imports": "error",
     "@typescript-eslint/no-shadow": "error",
+    "@typescript-eslint/no-unnecessary-boolean-literal-compare": "error",
+    "@typescript-eslint/no-unnecessary-condition": "error",
+    "@typescript-eslint/no-unnecessary-qualifier": "off",
+    "@typescript-eslint/no-unnecessary-type-arguments": "off",
```
timmartin19, line 71: It might be worth not allowing generic type default arguments. I find them to have very confusing behavior.  Mainly in that they don't infer if you only specify one.  So specifying one effectively specifies them all anyways

.eslintrc
```diff
@@ -39,13 +39,51 @@
...
+    "@typescript-eslint/no-unnecessary-type-constraint": "error",
     "@typescript-eslint/no-use-before-define": "error",
+    "@typescript-eslint/prefer-includes": "error",
+    "@typescript-eslint/prefer-nullish-coalescing": "error",
+    "@typescript-eslint/prefer-optional-chain": "error",
+    "@typescript-eslint/prefer-readonly": "off",
```
timmartin19, line 77: I honestly think we should keep this rule.  I personally really like the access modifiers and we definitely criminally underuse `readonly`

.eslintrc
```diff
@@ -39,13 +39,51 @@
...
+    "@typescript-eslint/prefer-readonly-parameter-types": "off",
```
timmartin19, line 78: I think it would be nice to pull this up from the eslint rule into the type space.  Especially because the eslint rule doesn't handle some edge cases (reassiging the parameter, array manipulation, etc).  Plus it's technically stricter/more accurate typing.

.eslintrc
```diff
@@ -39,13 +39,51 @@
...
+    "@typescript-eslint/prefer-reduce-type-parameter": "off",
+    "@typescript-eslint/prefer-string-starts-ends-with": "error",
+    "@typescript-eslint/prefer-ts-expect-error": "error",
+    "@typescript-eslint/promise-function-async": "error",
```
timmartin19, line 82: I personally love this rule and think we should introduce it team wide.

.eslintrc
```diff
@@ -39,13 +39,51 @@
     ],
     "@typescript-eslint/camelcase": "off",
     "@typescript-eslint/consistent-indexed-object-style": "error",
+    "@typescript-eslint/consistent-type-assertions": "error",
     "@typescript-eslint/consistent-type-definitions": ["error", "interface"],
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/member-ordering": [
```
timmartin19, line 46: I use this in my personal project (and import ordering).  95% of the time, it does what I want.  5% of the time it doesn't.  Personally, I'm cool with this.  That being said, if you don't include the autofix, it's incredibly annoying...

timmartin19 (approved):
Overall, I'm a big fan! I left a few comments on individual rules (I was also kind of responding to your comments in the PR body so think of them as continuing that thread).  I also like your idea of using `radiance` as a proving ground.

That being said, when we do go to introduce these to the PocketDerm repo (which I hope is soon), I'd recommend introducing a few at a time to avoid overwhelming everybody.

.eslintrc
michaeljaltamirano: I don't think I saw a rule with that, but I will double-check! I'm in agreement with you there.

.eslintrc
michaeljaltamirano: Yeah the only "undesirable" change in this PR was moving `__type` from the first property of `ThemeType` to the last, which I thought was okay.

.eslintrc
michaeljaltamirano: I think my concern is that this would be a particularly noisy rule. I am also obviously biased by virtue of leaving it off: I'm basically saying "If I have a limited amount of advocacy points to allocate, I personally don't think this rule is a priority for helping us write type-safe code that is less buggy." 

In my mind we already effectively distinguish between `readonly` and not by wrapping not `readonly` properties with `@observable`. Do you think that mental model is incomplete?

.eslintrc
michaeljaltamirano: I wonder if we could experiment with enforcing this only on `/models/**/*.ts`, and only using a `warn`? That would be a `PocketDerm`-only implementation. 

I think my concern is that while I understand it makes our code more documented/expressive, I don't know if adding that information would actually mean fewer bugs / confusion about functionality. Have you seen otherwise? (Or is that even a fair characterization?) I am just skeptical that the below is any meaningful difference:

```diff
export interface SelectorButtonProps {
+  readonly checked: boolean;
+  readonly children?: React.ReactNode;
+  readonly disabled?: boolean;
+  readonly icon?: React.ReactNode;
+  readonly onClick?: (
    event: React.MouseEvent<HTMLDivElement, MouseEvent> | React.KeyboardEvent,
  ) => void;
+  readonly selector?: SelectorType;
+  readonly size?: SizeType;
+  readonly type?: StyleType;
+  readonly [key: string]: unknown;
-  readonly checked: boolean;
-  readonly children?: React.ReactNode;
-  readonly disabled?: boolean;
-  readonly icon?: React.ReactNode;
-  readonly onClick?: (
    event: React.MouseEvent<HTMLDivElement, MouseEvent> | React.KeyboardEvent,
  ) => void;
-  readonly selector?: SelectorType;
-  readonly size?: SizeType;
-  readonly type?: StyleType;
-  readonly [key: string]: unknown;
}

File: src/utils/isDefined/index.ts [hunk 8]
```diff
@@ -0,0 +1,5 @@
+function isDefined<T>(arg: T | null | undefined): arg is T {
+  return arg !== undefined && arg !== null;
+}
```
joeyquarters, line 3: Curious if we have a preference for defining utility functions ourselves like this or if we should reach for something like [Lodash's `isNil`](https://lodash.com/docs/4.17.15#isNil)? This question could also be posed for the `isObject` utility function that we're defining as well.

joeyquarters (commented):
Great PR with detailed explanations! Looks good to me but I'll defer approval to others. Just had a small question around utility functions.

src/utils/isDefined/index.ts
michaeljaltamirano: @joeyquarters I would say that, as a team, we probably have a preference for reaching for existing packages. I'm happy to report that our team doesn't have a "Not Invented Here" problem. In `PocketDerm` we have a number of `lodash` packages--we just install the standalone ones instead of all `lodash`. We've also recently added `ramda` for folks who want to write more Functional Programming code.

The historical context of this repo is a little muddy because it all actually used to be located in `PocketDerm` until we spun it out into a separate repo (2+ years ago at this point, before I joined 🙂 ). So I don't know how we ended up with both a non-lodash `isobject` dependency and a defined `isObject` helper method, but the latter is now more useful to keep because we added the `val is Record<string, unknown>` type guard. The `isobject` type just returns `boolean` which isn't enough for our purposes. In fact, it looks like we can probably remove it now. But I wouldn't take that to be a NIH preference. 

We are, in general, pretty loose with accommodating people's preferences. For better or for worse most of our problems actually stem from having _too much choice 😅_ and so part of the impetus of this PR is to start limiting our choices in other areas where the benefits of consistency outweigh the cost of lack of flexibility. 

Hopefully this answers your question! Let me know if it does not.

joeyquarters: Thank you for the explanation! Yes, that answers my question 🙏

daigof (approved):
I trust your analysis. at some point we may want to move this to a shared repo like https://github.com/curology/eslint-config-curology

.eslintrc
```diff
@@ -39,13 +39,51 @@
     ],
```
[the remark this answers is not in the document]
michaeljaltamirano: Because extends loads it automatically, pretty much. Since it's redundant I don't think it's important to add. We add it for `emotion` because [emotion doesn't expose anything extensions/recommended set of rules/etc.](https://emotion.sh/docs/@emotion/eslint-plugin). I think that is clearer than adding other ones that we do not *need* to add.

michaeljaltamirano:
> I trust your analysis. at some point we may want to move this to a shared repo like https://github.com/curology/eslint-config-curology

Yep, already on it with some major version changes 🙂  https://github.com/curology/eslint-config-curology/pull/8. Next would be adding more common rules and opening PRs that handle those migrations/fixes. 

Will only be around for the first week of cooldown and I'll be getting this work into more workflows before going OOO.

.eslintrc
michaeljaltamirano: It doesn't look like this is a rule, and I think it's a niche enough case that we're fine not writing a custom rule for it 😅

.eslintrc
timmartin19: That's generally true.  I guess I might be a little raw, because I had a bug that would have been caught if I specified some properties as readonly...

michaeljaltamirano: @timmartin19 mind sharing me that PR/diff so I can be more informed?

timmartin19: Yeah, let me track it down.

timmartin19: https://github.com/curology/PocketDerm/pull/12364/files

Basically the fix was to wrap them in `@observable` but if I had marked them as `readonly` (because I didn't think they got updated), I would have gotten an error from typescript and known to convert them to observables.

timmartin19: Also, totally agree with the bias towards not overwhelming people.  And this rule would be very noisy.  It might be a "let's come back to it" kind of thing.

michaeljaltamirano: I see. That does seem like a good rule to apply to models to be able to more thoroughly check out usage. 

The only additional thing I'd note is that that rule is a `requires type information` rule. I added an `.eslint.tsconfig.json` file in this repo a little while ago, and this repo is small enough that perf isn't an issue, but I wonder how that would play out in our larger codebase(s).

PR #826, opened 2021-03-01
Export constants

matt-allan:
The `package.json` `main` field points to `dist/bundle.js` and the `module` field points to `dist/bundle-es/index.js`. These are from `src/index.ts`. The `index.ts` [only has a single export](https://github.com/curology/radiance-ui/blob/3f57bb79718862749dfa87d938488fff0b58d3d9/src/index.ts#L1), which exports the components. It doesn't export constants or icons.

Since only components are exported, that's all you can access when using the package entrypoint. If you need to use constants or icons, you have to import the paths directly and build the library for that environment yourself. From a cursory glance it looks like the constants are already in the bundle, they just aren't being exported.

I think the constants and the icons should probably both be exported, because they're listed in the documentation as being part of the public API. However exporting the icons bumps the bundle size to 1.1 MB, which fails bundlewatch. I think it's probably fine if the common.js bundle is large since anyone using that is probably either using node or a build tool that can treeshake, but I'm curious to hear everyone else's thoughts.

The use case for me is to be able to easily access the constants from node, so I can write them to a JSON file that can then be used in PHP templates. Basically I'm trying to do this:

```js
const radiance = require('radiance');

const { primaryTheme, secondaryTheme } = radiance;

// ...
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,1 +1,2 @@
 export * from './shared-components';
+export * from './constants';


michaeljaltamirano:
> ```js
> const radiance = require('radiance');
> 
> const { primaryTheme, secondaryTheme } = radiance;
> ```

@matt-allan  Does `const { primaryTheme } = require('radiance/lib/constants');` not work? 

I don't have super strong feelings about this export setup, but it's how it has been for as long as I've been here, I haven't had any personal pain around it, and the [documentation shows the nested path for constants](https://radiance-ui.curology.com/?path=/docs/constants-constants--available-constants).

daigof:
> I think the constants and the icons should probably both be exported, because they're listed in the documentation as being part of the public API.

I guess you're right, but if we add them to export then we will have 2 ways of retrieving the constants now.
The theme constants were supposed to be accessed via theme only. I guess the doc to access theme constants directly could be added here:
https://radiance-ui.curology.com/?path=/docs/constants-theme--border-radius

that would be something like 
```
import { primaryTheme } from 'radiance-ui/lib/constants';
...
primaryTheme.COLORS.white
```

sort of not the intended way to use it but can be added as an option.


If Michael suggestion doesnt work then we can still proceed adding constants to the main export but how will you deal with Icons?

matt-allan:
@michaeljaltamirano no, I get this error:

```
internal/modules/cjs/loader.js:883
  throw err;
  ^

Error: Cannot find module 'radiance/lib/constants'
```

Even if I could load that file, it's using ES modules, and [you can't use those with Node.js unless](https://nodejs.org/api/packages.html#packages_determining_module_system):
- the files end in .mjs.
- the package.json `type` is set to `module`

If it was a module I could import `./dist/bundle-es/constants/themes/primaryTheme.js` instead.

The only code that is available in common.js format is `./node_modules/radiance-ui/dist/bundle.js`, which is missing the export.

src/index.ts
```diff
@@ -1 +1,2 @@
 export * from './shared-components';
+export * from './constants';
```
michaeljaltamirano, line 2: Since we just need the theme, is there any objection to just `export { primaryTheme, secondaryTheme } from './constants';`?

michaeljaltamirano:
> @michaeljaltamirano no, I get this error:
> 
> ```
> internal/modules/cjs/loader.js:883
>   throw err;
>   ^
> 
> Error: Cannot find module 'radiance/lib/constants'
> ```
> 
> Even if I could load that file, it's using ES modules, and [you can't use those with Node.js unless](https://nodejs.org/api/packages.html#packages_determining_module_system):
> 
> * the files end in .mjs.
> * the package.json `type` is set to `module`
> 
> If it was a module I could import `./dist/bundle-es/constants/themes/primaryTheme.js` instead.
> 
> The only code that is available in common.js format is `./node_modules/radiance-ui/dist/bundle.js`, which is missing the export.

Duly noted. That makes sense. I've got no objection. Will look at bundlewatch stuff now. The numbers are relatively arbitrary, anyway.

michaeljaltamirano (approved):
You can adjust https://github.com/curology/radiance-ui/blob/master/bundlewatch.config.js#L5 as needed. Part of why it is set up the way it is is that if you track too many files, the slug doesn't work and the step fails. (It's a firebase limitation.)

I'm not worried about tracking the individual `constants` exports, so we can just bump up the bundle size accordingly.

src/index.ts
matt-allan: I think it would be helpful to have the other constants available for consistency but it's not as important as the themes.

I was just thinking we should export everything that's part of the public API so that the available exports don't differ depending on which module system you're using.

michaeljaltamirano: I'm cool with that. It looks like `dist` pretty much has all of them anyway, we just use relative imports within the library.

***

As an aside, one thing I've wanted to do is survey other major open source libraries and figure out what some "best practices" are that we're not doing. Blueprint, for example, uses the `@blueprint` namespace and has its exports from different paths/packages:

```tsx
import { Icon, Intent } from "@blueprintjs/core";
import { IconNames } from "@blueprintjs/icons";
```

I think *trying* to maintain that for us--with `lib/constants` and `/lib/icons` as the candidates--would be helpful, but I don't think it's a priority with the other stuff for us to get done this cooldown. I don't expect tons of people (anyone, really) to start pulling in the constants directly from `radiance-ui` after making these changes.

matt-allan:
> I guess you're right, but if we add them to export then we will have 2 ways of retrieving the constants now.
The theme constants were supposed to be accessed via theme only. 

@daigof I don't follow what you're saying. This PR isn't exporting any constants that aren't already exported in `src/constants/index.ts`. It looks like the only constants exported in that file are the ones [documented here](https://radiance-ui.curology.com/?path=/docs/constants-constants--available-constants). You can't access things like `COLORS` without using the theme. You _can_ access the theme constants by using the non-bundled code in `lib`, but that is already the case.

To clarify, this PR isn't exporting anything that isn't already exported elsewhere. What it's doing is making everything that's already accessible when using a bundler accessible when using the pre-bundled code.

For example, you can do this if you're using a bundler like Webpack or Rollup:

```js
```

That works because there's an ES module at `node_modules/radiance-ui/lib/constants.index.js`, and the bundler knows how to resolve it.

However, that doesn't work if you try to use:
1. The ES module in `dist/bundle-es/index.js` - used by [webpack and rollup](https://github.com/rollup/rollup/wiki/pkg.module) 
2. The common.js module in `dist/bundle.js` - used by node because we tell it to [here](https://github.com/curology/radiance-ui/blob/3f57bb79718862749dfa87d938488fff0b58d3d9/package.json#L5)
3. The UMD module in `dist/bundle.umd.js` - used by UMD I guess

...because they can only access the exports defined in `src/index.ts`. I think we should be exporting the same thing every time -- the module loader you're using shouldn't determine which parts of the public API are available.

matt-allan:
Closing in favor of #831.

PR #829, opened 2021-03-02
Consistent bundling

matt-allan:
This PR updates the rollup config and `package.json` metadata to ensure that the bundles work consistently, regardless of the module format, and to allow you to access every public entrypoint you can access from `lib` via the bundles.

This is a **breaking change** because it changes the modules you are allowed to access, and will cause consumers that were loading the code in `lib` that was transpiled by babel to use the code in `dist` that is transpiled by rollup + babel.

These changes allow the following code to work in the browser, in Node.js, and when using a bundler like Webpack:

```js
import { Button } from 'radiance-ui';
import { CameraIcon } from 'radiance-ui/lib/icons'
```

Notably the `lib` imports will use the rollup ESM modules in `dist` instead of the babel ESM modules in `lib`.

Imports of paths that aren't defined in `exports` will no longer work:

```js
// not allowed:
import { FlexContainer } from 'radiance-ui/lib/shared-components/optionButton/style';
import { primary } from 'radiance-ui/lib/constants/colors/primary';
```

We're doing that in a few places currently so those usages will need to be updated before this can be merged.

I tried to follow the [Webpack guidelines for package exports](https://webpack.js.org/guides/package-exports/#guidelines) and the [Node.JS suggestions](https://nodejs.org/api/packages.html#packages_approach_2_isolate_state) as much as possible. The setup is also very similar to [Preact's](https://github.com/preactjs/preact/blob/605bc5ec8804f031d9bbeff33d41c8c15ca13df5/package.json#L12).

This PR also drops support for UMD builds because they aren't able to use the same import paths as every other module system, requiring additional documentation, and UMD builds aren't as important anymore now that Node.JS and browsers support ES modules.


package.json
```diff
@@ -2,8 +2,30 @@
...
+  "module": "./dist/esm/index.js",
+  "exports": {
+    ".": {
+      "browser": "./dist/esm/index.js",
+      "import": "./dist/esm/index.js",
+      "require": "./dist/cjs/index.cjs"
+    },
+    "./lib/constants": {
+      "browser": "./dist/esm/constants/index.js",
+      "import": "./dist/esm/constants/index.js",
```
michaeljaltamirano, line 15: Should we move the changes in #826 in here as well so that this functionality is more "complete" or do we want two separate PRs to maintain the distinction?

michaeljaltamirano:
This looks good but can we pull in this work into a branch off of `master` instead of forked? It looks like CI hasn't run for this or #826, presumably because of the fork behavior.

package.json
matt-allan: The approach in this PR makes #826 unnecessary.

With the current setup the common.js version is a single file bundle, so the only way to include constants is to add them in the main index, which is what #826 does.

With the rollup changes in this PR we create multiple entrypoints for common.js and ES modules, so it's not necessary to add constants to the main index; we can just create a separate common.js module for them instead.

I think this approach is preferable since it allows you to import things in the same exact way regardless of module system, and you get to keep each bundle small.

michaeljaltamirano:
> This looks good but can we pull in this work into a branch off of `master` instead of forked? It looks like CI hasn't run for this or #826, presumably because of the fork behavior.

Sorry to be clear I meant pull in `radiance-ui` locally instead of a fork it. I didn't realize this repo had this behavior for forks until now.

matt-allan:
@michaeljaltamirano yeah, I think I have to close this one and open a new one though.

PR #831, opened 2021-03-03
Consistent bundling

matt-allan:
Note: I had to close #829 and open a new one to switch the branch.

This PR updates the rollup config and `package.json` metadata to ensure that the bundles work consistently, regardless of the module format, and to allow you to access every public entrypoint you can access from `lib` via the bundles.

This is a **breaking change** because it changes the modules you are allowed to access, and will cause consumers that were loading the code in `lib` that was transpiled by babel to use the code in `dist` that is transpiled by rollup + babel.

These changes allow the following code to work in the browser, in Node.js, and when using a bundler like Webpack:

```js
import { Button } from 'radiance-ui';
import { CameraIcon } from 'radiance-ui/lib/icons'
```

Notably the `lib` imports will use the rollup ESM modules in `dist` instead of the babel ESM modules in `lib`.

Imports of paths that aren't defined in `exports` will no longer work:

```js
// not allowed:
import { FlexContainer } from 'radiance-ui/lib/shared-components/optionButton/style';
import { primary } from 'radiance-ui/lib/constants/colors/primary';
```

We're doing that in a few places currently so those usages will need to be updated before this can be merged.

I tried to follow the [Webpack guidelines for package exports](https://webpack.js.org/guides/package-exports/#guidelines) and the [Node.JS suggestions](https://nodejs.org/api/packages.html#packages_approach_2_isolate_state) as much as possible. The setup is also very similar to [Preact's](https://github.com/preactjs/preact/blob/605bc5ec8804f031d9bbeff33d41c8c15ca13df5/package.json#L12).

This PR also drops support for UMD builds because they aren't able to use the same import paths as every other module system, requiring additional documentation, and UMD builds aren't as important anymore now that Node.JS and browsers support ES modules.


matt-allan:
Also need to look into this rollup warning:

```
(!) Mixing named and default exports
https://rollupjs.org/guide/en/#outputexports
The following entry modules are using named and default exports together:
src/shared-components/checkbox/index.tsx
src/shared-components/button/index.tsx
src/shared-components/radioButton/index.tsx
src/shared-components/button/components/roundButton/index.tsx

Consumers of your bundle will have to use chunk['default'] to access their default export, which may not be what you want. Use `output.exports: 'named'` to disable this warning
created dist/cjs in 15.3s
```

michaeljaltamirano:
> hared-components/checkbox/index.tsx

@matt-allan I've been moving this repo away from default exports in favor of named exports in almost all cases. I expect we can just remove the default exports in there and it'll be fine.

package.json
```diff
@@ -2,8 +2,30 @@
...
-  "module": "dist/bundle-es/index.js",
+  "main": "./dist/cjs/index.cjs",
+  "module": "./dist/esm/index.js",
+  "exports": {
+    ".": {
+      "browser": "./dist/esm/index.js",
+      "import": "./dist/esm/index.js",
+      "require": "./dist/cjs/index.cjs"
+    },
+    "./lib/constants": {
```
daigof, line 13: if we have these resolutions now, would it make sense to cleanup the below `files` options??? so they are not spit out as is in the package

michaeljaltamirano:
Oh that's a much longer build step now. Interesting. 

`(!) Entry module "src/icons/emojis/svgs/ElatedEmoji.tsx" is implicitly using "default" export mode, which means for CommonJS output that its default export is assigned to "module.exports". For many tools, such CommonJS output will not be interchangeable with the original ES module. If this is intended, explicitly set "output.exports" to either "auto" or "default", otherwise you might want to consider changing the signature of "src/icons/emojis/svgs/ElatedEmoji.tsx" to use named exports only.`
`https://rollupjs.org/guide/en/#outputexports`
`src/icons/emojis/svgs/ElatedEmoji.tsx`

We should probably suppress this warnings as well, I think they ultimately do not matter because of how we re-export them from `/icons`

package.json
matt-allan: I think the files is still necessary if you want 'go to definition' and Typescript types to work?

michaeljaltamirano:
@matt-allan Is the work that you were doing that necessitated these changes still on track for this cycle? I will be introducing some breaking changes this cycle and just want to make sure we don't end up blocking each other at any point 🙂

matt-allan:
@michaeljaltamirano it's going to  be a minute before I can work on this again. Feel free to make any changes necessary, it's easy enough for me to recreate the config now that I know how to do it.

matt-allan:
Closing in favor of smaller PRs.

PR #833, opened 2021-03-04
[Modals] - refactor out Paragraphs styling in Modals

daigof:
- updated doc
- Stories: refactor naming in DialogModal WithColor: refactor name and remove closeIcon logic since that has its own story

### DIFF
--- a/src/shared-components/dialogModal/__snapshots__/test.tsx.snap
+++ b/src/shared-components/dialogModal/__snapshots__/test.tsx.snap
@@ -76,14 +76,6 @@ exports[`<DialogModal /> renders dialog modal with custom color 1`] = `
   transform: translateY(100%);
 }
 
-.emotion-0 p {
-  margin-bottom: 1.5rem;
-}
-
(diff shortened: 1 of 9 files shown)


src/shared-components/dialogModal/__snapshots__/test.tsx.snap
```diff
@@ -76,14 +76,6 @@ exports[`<DialogModal /> renders dialog modal with custom color 1`] = `
   transform: translateY(100%);
 }
 
-.emotion-0 p {
```
michaeljaltamirano, line 79: Let's wrap the `modalBody` content in these tests with the new `Paragraph` component so that usage is canonical/more documented.

daigof: ok Im good with that

michaeljaltamirano (approved):
🙏 Let's just be sure to approve on chromatic and re-build and confirm things are fine in CI before merging in.

PR #848, opened 2021-03-10
Account for border and margin for Accordion child elements

jashieh:
## Background
If the child element put in `Accordion` component had `margin-top` or `margin-bottom` set, the `Accordion` would not account for that additional space when calculating its `max-height` while open. This caused issues where the inner content would be clipped when the `Accordion` was open.

## In this PR:
- Change `clientHeight` to `offsetHeight` to account for element border height.
- Add element's `margin-top` and `margin-bottom` into total height calculation.

## Example of clipping issue
Screen Shot 2021-03-09 at 4 36 17 PM

### DIFF
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -63,9 +63,15 @@ export const Accordion = ({
   const contentRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
+    const refStyles =
+      contentRef.current && window.getComputedStyle(contentRef.current);
+    const marginTopBottom = refStyles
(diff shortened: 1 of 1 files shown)


daigof (commented):
Thanks for coding the solution structure.
However there may be some cases not accounted for. I was running some tests and what would happen if some value is `NaN` => it would impact the current rendering

image

As I pointed out in our meeting we need to be extra sure we are not impacting the current app since Accordion is used widely in our apps.

We can do these:
1- See if you can account for NaN results to avoid mess up with the current calculation of nextHeight 
2- Release a beta version and test in PocketDerm. to do this just run `yarn run publish-package` in your branch (not master), see previous beta releases for naming convention. Test the beta version in PocketDerm, run cypress, etc, if everything OK we can approve this PR

PR #884, opened 2021-03-25
[Typography] Add Body Bold Typography

### LINKED ISSUE #808: Add "Body Bold" Typography
Screen Shot 2021-02-22 at 5 40 10 PM

Screen Shot 2021-02-22 at 5 40 02 PM

Not solved by just adding a `<strong>` tag around it, since the `primary` theme version is a darker `primary` color, so we will need to add proper handling for it in our `FONTS` theming.

michaeljaltamirano:
Fixes #808

### DIFF
--- a/src/shared-components/typography/index.ts
+++ b/src/shared-components/typography/index.ts
@@ -43,6 +43,12 @@ const bodyStyle = (theme: ThemeType) => `
   ${baseBodyStyles(theme)}
 `;
 
+const bodyBoldStyle = (theme: ThemeType) => `
+  ${bodyStyle(theme)}
+  font-weight: ${theme.TYPOGRAPHY.fontWeight.bold};
+  color: ${theme.COLORS.primary};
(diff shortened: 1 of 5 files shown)


src/shared-components/typography/index.ts
```diff
@@ -43,6 +43,12 @@ const bodyStyle = (theme: ThemeType) => `
   ${baseBodyStyles(theme)}
 `;
 
+const bodyBoldStyle = (theme: ThemeType) => `
+  ${bodyStyle(theme)}
+  font-weight: ${theme.TYPOGRAPHY.fontWeight.bold};
```
michaeljaltamirano, line 48: `font-weight` added for direct `css` usage if we need to use it instead of `Typography.Bold` (which is `<strong>`)

michaeljaltamirano:
Will merge in *after* next major release, since this can be a minor.

PR #904, opened 2021-03-31
[TypeScript] Use type-only import syntax

michaeljaltamirano:
Will clean up the diff in #903, and appears to potentially be required for next-gen tooling like ESBuild: https://github.com/storybookjs/storybook/pull/14380/files


michaeljaltamirano:
Not super sure why there's a bit of a size change. The `dist` output for `Typography` looks the same to me. But merging in regardless.

PR #905, opened 2021-04-01
[OptionButton] Add OptionButton.NotClickable presentational component

michaeljaltamirano:
One of the breaking changes in #903 is removing the ease of importing styled components that were never really intended to be used directly. One particular instance is used for our recommendations module:

Screen Shot 2021-04-01 at 4 25 54 PM

The markup imports *all* of the styled components that `OptionButton` uses besides the outerlying `button`. 

The branch name is `container/add-icon-container` because I first sought to replicate this behavior in the `Container` component, so that it would be clearly presentational and not functional (unlike a `*Button`), but there was enough CSS specific to `OptionButton` that I decided to just extend the behavior of `OptionButton`. 

The PR basically extracts the internals of `OptionButton` into a reusable component itself, and then defines an `OptionButtonContainer` with a div (rather than the `OptionButton` button), and exposes it as `OptionButton.NotClickable`. The styles are similarly non-additive: I extracted all non-focus styling into a shared variable to be used for both `OptionButton` and `OptionButtonNotClickable`, keeping the click affordances only on `OptionButton`. 

@benkolde: My assumption is that it makes sense to have this as a design asset, but the design asset (as currently put together) lacks the fidelity to know that we do not want to render a button (much less one that has clickable affordances) when it's in fact not supposed to behave like a button. 

Here's the diff in usage:

```diff
-    <FlexContainer>
-      <IconWrapper selected={false}>{icon}</IconWrapper>
-      <TextContainer>
-        <Text>{mapTextToCopy(text, optionId)}</Text>
-      </TextContainer>
-    </FlexContainer>
+    <OptionButton.NotClickable
+      optionType="radio"
+      icon={icon}
+      text={mapTextToCopy(text, optionId)}
+    />
   ```
   
[You can view the file in question in an open PR here.](https://github.com/curology/PocketDerm/pull/12738/files#diff-a2dd036f672155b18f85b3c335791a93b22b0b49590d6d9336699f2f8a1226c2)

### Other

- Updates the existing OptionButton story to capture more states in visual regression testing (e.g. `selected` + `secondary`).

### CODE REVIEW

File: src/shared-components/container/style.ts [hunk 1]
```diff
@@ -19,14 +19,16 @@ const messageStyle = (theme: ThemeType) => `
...
-export type ContainerType = 'clickable' | 'message';
+export type ContainerType = 'clickable' | 'message' | 'none';
 
 const getContainerTypeStyles = (theme: ThemeType, type?: ContainerType) => {
   switch (type) {
     case 'message':
       return messageStyle(theme);
     case 'clickable':
       return clickableStyle(theme);
+    case 'none':
```
michaeljaltamirano, line 30: Explicitly encoding the non-message and non-clickable style in a type, in the event that we continue building on this and the behavior of the four cases change.

src/shared-components/button/components/linkButton/index.tsx
```diff
@@ -8,7 +8,7 @@ import { ButtonContents, ButtonText } from '../../style';
 import Style from './style';
 import { COLORS_PROP_TYPES, ThemeColors } from '../../../../constants';
 
-export interface LinkProps {
+export interface LinkButtonProps {
```
michaeljaltamirano, line 11: Small unrelated change for consistency/clarity

src/shared-components/optionButton/index.tsx
```diff
@@ -4,17 +4,15 @@ import PropTypes from 'prop-types';
...
 export interface OptionButtonProps {
-  borderRadius?: string;
+  borderRadius?: valueof<ThemeType['BORDER_RADIUS']>;
   buttonType?: 'primary' | 'secondary';
   /**
    * Show custom icon in the unselected state
    */
   icon?: React.ReactNode;
-  /**
-   * Label
```
michaeljaltamirano, line 16: This appears to have been accidentally committed as it makes no sense in the storybook docs page for `onClick`.

src/shared-components/optionButton/index.tsx
```diff
@@ -4,17 +4,15 @@ import PropTypes from 'prop-types';
 import Style from './style';
 import { CheckmarkIcon } from '../../icons';
 import { isDefined } from '../../utils/isDefined';
+import { ThemeType } from '../../constants';
 
 export interface OptionButtonProps {
-  borderRadius?: string;
+  borderRadius?: valueof<ThemeType['BORDER_RADIUS']>;
```
michaeljaltamirano, line 10: I considered removing this prop entirely--it is not currently used--but decided against it (for now). This will make sure we only pass in `theme`-consistent values. However, it does not check if it's appropriate for the theme, e.g. you can forcibly pass in `'16px'` for Agency and it won't raise any errors.

src/shared-components/optionButton/index.tsx
```diff
@@ -43,35 +98,49 @@ export const OptionButton = ({
...
+      subtext={subtext}
+      text={text}
+    />
+  </Style.DisplayContainer>
+);
+
+/**
+ * Similar OptionButton styling without click elements
+ */
+OptionButton.Container = OptionButtonContainer;
```
michaeljaltamirano, line 142: I'm open to renaming this, since `Container` is typically used in our codebase (like in the `Accordion`) to wrap the component and apply uniform styling. I wouldn't even be against something extremely clear, like `OptionButton.NoClickFunctionality`. I would prefer it be a separate component like that than a prop.

ipc103: Good call - I would vote for the explicit name over `Container`. Maybe `OptionButton.NotClickable`?  

I'm not sure if there's a name that matches this UI concept - semantic-ui-react has a `Segment` [component](https://react.semantic-ui.com/elements/segment/) that looks kind of similar, but I think that's a different concept.

michaeljaltamirano: I think there is a way to extract a more obviously non-functional element out of this by expanding our main `Container` component. The problem is the usage (and presentation) matches so obviously to the `OptionButton` that it seems obfuscating to treat it as otherwise.

Given this implementation is on-the-go I'm comfortable waiting until we see a clearer pattern emerge from usage before placing the logic outside of the `OptionButton`.

ipc103 (approved):
💯  - thanks for making the change, definitely a good improvement.

PR #910, opened 2021-04-05
[Dependencies] Update Emotion 10 to 11

michaeljaltamirano:
### What and Why

This PR updates our `@emotion` dependency to use version 11 instead of version 10. [See docs for more details.](https://emotion.sh/docs/emotion-11) 

1. Replace all version 10 dependencies with their version 11 equivalents (often with a new namespace, e.g. `eslint-plugin-emotion` is now `@emotion/eslint-plugin`)
1. Run `@emotion/pkg-renaming` eslint rule (with `--fix`) as codemod
1. Remove version 10 TypeScript `paths` usage in favor of new [emotion.d.ts standard](https://emotion.sh/docs/typescript#define-a-theme)
1. Ensure `storybook` is compatible with our version 11 usage (since it still relies on version 10 internally)
1. Fix a `babel` warning to make an implied setting explicit
1. Run `yarn run test -u` to update the snapshots with the new serializer logic. 
    - Most of the lines of code in this PR are just diffs in snapshot tests. Our Chromatic visual UI tests show no regression, which is the more important signal than already hard-to-read snapshots. 
    - You can see a trivial example of the diff in [src/shared-components/toggle/__snapshots__/test.tsx.snap](https://github.com/curology/radiance-ui/pull/910/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.snap&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&file-filters%5B%5D=.yml&file-filters%5B%5D=dotfile&hide-deleted-files=true#diff-17add0c1f6ee2ac9f27ec542a2ddb149613c24f6e2bc0831e0d81c7f80b154cc), which just replaces the `emotion` classnames (e.g. `emotion-0`) with other classnames.
        - In more complex snapshots, the order of the classnames in the stylesheet also changes more significantly, but as mentioned the visual regression tests are our main source of truth for whether or not a change is changing behavior. 

### Skipped Snapshot Tests

This PR has been so long-standing because there has been a snapshot serialization regression in CI that I've tried long and hard without success to debug. It happens as follows:

1. *Some* snapshot tests encounter an issue where *usually only* for the first snapshot test are the serialized styles in CI different than locally.
    - Some tests, like `src/shared-components/field/test.tsx`, fully failed in CI.
1. For tests for which only the first test fails, _when commenting out the first test_, it ends up being the second snapshot that fails. 
1. Even SSHing into the CI build, copying the entire `radiance-ui` repository from that SSH instance to my local disk, and then running the test suite locally was not able to reproduce the issue. 

I'm not sure why I can't find instances of this in CI, but here is a link where you can see the failing snapshots, and see that the issue is that the serialization order is changing very insigificantly, but nothing else: https://app.circleci.com/pipelines/github/curology/radiance-ui/2469/workflows/65276136-7360-4103-9e7e-41dff0a89393/jobs/7590/parallel-runs/0/steps/0-105

**UPDATE: I actually ran into this again while trying to reduce the diff to make the PR easier to review, and also to remove all trace of my attempts to work around the issue.** Interestingly, just changing the syntax of applying styles conditionally ([PR diff](https://github.com/curology/radiance-ui/pull/910/commits/d9a2b75b0ecee568d3bde588fb13597ae1c8a3ec), to _fix_ the issue) broke the order of the serialized styles ([CI link](https://app.circleci.com/pipelines/github/curology/radiance-ui/2653/workflows/5667b453-23bb-43d4-957e-30d77ab5e98a/jobs/8456/parallel-runs/0/steps/0-105)). This suggests that the fix to the other issues might be similarly more explicitly falsy/nullish handling. [One additional test on the Accordion component was not successful](https://github.com/curology/radiance-ui/commit/51b9f12a25be2698331fcdd5eb80e71f13889b46), though, so the instances that cause it may be non-obvious.

Ultimately, I decided to skip these snapshot tests entirely. Now that we have visual regression testing, they are already of diminished value. I also kept the non-snapshot tests in the component tests, if present. 

### Considerations

1. To make migrations like this in the future earlier, we should consider wrapping css-in-js usage around a helper file that exports the methods we need. 
    - This applies to other 3rd party libraries moreso than `@emotion`, since we rely on it so heavily, but doing so would also allow us to more easily experiment with other 3rd party libraries that have no runtime element, like [Linaria](https://github.com/callstack/linaria).
    
### Release Plan

This PR is based to `next`. After it is merged in, I will cut a beta version of `radiance-ui@^22.x.x`. I will pull this into my own consumer app for testing, and it can also be pulled into our `gatsby` pages for testing. This will allow us to:

1. Merge in minor and patch version changes to `radiance-ui@^21.x.x` without emotion major version changes blocking in.
1. Allows us to pull in any patches required after this specific PR to be part of `22.0.0` instead of immediately patching it.


AlyssaKirstine:
I'm so excited for this! @michaeljaltamirano you got it figured out so soon. I'll take a look sometime soon, but ultimately, I'll leave it up to @chrisq21 to approve as the Gatsby v3 upgrade is now relevant to his cycle project.

src/shared-components/button/style.ts
```diff
@@ -135,12 +135,6 @@ const disabledStyles = (theme: ThemeType) => `
   color: ${theme.COLORS.textDisabled};
   cursor: not-allowed;
   fill: ${theme.COLORS.textDisabled};
-
-  &:visited,
-  &:hover {
-    opacity: 1;
```
chrisq21, line 141: So we no longer want to change the `opacity` and `color` on `visited` and `hover`?

chrisq21: Ah never mind, this makes sense since those values weren’t actually being changed.

michaeljaltamirano: Yep, and it seems like the redundant styling, paired with how we apply conditional styles, is part of what is causing the CI issues. 🤷

.storybook/main.ts
```diff
@@ -53,8 +56,25 @@ module.exports = {
       );
     }
 
+    /**
+     * Until Storybook migrates its own internal @emotion usage from
+     * v10 to v11, this allows us to maintain compatibility
+     */
+    const emotion11CompatibleConfig = {
```
chrisq21, line 63: Oof, nice discovery.

src/types/emotion.d.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference types="@emotion/react/types/css-prop" />
+import '@emotion/react';
+
+import { ThemeType } from '../constants/themes/types';
+
+declare module '@emotion/react' {
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  export interface Theme extends ThemeType {}
```
chrisq21, line 8: Do you know if this will have any influence on how we use `ThemeType` on the Gatsby sites?

Specifically [here](https://github.com/curology/curology-gatsby/blob/main/src/types/modules/%40emotion/styled/index.ts) and [here](https://github.com/curology/curology-gatsby/blob/main/src/types/modules/emotion-theming/index.ts)

michaeljaltamirano: Yep, we no longer need those files (this PR removes those equivalents here). This file will be the new standard: https://emotion.sh/docs/typescript#define-a-theme

chrisq21 (approved):
This looks good! Thanks for setting up this update so quickly. 

We'll be doing a lot of regression testing on the Gatsby V3 update, so I'll let you know if we find any `radiance-ui` or `emotion` related issues.

Side-note: I wasn't aware of the Chromatic visual UI tool. That sounds really useful.

PR #983, opened 2021-05-24
[TypeScript] Prefer React.FC For Component Typing

michaeljaltamirano:
### What & Why

Upgrading a consumer app from Emotion 10 to 11 surfaced a number of PropTypes type errors related to PropTypes definitions:
Screen Shot 2021-05-24 at 11 14 17 AM

After looking into the issue, using `React.FC<Props>` type definitions on the Radiance components surfaced PropTypes issues where there were differences in specificity between the TypeScript definitions and the PropTypes definitions. Interestingly, I thought these checks were already occurring, but that does not seem to be the case. 

This PR adds `React.FC` to all appropriate components and upgrading the PropTypes definitions accordingly. 

### Other 

1. Export `RadianceIconComponent` type for use in consumer apps.

### DIFF
--- a/src/shared-components/alert/index.tsx
+++ b/src/shared-components/alert/index.tsx
@@ -53,7 +53,7 @@ interface Alert extends React.FC<AlertProps> {
  *
  * All alerts are dimissable by clicking on them. However, you can use the `duration` prop to determine if the alert is sticky or dismissed on a timer (in units of seconds).
  */
-export const Alert: React.FC<AlertProps> = (alertProps) => {
+export const Alert: Alert = (alertProps) => {
   const {
     avatarSrc = '',
     content,
@@ -139,6 +139,8 @@ export const Alert: React.FC<AlertProps> = (alertProps) => {
     </Style.AlertContainer>
   );
 };
+
+Alert.Container = Style.AlertsContainer;
 
 Alert.propTypes = {
   avatarSrc: PropTypes.string,
--- a/src/shared-components/alert/index.tsx
+++ b/src/shared-components/alert/index.tsx
@@ -30,6 +30,10 @@ export interface AlertProps {
   [key: string]: unknown;
 }
 
+interface Alert extends React.FC<AlertProps> {
+  Container: typeof Style.AlertsContainer;
+}
+
 /**
  * Alerts should be used to show notifications or messages from (providers, support, or system).
  *
@@ -49,7 +53,7 @@ export interface AlertProps {
  *
  * All alerts are dimissable by clicking on them. However, you can use the `duration` prop to determine if the alert is sticky or dismissed on a timer (in units of seconds).
  */
-export const Alert = (alertProps: AlertProps) => {
+export const Alert: React.FC<AlertProps> = (alertProps) => {
   const {
     avatarSrc = '',
     content,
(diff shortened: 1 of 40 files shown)


src/shared-components/optionButton/index.tsx
```diff
@@ -115,43 +118,44 @@ export const OptionButton = ({
...
 OptionButton.NotClickable = OptionButtonNotClickable;
 
 OptionButton.propTypes = {
-  borderRadius: PropTypes.string,
+  borderRadius: BORDER_RADIUS_PROP_TYPES,
   buttonType: PropTypes.oneOf(['primary', 'secondary']),
   icon: PropTypes.node,
   onClick: PropTypes.func.isRequired,
-  optionType: PropTypes.oneOf(['radio', 'checkbox']).isRequired,
+  optionType: PropTypes.oneOf(['radio', 'checkbox'] as const).isRequired,
```
michaeljaltamirano, line 158: It is also unclear to me why `buttonType` does not get flagged by `optionType` does, if `as const` directive is not used. 🤷

src/shared-components/icon/index.tsx
```diff
@@ -43,7 +47,8 @@ Icon.propTypes = {
   displayInline: PropTypes.bool,
   fill: PropTypes.string,
   height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
-  IconComponent: PropTypes.elementType.isRequired,
+  // @ts-expect-error -- TypeScript type is too specific for PropTypes types
+  IconComponent: PropTypes.element.isRequired,
```
michaeljaltamirano, line 51: Could not figure this out, but since it is only ever used internally ignoring it doesn't matter for consumer app usage.

src/shared-components/container/style.ts
```diff
@@ -65,20 +64,11 @@ const Image = styled.img`
...
-type StyledContainer = StyledComponent<
-  DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>,
-  {
-    type?: ContainerType;
-  },
-  { theme?: ThemeType }
->;
-
-type CompositeContainer = StyledContainer & {
+const ContainerComponentForTypesOnly = styled.div<{
```
michaeljaltamirano, line 67: Rather than import the "exotic" types used, we just set up a variable to use the type definition. I was not able to use the value directly without first setting it in a variable.

src/shared-components/button/components/linkButton/index.tsx
```diff
@@ -11,9 +11,9 @@ import type { ButtonType } from '../../types';
 
 export interface LinkButtonProps {
   /**
-   * Specifies the tag or element to be rendered
+   * Specifies the tag or element to be rendered, like an 'a' or 'span'
    */
-  as?: 'a' | React.ElementType;
+  as?: string | React.ElementType;
```
michaeljaltamirano, line 16: The string type literal was being overwritten, so the type was `React.ElementType | undefined`. We also use `span` so the literal definition was incomplete, anyway. This definition is not super strict but should be fine for now.

src/shared-components/alert/index.tsx
```diff
@@ -136,9 +140,7 @@ export const Alert = (alertProps: AlertProps) => {
   );
 };
 
-Alert.Container = ({ children }: { children: React.ReactNode }) => (
```
michaeljaltamirano, line 139: Since this just accepts a node, and a node is overly permissive, I don't think there's any difference between this and using the styled component directly.

PR #1005, opened 2021-06-02
Improve valid HTML for SelectorButton

joeyquarters:
This change fixes `validateDOMNesting` console errors thrown by React in the PocketDerm repo.

Consumers of the `RadioButton` or `Checkbox` components may sometimes require more complicated DOM structures than plain text to describe the input choice. For example, in [our `SetCustomization` component](https://github.com/curology/PocketDerm/blob/master/resources/assets/jsx/components/patient/setCustomization/index.tsx), we have the following:

Screen Shot 2021-06-02 at 2 39 58 PM

This structure currently employs usage of `div` elements to create a flexbox layout. It also uses the `Chip` component, which utilizes a `div`. My first instinct was to maintain the `<p>` usage and convert any children to [valid phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content), but I think this change is more flexible for a variety of consumers (despite my constant urge to slash seemingly unnecessary `<div>`'s).

### DIFF
--- a/src/shared-components/selectorButton/style.ts
+++ b/src/shared-components/selectorButton/style.ts
@@ -104,7 +104,8 @@ const Selector = styled.div<{
   }
 `;
 
-const TextContainer = styled.p`
+const TextContainer = styled.div`
+  flex-grow: 1;
   margin-left: ${SPACER.medium};
   margin-top: ${SPACER.xsmall};
   min-width: 125px;
--- a/src/shared-components/selectorButton/__snapshots__/test.tsx.snap
+++ b/src/shared-components/selectorButton/__snapshots__/test.tsx.snap
@@ -100,6 +100,10 @@ exports[`<SelectorButton /> UI snapshots when Icon added displays check mark for
 }
 
 .emotion-9 {
+  -webkit-box-flex: 1;
+  -webkit-flex-grow: 1;
+  -ms-flex-positive: 1;
+  flex-grow: 1;
   margin-left: 1rem;
   margin-top: 0.25rem;
   min-width: 125px;
@@ -137,11 +141,11 @@ exports[`<SelectorButton /> UI snapshots when Icon added displays check mark for
       type="primary"
     />
   </div>
-  <p
+  <div
     class="emotion-9 emotion-10"
   >
     SelectorButton Text
-  </p>
+  </div>
 </div>
 `;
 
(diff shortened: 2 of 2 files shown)


src/shared-components/selectorButton/style.ts
```diff
@@ -104,7 +104,8 @@ const Selector = styled.div<{
   }
 `;
 
-const TextContainer = styled.p`
+const TextContainer = styled.div`
+  flex-grow: 1;
```
michaeljaltamirano, line 108: Is this preventing a regression within this component library, or something for convenience in userland?

michaeljaltamirano (approved):
I am comfortable with this change: the error in the console is very noisy, but I just haven't prioritized fixing it over other issues, so thank you for the initiative. 

My main hesitation is that being confident that this change is harmless will require some clever global searches in userland to make sure we're not relying on any particularly nested `p` styling in components built around this. There's also the fact that Radiance's current version is on Emotion 11, which has only been released in one of our `gatsby` repos. So you may either want to branch this off of https://github.com/curology/radiance-ui/tree/v21 and do a minor/patch release or wait until cooldown at which point the main app will be on Emotion 11 , or you can shepherd the PocketDerm upgrade PR while I'm OOO.

src/shared-components/selectorButton/style.ts
joeyquarters: I _thought_ it was preventing a regression when changing DOM elements, but it turns out I was just [breaking a nested CSS selector](https://github.com/curology/PocketDerm/blob/master/resources/assets/jsx/components/patient/setCustomization/selectors/moisturizerSelector/style.ts#L58-L60). I'll remove the `flex-grow` rule, as that particular styling is currently achieved via the consumer of the component.

michaeljaltamirano: We can also add a className to that inner div (or some other easily identifiable token) so that nested selectors like that in user land are less brittle. Sorry 😞

joeyquarters:
Ah, I wasn't aware of the Emotion version discrepancy. This PR is solely an attempt to quiet some noise (and improve some HTML!) on a component I will be working on this cycle, so it's not urgent.

I won't have enough bandwidth to shepherd the PocketDerm upgrade this cycle, though it would be interesting to work on! If it's alright with you, I think I'll let this hang out for a couple weeks until we get Emotion updated in Pocketderm, then hopefully I can do a more comprehensive analysis of this change prior to merging.

michaeljaltamirano:
> Ah, I wasn't aware of the Emotion version discrepancy. This PR is solely an attempt to quiet some noise (and improve some HTML!) on a component I will be working on this cycle, so it's not urgent.
> 
> I won't have enough bandwidth to shepherd the PocketDerm upgrade this cycle, though it would be interesting to work on! If it's alright with you, I think I'll let this hang out for a couple weeks until we get Emotion updated in Pocketderm, then hopefully I can do a more comprehensive analysis of this change prior to merging.

Yep, that sounds good. Again, thank you for the initiative on this 🙏🏼

daigof (approved):
👍

PR #1040, opened 2021-06-28
Native ES modules

matt-allan:
This PR converts the Radiance repo over to native ES modules. [Every supported version of Node supports native ES modules](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77) so we don't need to use Common JS anymore.

By default every `.js` file will now be parsed as an ES module. If you need the file to be parsed as a CommonJS module (i.e. because you're using a package that only supports CJS) you can use the `.cjs` extension.

We can't drop the CommonJS **bundle** yet because the `@emotion/styled` package is not fully setup for ES modules yet and breaks when you try to import `@emotion/styled/base`. You can repro it with this script:

```
import { Button } from './dist/bundle-es/index.js';

console.log(Button);
```

```js
$ node testing.js
node:internal/process/esm_loader:74
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/Users/matt/code/radiance-ui/node_modules/@emotion/styled/base' is not supported resolving ES modules imported from /Users/matt/code/radiance-ui/dist/bundle-es/shared-components/typography/index.js
Did you mean to import @emotion/styled/base/dist/emotion-styled-base.cjs.js?
```

In the interim if you need to use Radiance components from node you still have to use Common JS:

```js
const radiance = require('./dist/bundle.cjs');

const { Button } = radiance;

console.log(Button);
```

```
$ node testing.cjs
[Function: Button] {
  Container: <ref *1> [Function (anonymous)] {
    displayName: 'ButtonContainer',
    defaultProps: undefined,
    __emotion_real: [Circular *1],
    __emotion_base: 'div',
    __emotion_styles: [
      'label:ButtonContainer;',
   // ...
```


matt-allan:
Opened https://github.com/emotion-js/emotion/issues/2413 for the Emotion issue. I don't think we can fix it on our end.

matt-allan:
Going to close this until we are unblocked by Emotion.

PR #1041, opened 2021-06-28
Drop support for UMD modules

matt-allan:
This isn't really necessary now that ES modules are supported everywhere. If someone does need it they can always transpile the module themselves.


michaeljaltamirano (approved):
🚀

PR #1044, opened 2021-06-29
Use the babel runtime helpers instead of bundling

matt-allan:
This PR switches our Babel builds from bundling the helpers to depending on the `@babel/runtime` package instead. The `@babel/runtime` package is already a dependency of Radiance's dependencies (i.e. `react-transition-group`), so we can reduce the bundle size a bit by not bundling another copy of the helpers as part of Radiance.

See https://github.com/rollup/plugins/tree/master/packages/babel#helpers and https://babeljs.io/docs/en/babel-plugin-transform-runtime for more info.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -26,6 +26,7 @@ const getPresets = (modules) => [
 ];
 
 const plugins = [
+  '@babel/plugin-transform-runtime',
   '@babel/plugin-proposal-export-namespace-from',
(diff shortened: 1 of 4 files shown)


matt-allan:
I need to tweak the `external` setting for Rollup to account for deep imports from `@babel/runtime/*`. Going to make a separate PR for that.

PR #1048, opened 2021-06-30
Export `injectGlobalStyles`

matt-allan:
This is necessary to use the themes, so we need to export it. We're using 'deep imports' from the filesystem at the moment to access them.

Longer term since we're always passing these in to the `Global` component, should we export a component instead? I.e.

```jsx
// in Radiance

export BrandStyles = <Global<ThemeType> styles={brandStyles} />;
export ResetStyles =  <Global styles={resetStyles} />;

// in the consumer

const Layout = ({ children }: LayoutProps) => (
  <>
    <ResetStyles />
    <BrandStyles />
    <MainContainer id="main-container">
      <PageContainer>{children}</PageContainer>
    </MainContainer>
  </>
);
```

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -3,3 +3,4 @@ export { default as throwOnUndefinedProperty } from './throwOnUndefinedProperty'
 export { default as withDeprecationWarning } from './withDeprecationWarning';
 export { default as isObject } from './isObject';
 export { applyPrimaryThemeVerticalOffset } from './themeStyles';
+export * from './injectGlobalStyles/style';


michaeljaltamirano (approved):
It looks like we do not deviate from the usage in your PR description in our main app, I did not check but I imagine it's the same for the `gatsby` pages. 

I think moving that usage into a new component makes sense. It's possibility some degree of flexibility was desired, or maybe it was laziness/not wanting to tighten up the component library API, but at present time we should pretty much always prefer tightening.

PR #1051, opened 2021-06-30
Create components for global styles

matt-allan:
This PR creates components for the global styles so we can use those instead of grabbing the CSS from `lib/utils` and combining it with the `Global` component from Emotion every time.


michaeljaltamirano (approved):
🚀

PR #1057, opened 2021-07-01
Move built assets from dist to lib

matt-allan:
My last release broke Type resolution 😢

TypeScript cannot parse conditional exports, so it's unable to resolve types for paths like `radiance-ui/lib/constants` if they're actually in `radiance-ui/dist/constants`. Using the same name for the built assets ensures TypeScript can resolve the path correctly without having to parse the conditional exports.

We _could_ build the types to `lib` and the bundles to `dist` but that seemed confusing.

See https://github.com/microsoft/TypeScript/issues/33079

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,7 +76,7 @@ typings/
 **/.DS_Store
 
 # build files
-dist/
+lib/
 .out
 .size-snapshot.json
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+# Unreleased
+
+- [Internal] Moved built files from `dist` to `lib`. This is neccessary for TypeScript to properly resolve paths like `radiance-ui/lib/constants`, because TypeScript is unable to parse conditional exports.
+
 # v23.0.0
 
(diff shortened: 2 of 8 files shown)


michaeljaltamirano:
> My last release broke Type resolution 😢
> 
> TypeScript cannot parse conditional exports, so it's unable to resolve types for paths like `radiance-ui/lib/constants` if they're actually in `radiance-ui/dist/constants`. Using the same name for the built assets ensures TypeScript can resolve the path correctly without having to parse the conditional exports.
> 
> We _could_ build the types to `lib` and the bundles to `dist` but that seemed confusing.
> 
> See [microsoft/TypeScript#33079](https://github.com/microsoft/TypeScript/issues/33079)

Do you suspect this is kind of just working around the problem of clumsy import syntax? e.g., shouldn't we probably just tear off the band-aid and omit `/lib` from the import paths and instead prefer something like `radiance-ui/constants`? This would be more similar to other libraries like `rxjs/operators`. 

FWIW the migration in our main app would be painless: we can already apply `eslint --fix` across the entire repository without being blocked by errors from old code. I can't say the same for our site repos.

michaeljaltamirano (approved):
Approving with the understanding that this satisfies the requirements in the short-term, and we can kick the can of more breaking changes to a future major version

matt-allan:
@michaeljaltamirano yes I agree that would be preferable long term. If the paths were simpler we wouldn't need export map rewriting at all.

matt-allan:
Actually we could add conditional exports to export both versions to make the migration easier, then deprecate the `lib` versions in a future release 🤔

PR #1080, opened 2021-07-19
[Glyphs] Add MelasmaGlyph

michaeljaltamirano:
Used for secondary theme only, otherwise returns null. 

Screen Shot 2021-07-19 at 1 15 14 PM

Commit history is long because I thought about adding some tool before realizing it wasn't an appropriate use case 🤡


MonsterEyes (approved):
Thanks @michaeljaltamirano 🙏🏼

PR #1110, opened 2021-08-06
Add aria-label to dropdown component

meron-yemane:
This dropdown component was causing 'Accessibility check failed!' errors in my cypress tests. Below is a screenshot of the produced error. Simple fix was just to add an aria-label to the component. 
Screen Shot 2021-08-06 at 9 41 32 AM

PR #1126, opened 2021-08-11
[Dropdown] Add optional id prop

michaeljaltamirano:
This PR continues the work of #1110 by adding an optional id prop so that we can associate `<label for="unique-field-identifier" />` with the Dropdown by passing the id to it, e.g. `<select id="unique-field-identifier" />`. We previously removed ID in #307 because it was not unique, but this ID is meant to be used in conjunction with our existing form components. 

Future PRs can look into more robust a11y handling via `react-aria` and consideration of orchestrating certain prop configurations as non-optional so as to raise type errors in accordance with a11y errors.


meron-yemane (approved):
🤘🏾

PR #1159, opened 2021-09-15
[ImmersiveModal] Update Markup and Focus Handler

michaeljaltamirano:
### CHANGELOG

1. Add bespoke focus handling via `@react-aria`'s `useFocusManager` to `ImmersiveModal` component via `ImmersiveModalContent` sub-component.
1. ~~Move definitions in `style.ts` that only apply to `ImmersiveModalContent` markup into `/immersiveModalContent/style.ts` definition.~~
    - ~~I'm not super crazy about this setup, but also wasn't too happy with adding `ImmersiveModalContent` as a sub-component within the `immersiveModal/index.tsx` file 🤷~~
    - ~~If the diff is difficult to understand, I can make all the changes within the same `immersiveModal/index.tsx` file.~~
    - I've decided to keep the logic in the main ImmersiveModal component for ease of use--moving things into a sub-component blew up the diff, and seemed more an application of a dogmatic principle (one component per file) than useful. 
    
Review App for testing: https://curology-radiance-pr-1159.herokuapp.com/?path=/story/components-immersivemodal--with-footer-buttons

### What & Why

Our `DialogModal` has preferable focus behavior where, when initiating the modal with a keyboard, we, with the help of `@react-aria/focus`, contain focus within the modal, automatically focus on the first focusable element in the modal, and restore focus to the previously focus element when it closes.

https://user-images.githubusercontent.com/13544620/133697285-13e4673d-99ac-425d-bb59-8fc752747f6d.mov

I attempted to add the same functionality to the ImmersiveModal, but there were some issues so I pulled out the handling in #517.

I'm currently auditing accessibility in our app and noticed that there's sub-optimal behavior when using `ImmersiveModal` because it lacks the `autoFocus` functionality. As a result it requires tabbing multiple times until we focus into the modal content, at which point the `contain` functionality takes effect, but if there are a large number of focusable elements between the modal initiating element and the actual modal content, it may be very confusing behavior. 

So this PR re-attempts what I tried to add prior to #517, learning from the bespoke focus handling there, and doing more robust QA:

https://user-images.githubusercontent.com/13544620/133701480-72a11742-3607-42b9-b672-d02a9cf30646.mov

There are no *real* style regressions with these changes: Chromatic failed because there is now a focus outline on the `CrossIconButton`. (See build here: https://www.chromatic.com/build?appId=5f6e46f1e7081000221523cf&number=1214). This behavior is dependent on `focus-visible`, though, meaning desktop and normal mobile touch users will not see this outline, only keyboard uses who trigger the modal with the keyboard will. 

The biggest style change is that the `CrossIconButton` is now fixed, and I think it looks better.

Before:

https://user-images.githubusercontent.com/13544620/133697547-0fe04e94-b0e3-419b-b9f2-c1d314904b4a.mov

After:

https://user-images.githubusercontent.com/13544620/133697557-0de20336-4f2b-437d-93ad-1bf364db037d.mov

This is also the result of the most significant markup change that isn't just moving the existing UI components around: this component currently renders up to **three** of the same button: the mobile button, the desktop header button, and then the button we show when the desktop header is `display: none`. Now there are two buttons in the markup, which seems more justifiable given the different desktop/mobile handling. 

These multiple buttons were also why I was seeing the desktop click behavior in #517: we were maintaining focus on a button that was no longer visible, but that stayed at the top of the modal with scrollable content, so when we click the break focus the content scrolls to that element that had focus.

Ultimately, after wrestling with this component for some time, I'd for us to consider **using it very sparingly**. Adding this behavior already took a lot of time, there isn't an easy `@react-aria` hook (like `useDialogModal`) that we can eventually migrate to for greater robustness, and the implementation is unorthodox (we have to make the content container focusable, since scroll containers are not automatically focusable: https://www.chromestatus.com/feature/5231964663578624).


src/shared-components/immersiveModal/style.ts
```diff
@@ -37,7 +36,7 @@ const Overlay = styled.div`
   }
 `;
 
-const CrossIconContainer = styled.button`
+const CrossIconButton = styled.button<{ showDesktopHeaderBar?: boolean }>`
```
michaeljaltamirano, line 39: Renamed to reflect that this is, in fact, a button. `DialogModal` still uses a `div` and so I kept the name container there.

src/shared-components/immersiveModal/index.tsx
```diff
@@ -180,49 +179,17 @@ export const ImmersiveModal: ImmersiveModal = ({
...
+                    footerContent={footerContent}
+                    headerImage={headerImage}
+                    showDesktopHeaderBar={showDesktopHeaderBar}
+                    showMobileHeaderBar={showMobileHeaderBar}
+                    title={title}
+                    handleCloseIntent={handleCloseIntent}
                   >
-                    <Style.CrossIconContainer
-                      onClick={handleCloseIntent}
-                      tabIndex={showDesktopHeaderBar ? -1 : 0}
```
michaeljaltamirano, line 190: Because there is only one `CrossIconButton` now (formerly named `CrossIconContainer`) we can remove the `tabIndex` handling.

src/shared-components/immersiveModal/immersiveModalContent/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+  title,
+  handleCloseIntent,
+  children,
+}: ImmersiveModalContentProps) => {
+  const focusManager = useFocusManager();
+
+  /**
+   * It is not typical modal behavior to be able to scroll. Consequently, our
+   * keyboard autofocus handling means when users try and scroll in the ImmersiveModal,
+   * nothing happens. We making the content of the container tabbable such that when the
```
ipc103, line 42:
```suggestion
   * nothing happens. We make the content of the container tabbable such that when the
```

ipc103 (approved):
👍 👍   LGTM - functionality works well. Nice work figuring this out. I agree that it would be nice to figure out alternatives to this component where possible.

src/shared-components/immersiveModal/style.ts
```diff
@@ -56,8 +63,13 @@ const CrossIconContainer = styled.button`
   cursor: pointer;
 
   ${MEDIA_QUERIES.mdUp} {
-    top: 16px;
+    top: ${({ showDesktopHeaderBar }) =>
+      showDesktopHeaderBar === true ? SPACER.x5large : '4.5rem'};
+    transition: top ${ANIMATION.defaultTiming}
```
michaeljaltamirano, line 68: Since we're now using this as one button for desktop, I added a transition to make the positioning look sleeker with scrollable content:

https://user-images.githubusercontent.com/13544620/134046857-0ec2a925-feb7-419b-900d-4def6c421983.mov

The timing matches our other `opacity` transitions since they all rely on `ANIMATION.defaultTiming`.

PR #1175, opened 2021-09-22
Add cleanse, hydrate, and protect glyphs [PX-51]

jashieh:
[Figma](https://www.figma.com/file/YV994I2jchdrCpRmwpGIX0/New-Icons-for-Radiance?node-id=1%3A558)
Adds the following glyphs:

Screen Shot 2021-09-22 at 4 02 56 PM


daigof (commented):
Michelle needs to add it to the Radiance Figma file first. and think on a counterpart for Agency to put it in that Figma file as well. 
Radiance is much more structured as a design system than before we cannot just add freely without design consent. We should take the Name from the Radiance Figma file.
If no Agency icon is planned we should pass it as null in your second arguments.

Radiance Figma: https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance-UI?node-id=2300%3A18242&viewport=476%2C519%2C0.8739789724349976

Agency Figma: https://www.figma.com/file/FjOFrQaYVSdlaMIhpgR192/Ruth-UI?node-id=2300%3A9919&viewport=480%2C520%2C0.8739789724349976

nic-git:
I re-requested review by accident, but if the Figma work is still required we can raise in our Alignment meeting

daigof:
@nic-git yes Nic we need help coordinating this push, The glyphs need to be added here
https://www.figma.com/file/RZIO3wuO5Uk7sPZoXstAuXIA/Radiance-UI?node-id=2300%3A9919&viewport=476%2C519%2C0.8739789724349976

We take from there the **name** of the Glyph as well, we cannot just make up the names. 
For example why this is important: there is a Sunblock glyph already, 
Alternatevely we can decide to add these icons in the Routine project folder and deal with Radiance later.

michaeljaltamirano:
> other icons are passing the same SVG for both themes as well so I guess this can go as is

@daigof Let's just make sure that when we update the Radiance toolkit to update the Ruth icons with these same glyphs.

daigof:
Yeah so yeah you right, we will tackle these during cooldown, there is pending conversation about the Tab component as well

PR #1177, opened 2021-09-24
[SegmentedControl] Refinements

michaeljaltamirano:
### CHANGELOG

1. Run `eslint --fix` to address new standards since branch was created.
1. Replace `80px` `border-radius` with `theme.BORDER_RADIUS.large`.
    - It is accurately `0` for the secondary theme, and it is `32px` for the primary theme, which is equal to `80px` (you can edit in your dev tools to examine this.)
1. Update `transition: transform 0.22s` to `transition: transform 0.3s` to match `300ms` specified in Figma spec: Screen Shot 2021-09-24 at 12 24 01 AM
1. Move `border` style from item container to individual segment items to prevent issue of `Indicator` being presentationally different from the other segment items.
    - Before: Screen Shot 2021-09-24 at 12 29 39 AM
         - The height of the `position: absolute` element was larger because the `border` on the parent was not affecting the height. 
    - After: Screen Shot 2021-09-24 at 12 29 46 AM
    
    **TODO:** There' still some refinement that could be made here: the absolutely positioned indicator is 172px width compared to 168px width for the non-indicator buttons, so they should be made more consistent, probably. 
1. Update the width handling--by removing the horizontal padding from `padding: 0.25rem`--such that the `position: absolute` Indicator element width matches the other elements. 
    - Before: 
    
    https://user-images.githubusercontent.com/13544620/134618026-81e8636e-56aa-4f63-b4cd-1de188d9fd1c.mov
    - After: 
    
    https://user-images.githubusercontent.com/13544620/134618040-3ecd3bb8-f0bf-4b4a-a1ce-47c14e694df8.mov
1. Add visual regression tests for un-selected focus state. 

### TODO for future PRs

The current functionality sits somewhere between a [RadioGroup](https://react-spectrum.adobe.com/react-aria/useRadioGroup.html) and a [TabList](https://react-spectrum.adobe.com/react-aria/useTabList.html). In both examples the keyboard controls use tabbing to focus the options, and then arrow keys to toggle between the options, where changing the selection is done automatically, without a button "click" event. 

If we decide to use neither of these paradigms, the biggest problem right now is that because the Indicator comes before the more appropriately interactive button elements, toggling through the options sometimes paradoxically goes backwards in order. This video shows normal tabbing behavior: the first selection is the indicator, and then it starts from the first available button, and then skips the button that matches the indicator selection and proceeds to the next focusable element:

https://user-images.githubusercontent.com/13544620/135016279-dce30f20-ad01-4d1a-8a08-74d7811647e7.mov

This needs to be fixed in a future PR, but for now we are maintaining this behavior to unblock UI development this week.

### DIFF
--- a/src/shared-components/segmentedControl/__snapshots__/test.tsx.snap
+++ b/src/shared-components/segmentedControl/__snapshots__/test.tsx.snap
@@ -55,7 +55,6 @@ exports[`<SegmentedControl /> renders a regular segmented control 1`] = `
   -moz-transform: translate3d(0px, 0, 0);
   -ms-transform: translate3d(0px, 0, 0);
   transform: translate3d(0px, 0, 0);
-  box-shadow: 0px 2px 4px rgba(51, 46, 84, 0.05);
   border: 4px solid #EDEDF0;
 }
 
@@ -83,7 +82,7 @@ exports[`<SegmentedControl /> renders a regular segmented control 1`] = `
   top: 0;
   left: 0;
   bottom: 0;
-  border: 4px solid #EDEDF0;
+  padding: 0.25rem;
   color: #706D87;
   font-size: 0.875rem;
   line-height: 1.71;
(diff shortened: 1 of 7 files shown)


winstonkim (approved):
oh didn't even realize the component looked the same with `borderRadius.large`. good catch! 😅 

looks good - thanks for refining this!

PR #1252, opened 2021-11-05
Add .sr-only class for screen reader text

joeyquarters:
Adds an `.sr-only` class to the global brand styles, taking inspiration from [Bootstrap's screen reader utility](https://getbootstrap.com/docs/4.5/utilities/screen-readers/).

If there is a better way to expose this styling globally via an emotion `css` function, I'd also be very open to that implementation. I suppose we could expose it as a styled component, but that seems much less ergonomic for the purposes of using these styles on various elements.

### DIFF
--- a/src/shared-components/globalStyles/style.ts
+++ b/src/shared-components/globalStyles/style.ts
@@ -265,4 +265,16 @@ export const brandStyles = (theme: ThemeType) => `
   .no-scroll {
     overflow: hidden;
   }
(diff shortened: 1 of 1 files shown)


joeyquarters:
Closing in favor of pulling in [the `@react-aria/visually-hidden` library](https://react-spectrum.adobe.com/react-aria/VisuallyHidden.html)

PR #1308, opened 2021-12-13
Update default input color

jashieh:
The color on inputs were showing up as blue on IOS 15 devices because Apple introduced a default color to all inputs. This sets our own default color so the inputs don't fallback to the IOS default.

### DIFF
--- a/src/shared-components/globalStyles/style.ts
+++ b/src/shared-components/globalStyles/style.ts
@@ -221,7 +221,7 @@ export const brandStyles = (theme: ThemeType) => `
     transition-timing-function: ease-in-out;
   }
 
-  input {
+  input:is([type="button"], [type="submit"], [type="reset"]), input[type="file"]::file-selector-button, button {
     color: ${theme.COLORS.primary};
   }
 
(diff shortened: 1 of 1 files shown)


src/shared-components/globalStyles/style.ts
```diff
@@ -221,6 +221,10 @@ export const brandStyles = (theme: ThemeType) => `
     transition-timing-function: ease-in-out;
   }
 
+  input {
```
michaeljaltamirano, line 224: I think we want to make this selector identical to the one that is causing an issue for iOS 15 Safari: 

```css
input:is([type="button"], [type="submit"], [type="reset"]), input[type="file"]::file-selector-button, button {
    [declarations]
}
```

The `button` part should cover most of our use-cases, but I think to be safe we should copy the entire selector.

michaeljaltamirano (approved):
LGTM after code comment change!

michaeljaltamirano:
Hm alright let me take a look at Chromatic

michaeljaltamirano:
> Hm alright let me take a look at Chromatic

Thanks for approving the changes in Chromatic! For posterity in case anyone else looks at this PR in the future, the selector ended up changing the presentation of a button that had been relying on browser default behavior before. We've been relying on it being `black` for both themes. However, the primary value is what we want, since that is what is in Figma:

Screen Shot 2021-12-14 at 10 27 09 AM

Before making this fix for iOS 15 it looks like this:

Screen Shot 2021-12-14 at 10 16 54 AM

Thanks for the speedy fix, Justin!

PR #1330, opened 2021-12-28
[ESLint] Add Hooks Rules

michaeljaltamirano:
I wasn't aware we haven't been including these 🥸 

WIP

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,12 +4,13 @@
     "ecmaVersion": 2018,
     "sourceType": "module",
     "project": "./.eslint.tsconfig.json",
-    "allowAutomaticSingleRunInference": false,
+    "allowAutomaticSingleRunInference": false
   },
   "extends": [
     "plugin:@typescript-eslint/recommended-requiring-type-checking",
     "plugin:jest/all",
     "plugin:sonarjs/recommended",
+    "plugin:react-hooks/recommended",
     "curology"
   ],
   "plugins": ["@emotion"],
@@ -143,6 +144,8 @@
     ],
     "react/state-in-constructor": "off",
     "react/static-property-placement": "off",
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "error",
     "sonarjs/no-nested-template-literals": "off"
   },
   "overrides": [
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -65,7 +65,7 @@ export const Accordion: Accordion = ({
     if (contentHeight !== nextHeight) {
       setContentHeight(nextHeight);
     }
-  });
+  }, [isOpen, contentHeight]);
 
   const handleKeyDown = (event: React.KeyboardEvent): void => {
     if (!disabled && event.key === keyboardKeys.enter) {


src/shared-components/alert/index.tsx
```diff
@@ -24,7 +24,7 @@ export interface AlertProps {
   content: React.ReactNode;
   ctaContent?: React.ReactNode;
   duration?: string | number;
-  onExit?: ((rest: Omit<AlertProps, 'onExit'>) => void) | (() => void);
+  onExit?: () => void;
```
michaeljaltamirano, line 27: Breaking change, will release as a new major: this `onExit` usage with the rest of the props is no longer used and does not really make sense either (since the component usage has access to all of the props). Including all `alertProps` in the effect was a dependency nightmare so this cleans that up.

All consumer usage should also pass in `onExit` with referential integrity but since the Alert is not given to being re-rendered it is not a big deal if it is not.

src/shared-components/immersiveModal/index.tsx
```diff
@@ -211,6 +198,19 @@ export const ImmersiveModal: ImmersiveModal = ({
     const modalDesktopScrollingElementInstance =
       modalDesktopScrollingElement.current;
 
+    const handleScroll = throttle(() => {
```
michaeljaltamirano, line 201: Moving this inside the effect removes the need to specify `handleScroll` as a dependency.

src/shared-components/segmentedControl/index.tsx
```diff
@@ -58,6 +54,10 @@ export const SegmentedControl: React.FC<SegmentedControlProps> = ({
     }
   };
 
+  if (itemsCount === 0) {
```
michaeljaltamirano, line 57: All hooks must run in the same order, and in their entirety. Early returns violate this rule.

src/shared-components/segmentedControl/index.tsx
```diff
@@ -42,11 +38,11 @@ export const SegmentedControl: React.FC<SegmentedControlProps> = ({
       setTargetWidth(targetRef.current.offsetWidth);
       setTransform(
         `translate3d(${
-          targetRef.current.offsetWidth * activeSegmentIndex
+          targetRef.current.offsetWidth * initialActiveIndex
         }px, 0, 0)`,
       );
     }
-  }, [targetRef]);
+  }, [initialActiveIndex]);
```
michaeljaltamirano, line 45: `initialActiveIndex` captures that this is an effect that runs `onMount`, all other changes after rendering are handled by `onSegmentClick`, not with side-effects.

src/tests/testingLibraryHelpers.tsx
```diff
@@ -16,7 +16,7 @@ interface RenderOptions extends ReactTestingLibrary.RenderOptions {
  * container in their unit test** will have test failures on CI due to
  * serialization order being off.
  */
-const usePortalContainer = () => {
+const addPortalContainer = () => {
```
michaeljaltamirano, line 19: `use*` syntax is reserved for the linter and so should be avoided when not really a hook

src/shared-components/tabs/index.tsx
```diff
@@ -8,7 +8,6 @@ export interface TabType {
   text: string;
 }
 
-// Hardcoding TabType below to show the actual type in the docs
```
michaeljaltamirano, line 11: Use of Generic made this no longer apply

src/shared-components/alert/index.tsx
```diff
@@ -59,30 +59,33 @@ export const Alert: Alert = (alertProps) => {
...
-    const { onExit: _onExit, ...otherProps } = alertProps;
+    if (isDefined(timer.current)) {
+      window.clearTimeout(timer.current);
+    }
 
     window.setTimeout(() => {
       setExited(true);
-      onExit(otherProps);
+
+      if (onExit) {
```
michaeljaltamirano, line 84: No longer giving it a default no-op function parameter so we check if it's provided first.

src/shared-components/alert/index.tsx
```diff
@@ -59,30 +59,33 @@ export const Alert: Alert = (alertProps) => {
...
-  let timer: number | undefined;
-
-  const alertExitHandler = () => {
+  const alertExitHandler = React.useCallback(() => {
     setExiting(true);
-    window.clearTimeout(timer);
 
-    const { onExit: _onExit, ...otherProps } = alertProps;
+    if (isDefined(timer.current)) {
+      window.clearTimeout(timer.current);
```
michaeljaltamirano, line 78: There's no bugs from calling `window.clearTimeout(undefined)` but this just makes the `number | undefined` relationship more explicit. Using a ref also simplifies our dependency array for `useCallback` since we would otherwise have to add it.

src/shared-components/alert/index.tsx
```diff
@@ -92,17 +95,17 @@ export const Alert: Alert = (alertProps) => {
       return;
     }
 
-    timer = window.setTimeout(
```
michaeljaltamirano, line 95: `Assignments to the 'timer' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect.`

PR #1373, opened 2022-02-01
Set radiance-ui compat to >14.x to resolve compat issues

michaeljaltamirano:
TODO

https://github.com/curology/radiance-ui/pull/1283
https://github.com/curology/radiance-ui/pull/1313


michaeljaltamirano:
Heroku builds with the most recent compatible version of node and I do not see a way to set the review apps to a specific version

PR #1374, opened 2022-02-02
Add onFocus prop to Dropdown

joeyquarters:
This PR exposes the `onFocus` event from the underlying elements of the `Dropdown` component.

**Use case**
We would like to instrument an analytics event when a user focuses on the dropdown to observe interaction behavior. We can derive certain observations like "Users who at least focused the dropdown in this form were 50% less likely to bounce" or "30% of users focused on this dropdown, but only 50% of those who focused ultimately made a selection".

### DIFF
--- a/src/shared-components/dropdown/index.tsx
+++ b/src/shared-components/dropdown/index.tsx
@@ -34,6 +34,10 @@ interface DropdownProps<T> {
    * The handler to be invoked on option change
    */
   onChange: (option: T) => void;
+  /**
+   * The handler to be invoked on focus
+   */
+  onFocus?: (event?: React.FocusEvent) => void;
   options: T[];
   /**
    * Specifies maximum height of the expanded dropdown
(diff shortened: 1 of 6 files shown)


michaeljaltamirano (approved):
Code LGTM. I added a commit because I realized that we weren't type-checking the stories in CI, and that the required prop for `<MobileDropdown />` direct usage wasn't being caught in our workflow. Sorry for stepping on the commit toes.

src/shared-components/dropdown/index.tsx
```diff
@@ -34,6 +34,10 @@ interface DropdownProps<T> {
    * The handler to be invoked on option change
    */
   onChange: (option: T) => void;
+  /**
+   * The handler to be invoked on focus
+   */
+  onFocus?: (event?: React.FocusEvent) => void;
```
michaeljaltamirano, line 40: Would it be worthwhile to rename this `onDropdownContainerFocus`? I might expect `onFocus` to cover focusing the individual dropdown element, but this is specifically for the container, so it might be helpful to distinguish it.

PR #1465, opened 2022-03-31
Bump emotion packages

michaeljaltamirano:
Testing out why #1422 is brittle in Chromatic


michaeljaltamirano:
Looks like it's because of fonts, see #1470

PR #1470, opened 2022-04-05
[Storybook] Update Config

michaeljaltamirano:
There've been some minor version changes we haven't incorporated, like the deprecated `--static-dir` CLI usage, so this PR goes through the [MIGRATION file](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md) and updates some usages.


.storybook/main.ts
```diff
@@ -59,31 +59,26 @@ const config = {
       );
     }
 
-    /**
-     * Until Storybook migrates its own internal @emotion usage from
-     * v10 to v11, this allows us to maintain compatibility
-     */
-    const emotion11CompatibleConfig = {
```
michaeljaltamirano, line 66: 6.4 added the `emotionAlias` helper so we do not need to manually set this anymore: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#emotion11-quasi-compatibility

public/stylesheet.css
```diff
@@ -0,0 +1,40 @@
+/* Includes just woff2 for Chromatic testing */
```
michaeljaltamirano, line 1: Chromatic has recently gotten brittle, e.g. roughly 5 specs fail visual regression testing because the fonts are not loading before the Snapshots. The Chromatic GUI suggested moving these external assets into Storybook for better performance.

Here's an example of a failure: https://www.chromatic.com/build?appId=5f6e46f1e7081000221523cf&number=1700

And here's an example of the messaging: https://www.chromatic.com/test?appId=5f6e46f1e7081000221523cf&id=624b9c28d0d7ba003a869676

Screen Shot 2022-04-04 at 8 18 33 PM

PR #1517, opened 2022-04-19
Add Chip info color

winstonkim:
Adds info color for `Chip`. This new color is used in the site product pages + nav.

### DIFF
--- a/src/shared-components/chip/index.tsx
+++ b/src/shared-components/chip/index.tsx
@@ -5,7 +5,7 @@ import { useTheme } from '@emotion/react';
 import Style from './style';
 import type { ThemeColors, ThemeType } from '../../constants';
 
-export type StatusType = 'primary' | 'success' | 'error' | 'white';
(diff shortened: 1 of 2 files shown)


michaeljaltamirano (commented):
I went ahead and approved the first Curology visual change in Chromatic 👍: https://www.chromatic.com/test?appId=5f6e46f1e7081000221523cf&id=625ef3d1db44c1003a0f7956

I am wondering if these styles are specifically for Curology vs. Agency, or both? The Agency `info` chip matches the default for non-low contrast, and the low-contrast version somewhat clashes against the default Agency background: https://www.chromatic.com/test?appId=5f6e46f1e7081000221523cf&id=625ef3d1db44c1003a0f7957

I think before merging in these changes, though, we'd want the `info` chip variations documented in Radiance. Would the designer you're working with be able to do that (through the normal Design steps) before merging this in?

winstonkim:
@michaeljaltamirano thanks, i've connected with @sierrabvk about these questions and will let you know when there's an update!

PR #1621, opened 2022-06-16
[a11y] Add screen reader indicator for invalid input

azimmerly:
### What & Why
The `Field` component does not currently indicate a failure/invalid state to screen-readers. This PR addresses this issue by adding an `output` element to `VerificationMessages` and adding an [aria-invalid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-invalid) state to the input field.

---
Screen Shot 2022-06-16 at 9 54 35 AM

https://user-images.githubusercontent.com/83731749/174132070-79126982-ce24-47ca-a0fc-a7e15a979f9f.mov


K4orta (approved):
LGTM

PR #1667, opened 2022-07-26
Add Logo to Theme Constants

iiimosley:
A (gross) solution for being able to making the brand icon available through direct through Radiance.


ZeMunchkin (commented):
@iiimosley It looks like the other PR with the logos in PocketDerm was merged. Wasn't sure if that was a temporary workaround or if you still needed this PR to be reviewed?

PR #1706, opened 2022-08-25
[og-387] [og-388] update radiance modal components to use react-aria

irislam94:
- Adds @react-aria/dialog package in order to use `useDialog` hook so that screen readers can announce that the user is in a modal
- Adds `useDialog` hook to both `ImmersiveModal` and `DialogModal`
- Renames `onCloseIconClick` to `onClick` in `DialogModal` and makes this property required
- Updates tests
- Updates the `DialogModal` in storybook (since props changed)
  - Removes duplicate documentation now that close icon is default

### DIFF
--- a/stories/dialogModal/index.stories.tsx
+++ b/stories/dialogModal/index.stories.tsx
@@ -31,7 +31,12 @@ export const Default = () => {
       </Button>
 
       {openModal && (
-        <DialogModal title="Heads up!">
+        <DialogModal
+          title="Heads up!"
+          onClose={() => {
+            setOpenModal(false);
+          }}
+        >
           <DialogModal.Paragraph>
             This will remove the cleanser and moisturizer from your free trial,
             too. Just the custom bottle will be sent your way!
@@ -80,7 +85,12 @@ export const DefaultOpened = () => {
       </Button>
 
       {openModal && (
-        <DialogModal title="Heads up!">
+        <DialogModal
+          title="Heads up!"
+          onClose={() => {
+            setOpenModal(false);
+          }}
+        >
           <DialogModal.Paragraph>
             This will remove the cleanser and moisturizer from your free trial,
             too. Just the custom bottle will be sent your way!
@@ -128,6 +138,9 @@ export const WithColor = () => {
         <DialogModal
           title="Heads up!"
           backgroundColor={theme.COLORS.background}
+          onClose={() => {
+            setWithColor(false);
+          }}
         >
           <DialogModal.Paragraph>
             This will remove the cleanser and moisturizer from your free trial,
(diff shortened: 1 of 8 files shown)


stories/dialogModal/index.stories.tsx
```diff
@@ -128,6 +138,9 @@ export const WithColor = () => {
         <DialogModal
           title="Heads up!"
           backgroundColor={theme.COLORS.background}
+          onClose={() => {
+            setWithColor(false);
```
azimmerly, line 142: [nit] Not a big deal and I know you didn't add this, but I feel like `setWithColor` is a slightly confusing name. Maybe something like `setOpenModalWithColor`?

azimmerly (approved):
LGTM!

azimmerly:
Also, curious why `DialogModal` uses `role: 'alertdialog'` while `ImmersiveModal` uses `role: 'dialog'`.

stories/dialogModal/index.stories.tsx
irislam94: Good callout! I'll make the change

irislam94:
> Also, curious why `DialogModal` uses `role: 'alertdialog'` while `ImmersiveModal` uses `role: 'dialog'`.

Here's the documentation for [alertdialog](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alertdialog_role) and [dialog](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role). Based off of our use case for each, it seems that `DialogModal` is used for more urgent information that may require user action (e.g. `CancelTrialConfirmModal`, `ConfirmRemoveRecommendedItemModal`) while `ImmersiveModal` is more informative (e.g. `ReferralProgramReferredModal`, `UploadPhotoTipsModal`)

jeff-dombroski (approved):
💯 Nice! Appreciate the consideration of difference between the two modals, not sure I would've caught that 🙏

irislam94:
> 💯 Nice! Appreciate the consideration of difference between the two modals, not sure I would've caught that 🙏

It was all Elysia! Not sure I would've caught that either 🙈

CurologyJonathanHartley:
@irislam94, is there a reason you haven't merged this in yet? Just curious as I was about to create a `radiance-ui` PR but didn't know if there was something I should know prior to doing so. Thanks.

irislam94:
> @irislam94, is there a reason you haven't merged this in yet? Just curious as I was about to create a `radiance-ui` PR but didn't know if there was something I should know prior to doing so. Thanks.

I think it was just Friday and I was out all of last week for the offsite/pto and I didn't want to update radiance without updating pocketderm shortly after. I'm planning on merging this in today!

PR #1740, opened 2022-09-21
GE-488: Fix select element accessibility

serickson-curology:
The primary purpose of this PR is to allow for keyboard navigation of select elements using our shared UI component libraries. To accomplish this, I replaced our javascript implementation of the select for desktop with a slightly adjusted implementation of the mobile component, which has been using a regular select html element.


src/shared-components/dropdown/index.tsx
```diff
@@ -125,35 +77,16 @@ export const Dropdown = <T extends OptionType>({
...
-      />
-    );
-  }
-
-  const currentOption = options.find((option) => option.value === value);
-
+  // if (touchSupported) {}
   return (
-    <DesktopDropdown
+    <GenericConfigurableDropdown
```
joeyquarters, line 82: Since we've done away with the mobile and desktop components, do you think it might cause less indirection to move the logic contained in `GenericConfigurableDropdown` into this `Dropdown` component?

joeyquarters (commented):
Left some questions as a first-pass look! Thanks for tackling this 💪

stories/dropdown/index.stories.tsx
```diff
@@ -27,7 +27,7 @@ const options = [
   { value: 1, label: 'First option' },
   { value: 2, label: 'Second option' },
   { value: '3', label: 'Third option' },
-  { value: '4', label: 'Fourth option' },
```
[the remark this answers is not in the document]
serickson-curology: This was me being sloppy at cleaning up. I made the list much much longer for showing to Adam so I could give a better example view of what the list would look like with 50 options and then forgot that we have 5 initially rather than four. I'll restore the fourth option.

src/shared-components/dropdown/index.tsx
```diff
@@ -125,35 +77,16 @@ export const Dropdown = <T extends OptionType>({
...
-        options={options}
-        textAlign={textAlign}
-        value={value}
-      />
-    );
-  }
-
-  const currentOption = options.find((option) => option.value === value);
-
+  // if (touchSupported) {}
```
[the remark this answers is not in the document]
serickson-curology: I definitely agree. Thanks for catching that miss on my part

src/shared-components/dropdown/index.tsx
serickson-curology: I think this is a good suggestion. I was a bit worried about condensing it all down into that top-level component in one go because I am still pretty new and having two modules felt like it would allow for cleaner context-based configuration in the future, but this pattern doesn't really buy us anything right now

pmp-curology (approved):
🙌

pmp-curology (approved):

src/shared-components/dropdown/index.tsx
```diff
@@ -39,10 +38,6 @@ interface DropdownProps<T> {
    */
   onDropdownContainerFocus?: (event?: React.FocusEvent) => void;
   options: T[];
-  /**
-   * Specifies maximum height of the expanded dropdown
-   */
-  optionsContainerMaxHeight?: string;
```
meron-yemane, line 45: @serickson-curology Now that we've removed this prop, there are [some components ](https://github.com/curology/PocketDerm/blob/79065978c1ca8899a454c32efb62ab140256cd9c/resources/assets/jsx/components/formComponents/formField/select.tsx#L55) in Pockederm that are erroring out because they are still passing it.  Is it safe to remove?

PR #1745, opened 2022-09-26
meron/add border and update border-radius of OptionButton

meron-yemane:
Discussed with Adam and we decided it best to add a border back to the `OptionButton` component for consistency reasons and Agency compatibility. Adding the border back in revealed `border-radius` discrepancies between `ImageContainer` and 'Image` so adjustments there are also added as part of this pr (before and after shown below). 
Screen Shot 2022-09-26 at 12 59 17 PM
Screen Shot 2022-09-26 at 12 59 46 PM


LilCare (approved):
Looks good! Glad you found a solution that works!

MonsterEyes (approved):
Nice solution

PR #1748, opened 2022-09-27
Adding optional ariaLabel prop for Dropdown to allow for accessible use without a standalone label

serickson-curology:


meron-yemane (approved):
🚀

PR #1764, opened 2022-10-11
Add support for additional attribute passing to Dropdown component

serickson-curology:
Updating the dropdown component to allow for more flexible use in the consuming library. Basically I am removing the explicit option for `aria-label` and replacing it with support for optional `...rest` attributes. Had to update the docs and snapshot to go along with these changes.

### DIFF
--- a/src/shared-components/dropdown/index.tsx
+++ b/src/shared-components/dropdown/index.tsx
@@ -24,10 +24,6 @@ export interface OptionType {
 }
 
 interface DropdownProps<T> {
-  /**
-   * An optional prop to allow for the select to have an aria-label attribute added for applications where a standalone label isn't ideal.
-   */
-  ariaLabel?: string | undefined;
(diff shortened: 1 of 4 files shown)


src/shared-components/dropdown/index.tsx
```diff
@@ -61,7 +61,7 @@ export const Dropdown = <T extends OptionType>({
   options,
   textAlign = 'left',
   value,
-  ariaLabel,
+  ...rest
```
LilCare, line 64: Just curious: Are we fine with any attribute? Or do we want to hold onto a list of accepted values? We could then update to validate `...rest` and reject any unknown attributes.

LilCare (commented):
I'm definitely not an expert on component libraries, but it seems to me that we would want it to have a discrete number of possible attributes and values. Just curious: do you know what the best practice is here?

PR #1804, opened 2022-11-30
Add support for React 18

trappar:
This PR makes the following changes:
* Adds support for React 18 by expanding the peer-dependencies on `react` and `react-dom`.
* Removes `react` and `react-dom` as direct dependencies and instead declares them as dev-dependencies. This allows consuming applications to fully control the version of react.
* Updates Storybook to 6.5, which adds basic support for React 18.
* Updates `@react-aria/dialog` and `@react-aria/focus` to versions which support React 18.
* Fixes a minor issue with the jest module resolution config that started occurring due to some package version bump.
* Removes PropTypes, as they are not compatible with React 18.

These changes provide a path forward for consumers of `radiance-ui` who want to switch to React 18 while maintaining the same support for React 17 which already existed.


azimmerly:
Thanks for updating this. I wonder if its work consuming this package locally [using yalc](https://www.notion.so/curology/Radiance-UI-3e2736219ba14a91b14075a3d9cd922e#2b46407e25b6473f9f16be521c560b02) in both the site and app to ensure everything remains ok.

CurologyJonathanHartley:
> Thanks for updating this. Might be worth testing out consuming this package locally [using yalc](https://www.notion.so/curology/Radiance-UI-3e2736219ba14a91b14075a3d9cd922e#2b46407e25b6473f9f16be521c560b02) in both the site and app to ensure everything remains ok.

Agreed. Based on things myself and other devs have run into, yalc is the only reliable way we've found to dev radiance-ui locally.

trappar:
Tested using yalc in both curology-gatsby and PocketDerm repos (without changing anything else like upgrading React), and it worked in both without issue.

trappar:
I've searched for any usage of the removed exports (such as `PRIMARY_COLORS_PROP_TYPES`) in the Curology org and found that there are none. These changes seem very unlikely to be breaking, but I will still do a major version bump when I publish a new version just to be safe.

pmp-curology (approved):
Thanks for taking care of this! Great work.

azimmerly (approved):
Looks good!

MonsterEyes (approved):
React 18 🚀

PR #1830, opened 2022-12-30
GE-746: Adding info styling for chip

serickson-curology:
GE-746 required a new style of chip. Style info was taken from Figma and should match current styles

### DIFF
--- a/src/shared-components/chip/index.tsx
+++ b/src/shared-components/chip/index.tsx
@@ -4,7 +4,7 @@ import { useTheme } from '@emotion/react';
 import Style from './style';
 import type { ThemeColors, ThemeType } from '../../constants';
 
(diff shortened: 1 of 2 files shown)


src/shared-components/chip/index.tsx
```diff
@@ -46,6 +46,16 @@ const getStyles = ({ isLowContrast, status, theme }: GetStylesProps) => {
     }
   }
 
+  if (status === 'info') {
```
LilCare, line 49: [nit] For consistency with status = PRIMARY with low contrast, can we separate this into two separate conditionals? As a wider engineering org, we tend to avoid nested conditionals wherever possible.

LilCare (approved):
Looks good! I have one small request.

PR #1847, opened 2023-01-23
add spironolactone glyph

bettyleung:
https://linear.app/curology/issue/RE-2028/add-topical-spiro-glyph-to-radiance-storybook

Screen Shot 2023-01-22 at 8 56 16 PM


MonsterEyes (approved):
LGTM!!
Optionally you may want to have another engineer that has done this work before review it.

PR #1875, opened 2023-07-18
[money-630] Replace larssiet font with skintype-text

cherimarie:
Related: https://github.com/curology/curology-gatsby/pull/2432

Question: 
I'm surprised that `.yarn/` and `yarn.lock` aren't git ignored. Is that intentional? Should I be committing `.yarn/`?

Local Storybook, showing all font styles default to 'skintype-test' instead of 'larssiet'
Screen Shot 2023-07-17 at 6 55 34 PM

Related PRs:
 - https://github.com/curology/PocketDerm/pull/21008
 - https://github.com/curology/curology-gatsby/pull/2432


public/stylesheet.css
```diff
@@ -7,14 +7,16 @@
...
-  font-family: 'larssiet';
-  src: url('/fonts/34535B_1_0.woff2') format('woff2');
+  font-family: 'skintype-text';
+  src: url('/fonts/skintype-text/SkintypeText-Regular.woff2') format('woff2'),
+  url('/fonts/skintype-text/SkintypeText-Regular.otf') format('otf');
 }
 
 @font-face {
-  font-family: 'larssiet';
+  font-family: 'skintype-text-bold';
```
cherimarie, line 16: Is this right? Should it just be 'skintype-text'?

src/constants/fonts/primary.ts
```diff
@@ -2,7 +2,7 @@ const PRIMARY_FONTS = {
   headerFont:
     "'nocturno', Georgia, TimesNewRoman, 'Times New Roman', Times, Baskerville, serif",
   baseFont:
-    "'larssiet', 'Avant Garde', Avantgarde, 'Century Gothic', CenturyGothic, AppleGothic, Verdana, sans-serif",
+    "'skintype-text', 'Avant Garde', Avantgarde, 'Century Gothic', CenturyGothic, AppleGothic, Verdana, sans-serif",
```
cherimarie, line 5: Unrelated to this PR, Avant Garde/avantgarde font doesn't seem to be loading. If I inspect source on the live site right now and remove larssiet from the stack, I get Century Gothic. This could be an environment-specific issue, but there's also no other reference to Avant Garde here, including any import. Should we do something about that?

ZeMunchkin (approved):
Changes make sense to me. There's some CI failures, but assuming those get worked out I don't see anything blocking.

When we tried adding a new font for admin last year, I remember also needing to add it to PocketDerm font files for them to take effect: https://github.com/curology/PocketDerm/pull/18868

I'm not sure if that's also the case for patient-facing, but looking through things in PocketDerm, Larssiet is part of the fonts in the layout. You're probably all over that since this piece would need to be changed first, but wanted to point that in out just in case :) Those should be updated when the radiance version gets bumped.
https://github.com/curology/PocketDerm/blob/master/resources/views/app/layout/curology/fonts.blade.php

cherimarie:
@ZeMunchkin great callout, and yep, I'm on it! https://github.com/curology/PocketDerm/pull/21008

PR #1876, opened 2023-10-30
Add disableHover prop to Tooltip component

elysiagabe:
Currently, the tooltip will open either by clicking or hovering over the trigger when on desktop. On mobile, the tooltip opens only on click. This PR introduces the option to disable the hover behavior, such that the tooltip will only open on click for both mobile & desktop.

### DIFF
--- a/src/shared-components/tooltip/index.tsx
+++ b/src/shared-components/tooltip/index.tsx
@@ -32,6 +32,10 @@ export interface TooltipProps {
    */
   defaultOpen?: boolean;
   /**
+   * When true, the tooltip will open on click only and not on hover
+   */
+  disableHover?: boolean;
+  /**
    * Programatically control the tooltip to never show (false) or function as normal (true)
    */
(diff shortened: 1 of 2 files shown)


src/shared-components/tooltip/index.tsx
```diff
@@ -96,19 +101,29 @@ export const Tooltip: React.FC<TooltipProps> = ({
...
+    if (disableHover) {
+      return;
+    }
+    setHovered(true);
+  };
+
+  const handleMouseLeave = () => {
+    if (disableHover) {
+      return;
+    }
```
mdhamija94, line 114: I think this is harmless for the sake of being explicit/keeping parity between enter/leave, but might not be necessary since the `hovered` will already by set to false when this function is triggered when `disableHover` is true.

mdhamija94 (approved):
Looks good to me! One comment, but could go either way on it

PR #1881, opened 2024-05-17
Adds required/optional tag option to accordion

irislam94:
- Adds required/optional tag option to accordion
  - `isRequired` will show the text "Required" or "Optional" if the `showRequiredTag` property is set to "true"

Screenshot 2024-05-16 at 2 34 42 PM

Screenshot 2024-05-17 at 10 07 01 AM

### DIFF
--- a/src/shared-components/accordion/index.tsx
+++ b/src/shared-components/accordion/index.tsx
@@ -16,6 +16,10 @@ export interface AccordionProps {
   disabled?: boolean;
   /** determine if the accordion is collapsed (false) or expanded (true) */
   isOpen: boolean;
+  /** show if the photo is required */
+  isRequired?: boolean;
+  /** when true, shows Required/Optional tag */
+  showRequiredTag?: boolean;
   /** when true, border lines between accordions and title/children nodes will disappear */
   noBorder?: boolean;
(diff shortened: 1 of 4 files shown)


src/shared-components/accordion/index.tsx
```diff
@@ -16,6 +16,10 @@ export interface AccordionProps {
   disabled?: boolean;
   /** determine if the accordion is collapsed (false) or expanded (true) */
   isOpen: boolean;
+  /** show if the photo is required */
+  isRequired?: boolean;
+  /** when true, shows Required/Optional tag */
+  showRequiredTag?: boolean;
```
CurologyJonathanHartley, line 22: I might change the name of `showRequiredTag` to something like `displayRequiredOrOptionalText`.

The naming as is feels too similar to `isRequired`. Even with the comment block, I didn't fully understand the point of having both variables until looking at the actual code that references them.

CurologyJonathanHartley (approved):

src/shared-components/accordion/index.tsx
irislam94: good idea, I'll make the change!

PR #1882, opened 2026-06-24
adding secondaryRebrandTheme for agency rebrand

ericchalifour-spec:
Adds a new `secondaryRebrandTheme` to radiance-ui — a standalone, fully independent copy of `secondaryTheme`. This gives us a separate theme we can iterate on for the agency rebrand without touching (or risking regressions in) the existing `secondary` theme.

The new theme is an exact duplicate of `secondary` today; the intent is for design values (colors, typography, etc.) to diverge here over the course of the rebrand.

## Changes

**New theme value modules** (copies of their `secondary` counterparts, independently editable):
- `src/constants/colors/secondaryRebrand.ts`
- `src/constants/borderRadius/secondaryRebrand.ts`
- `src/constants/boxShadows/secondaryRebrand.ts`
- `src/constants/fonts/secondaryRebrand.ts`
- `src/constants/typography/secondaryRebrand.ts`

**New theme object:**
- `src/constants/themes/secondaryRebrandTheme.ts` — composes the modules above, `__type: 'secondaryRebrand'`

**Wiring:**
- `src/constants/themes/index.ts` — export `secondaryRebrandTheme`
- `src/constants/themes/types.ts` — add `secondaryRebrandTheme` to the 5 theme unions and `'secondaryRebrand'` to the `ThemeType['__type']` union
- `src/constants/themes/test.ts` — extend the key-parity test to cover the new theme

## Notes
- No existing themes or components are modified — purely additive.
- `secondaryRebrand` must keep the same keys as the other themes (enforced by the parity test in `themes/test.ts`).
- Consumers (e.g. Pocketderm) will pick this up via `radiance-ui/lib/constants` after this is released and their dependency is bumped. The consumer-side theme selection wiring is intentionally **not** part of this PR.

## Testing
- ✅ `jest src/constants/themes` — parity test passes (incl. new theme)
- ✅ `tsc` — 0 type errors


ericchalifour-spec:
@claude

amaidah (approved):

amaidah:
> @claude

ah i don't think we have it setup here

src/constants/themes/types.ts
```diff
@@ -35,5 +41,5 @@ export interface ThemeType {
   COLORS: Colors;
   FONTS: Fonts;
   TYPOGRAPHY: Typography;
-  __type: 'primary' | 'secondary' | 'tertiary';
+  __type: 'primary' | 'secondary' | 'secondaryRebrand' | 'tertiary';
```
alejandrogomez-curology, line 44: All four conditionals in themeStyles/index.ts are open if statements, never a switch with a default exhaustive check. TypeScript compiles cleanly but silently ignores unhandled members of the union. Consider adding a compile-time assertNever guard or converting the helpers to a lookup map keyed on `__type`, so adding a future theme is caught at build time rather than in production.

alejandrogomez-curology:
`themeStyles/index.ts` is not in this diff but is broken by the new `__type: 'secondaryRebrand'` value.

`setThemeFontWeight` (line 34) only matches `__type === 'secondary'`, so `secondaryRebrand` falls through to the else-branch and returns `fontWeight.bold` instead of `fontWeight.normal`. Any component using this helper with the new theme will render text with the wrong weight.

https://github.com/curology/radiance-ui/blob/master/src/utils/themeStyles/index.ts#L34

Fix: extend the guard to `theme.__type === 'secondary' || theme.__type === 'secondaryRebrand'`.

alejandrogomez-curology:
`setSecondaryHeadingFont` (line 22) checks `__type === 'secondary'` and returns `''` for everything else. With `secondaryRebrand`, Display / Heading / Title components will never apply `font-family: quadrant-text` — they'll silently fall back to the base sans-serif despite the theme having the correct `headerFont` value defined.

https://github.com/curology/radiance-ui/blob/master/src/utils/themeStyles/index.ts#L22

Same fix pattern as above.

src/constants/themes/types.ts
ericchalifour-spec: Good call! I added the definition for theme type to src/utils/themeStyles/index.ts so that if a new theme is added to this union but is unhandled it'll throw an error at build.

ericchalifour-spec:
@alejandrogomez-curology comments addressed and new commit pushed. Thanks for catching those errors!

alejandrogomez-curology (approved):

